PR #10, opened 2015-05-18
Added magic comment for encoding in python scripts.

seijikun:
Using the `Tools\Running a local Script` function, I am constantly getting ErrorMessages like:
```
Exception in thread "Thread-7" SyntaxError: Non-ASCII character in file '/mnt/coding/Java/PlayOnLinux/POL-POM-5/src/main/python/v4wrapper.py', but no encoding declared; see http://www.python.org/peps/pep-0263.html for details
```

This is a possible fix for that. But I am not too familiar with python, so that might be the wrong approach.

### DIFF
--- a/src/main/python/SetupWindow/SetupWindowManager.py
+++ b/src/main/python/SetupWindow/SetupWindowManager.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# coding=utf-8
 
 # Copyright (C) 2015 Pâris Quentin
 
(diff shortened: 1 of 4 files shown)


qparis:
It is in fact the right way to fix that. 
The bug comes from the copyright that has been added before. We really should write some unit test to cover this case.
Many thanks

seijikun:
I will write a JUnit-Test, if you haven't started doing so already.

Issue #12, opened 2015-05-19
LegacyScript execution fails since f9329e2


seijikun:
Since commit f9329e2, the execution of LegacyScripts fails on my Notebook. This is the exception I am receiving:
```
Exception in thread "MainThread" Traceback (most recent call last):
  File "/mnt/extern/Coding/Java/PlayOnLinux/POL-POM-5/src/main/python/v4wrapper.py", line 25, in <module>
    from SetupWindow.SetupWindowNetcatServer import SetupWindowNetcatServer
  File "/mnt/extern/Coding/Java/PlayOnLinux/POL-POM-5/src/main/python/SetupWindow/SetupWindowNetcatServer.py", line 21, in <module>
    from SetupWindowNetcatCommandParser import SetupWindowNetcatCommandParser
  File "/mnt/extern/Coding/Java/PlayOnLinux/POL-POM-5/src/main/python/SetupWindow/SetupWindowNetcatCommandParser.py", line 22, in <module>
    from com.playonlinux.framework import EnvironementHelper
ImportError: cannot import name EnvironementHelper
Exception in thread "Thread-8" Traceback (most recent call last):
  File "/mnt/extern/Coding/Java/PlayOnLinux/POL-POM-5/src/main/python/v4wrapper.py", line 25, in <module>
    from SetupWindow.SetupWindowNetcatServer import SetupWindowNetcatServer
  File "/mnt/extern/Coding/Java/PlayOnLinux/POL-POM-5/src/main/python/SetupWindow/SetupWindowNetcatServer.py", line 21, in <module>
    from SetupWindowNetcatCommandParser import SetupWindowNetcatCommandParser
  File "/mnt/extern/Coding/Java/PlayOnLinux/POL-POM-5/src/main/python/SetupWindow/SetupWindowNetcatCommandParser.py", line 22, in <module>
    from com.playonlinux.framework import EnvironementHelper
ImportError: cannot import name EnvironementHelper
```
The new python-scripts work flawlessly.

qparis:
This is interesting, so it means than python needs to have access to the constructor. Reverting back as it was only a good practice

qparis:
My mistake, it was infact a refactor problem. (A spellcheck has been made).
Keeping the constructor private

seijikun:
And I've been searching for the issue for multiple hours...
Life can be so easy sometimes.

qparis:
Problem fixed :)

PR #13, opened 2015-05-19
Implemented TestCase for the LegacyScript-Wrapper.

seijikun:
I now implemented a TestCase which is running a LegacyScript creating a temporary file and checking if that file exists afterwards. The way I used the PythonInterpreter here is somewhat ugly. Maybe we should create a static HelperClass for instantiating and setting up PythonInterpreters?

### DIFF
--- a/src/main/java/com/playonlinux/domain/Script.java
+++ b/src/main/java/com/playonlinux/domain/Script.java
@@ -75,10 +75,7 @@ public abstract class Script implements BackgroundService {
             @Override
             public void run() {
                 try {
-                    File pythonPath = new File("src/main/python");
-                    System.getProperties().setProperty("python.path", pythonPath.getAbsolutePath());
-                    PythonInterpreter pythonInterpreter = new PythonInterpreter();
-                    executeScript(pythonInterpreter);
+                    runScript();
                 } catch (PyException e) {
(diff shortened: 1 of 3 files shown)


qparis:
I don't understand why you need to run all that steps.
Wouldn't something like 
```java
playOnLinuxBackgroundServicesManager.register(testScript);
```

sufficient?

seijikun:
I tried what you suggest, but since this is asynchronous, I would have to put a sleep in the TestCase.
Waiting resulted in mixed results (pass / not pass) on my notebook and I had to use sleeps of 30 seconds and up (it takes so long until all python scripts are compiled I guess).
I thought that mixed results are nothing a TestCase should rely on, so I decided to use the synchronous approach. A nice side effect is that I can directly catch all Exceptions here. (this is where I found #12).

qparis:
Can you have a look at my last commit and see if it is fine for you?

seijikun:
That makes many things easier.
Even if the commit went somewhat wrong. ;)

qparis:
Line 32: Can you use File.createTemporaryFile instead?  (with delete on exit statement)

seijikun:
I don't understand, I am afraid.
The Java-Code does not create the file, the bash script does.

qparis:
Oh you are right. Sorry I'm tired

PR #14, opened 2015-05-19
Fix small Typo resolving FileNotFound issue.

seijikun:
Fixed small typo which caused that the `bash/lib/scripts` - script could not be found.


qparis:
Thanks :D

Issue #7, opened 2015-05-17
[Easy] Implement POL_Download in bash


qparis:
Currently, POL_Download is implemented in Python: 
This feature need to be implemented in bash for backward compatibility. 
As the feature is already implemented in the java core, the bash implementation will be straightforward.

The following code will download a script.

```python
from com.playonlinux.framework import SetupWizard
from com.playonlinux.framework import Downloader

setupWindow = SetupWizard("Test script")
Downloader(setupWindow).get("http://files.metaquotes.net/metaquotes.software.corp/mt4/mt4setup.exe").check("e42ebc9ebc2d8975de5a0c94386d1a6a")

setupWindow.close()
```

To achieve that you need to: 
 * Complete bash/lib/scripts (following the same structure than the commands you can find in bash/lib/setupWindow)
```bash
POL_Download ()
{
	# Download a file and place it in the current directory
	# Usage: POL_Download [URL] [MD5]

        # To be implemented
}
```

 * Add a parser case in SetupWindowCommandParser.py responding to the command you have created before.

4thguy:
I want to see if I understood your request correctly.

You want a script that performs a download (wget), checks the file against a hash, and returns a success or fail. Then you want to check the result using Python in SetupWindowCommandParser.py.

qparis:
Not exactly, the download part has already been implemented.

Another guy has already started on this one, have a look at the pull request list :-)

4thguy:
OK, thanks for the heads up.

My Python-fu is very weak, but hopefully I'll be able to contribute more with the Java aspects of the project.

qparis:
Thanks :-)

You can have a look at the pull request and do the same for POL_SetupWindow_free_presentation if you want

qparis:
Fixed by commit 632b240aa6f09d4b47f2d2b022a530d975600e59
Many thanks to seijikun

PR #15, opened 2015-05-19
[FIX] [Github #7] Implement POL_Download

seijikun:
Initial implementation of POL_Download.
I hope I understood the system correctly.


qparis:
Yep that's it!
I think it would be great to put another URL on the example script tough :)

seijikun:
Would that be better? :)

qparis:
Perfect

PR #17, opened 2015-05-21
[FIX #6] Add scene for download failure to the InstallWindow.

### LINKED ISSUE #6: [Easy/Medium] Create a scene in InstallWindow in the download fail case
InstallWindow is an observer of RemoteAvailableInstallers class.
RemoteAvailableInstallers' responsibility is to download the list of available PlayOnLinux installers. This class has three possible state: 
* Downloading in progress
* Download is done 
* Download has failed

The latter case is not handled by InstallWindow class, which means that install window will not tell the user if the download has failed. 
To support this, we need to go to the method: 
   public void update(RemoteAvailableInstallers remoteAvailableInstallers) 
 
Then, create a "failedScene" which contains a text explaining the user that the download has failed, and a button so that he/she can retries. The button will behave exactly the same as refreshButton which is in the same class.

seijikun:
It's not the most beautiful, but it probably is as beautiful as a UI from me will ever get. :)
Should the font-size of the message be increased?


qparis:
Why are you importing Swing?

Also, can you attach a screenshot so that I can see? :)

Many thanks!

seijikun:
The swing-import was unintended. IntelliJ going crazy :)
Screenshot

qparis:
It is very clean.

PR #19, opened 2015-05-21
Rewrote InstallWindow's mainScene with automatic sizing.

seijikun:
I rewrote InstallWindow's mainScene with automatic sizing by using the BorderPane and its regions.


qparis:
Awesome!

qparis:
I reverted your change so that we can fix a small issue. Feel free to have a look. Apart from that details, good job!

seijikun:
The miniature would not be a problem for the right region. But the checkboxes are.
Would it be ok to move the checkboxes to the bottom of the window (bottom left, below the list)?

qparis:
I am even open to completely redesign the window, for exemple like Ubuntu Software Center or any other. 
If you have some plans, you can suggest :)

Otherwise isn't it possible to put the checkboxes in a GridPane?

seijikun:
I just got an idea of how to do it. :)

Issue #24, opened 2015-05-22
[Medium] Add filter feature in InstallWindow 


qparis:
Three filter checkbox have been added to the installation window. We now need to connect these three checkboxes to the domain object. 
In order to do that, we need to:
* Add the following two methods in the com.playonlinux.ui.api.RemoteAvailableInstallers interface 
** Iterable<ScriptDTO> getAllScripts(String filter, boolean commercial, boolean requiresNoCd, boolean testing);
** Iterable<ScriptDTO> getAllScriptsInCategory(String categoryName, boolean commercial, boolean requiresNoCd, boolean testing) throws PlayOnLinuxError;

* The original methods should call these new method with the default behavior (true, false, false) 
* Implement the two methods in RemoteAvailableInstallersPlayOnLinuxImplementation
* Use these new methods in AvailableInstallerListWidget

The design of this feature can be improved later to allow to add extra filters.

seijikun:
I am currently working with this code anyway, so I'll take this one.

seijikun:
Is it intended, that `ScriptInformationsDTO` does not contain the commercial flag?

qparis:
isCommercial = !isFree in fact. 
It is the way the information is stored in the database.

seijikun:
This has been implemented by #91.

qparis:
We still need to implement the behavior of the filter, but it should be easy. The big part of the work has been done :)

PR #27, opened 2015-05-23
Refactor the Filtering of Installers

seijikun:
I've refactored the filtering for Installer-Scripts displayed in the InstallWindow.
This is a fix for #24.

Abstract:
I created a class (InstallerFilter) which handles all filter parameters as well as the filtering itself. This extracts the filter-logic out of the View-Part (AvailableInstallerList) and gives us a central point where we can easily change this behavior or add new parameters. InstallerFilter is an Observable, automatically notifying observers (the AvailableInstallerList) that the filter-parameters have changed, causing a refiltering and a UI refresh.

### DIFF
--- a/src/main/java/com/playonlinux/services/RemoteAvailableInstallersPlayOnLinuxImplementation.java
+++ b/src/main/java/com/playonlinux/services/RemoteAvailableInstallersPlayOnLinuxImplementation.java
@@ -23,6 +23,7 @@ import com.playonlinux.common.dtos.*;
 import com.playonlinux.domain.PlayOnLinuxError;
 import com.playonlinux.injection.Scan;
 import com.playonlinux.injection.Inject;
+import com.playonlinux.ui.api.InstallerFilter;
 import com.playonlinux.webservice.RemoteAvailableInstallers;
 
 import java.net.MalformedURLException;
@@ -96,7 +97,15 @@ public class RemoteAvailableInstallersPlayOnLinuxImplementation extends Observab
 
     @Override
     public Iterable<ScriptDTO> getAllScripts() {
-        return getAllScripts(null);
+        List<ScriptDTO> scripts = new ArrayList<>();
+        for(CategoryDTO categoryDTO: new ArrayList<>(categoriesDTO)) {
+            for(ScriptDTO scriptDTO: new ArrayList<>(categoryDTO.getScripts())) {
+                scripts.add(scriptDTO);
+            }
+        }
+
+        Collections.sort(scripts, new ScriptDTO.AlphabeticalOrderComparator());
+        return () -> scripts.iterator();
     }
 
     @Override
@@ -116,13 +125,21 @@ public class RemoteAvailableInstallersPlayOnLinuxImplementation extends Observab
     }
 
     @Override
-    public Iterable<ScriptDTO> getAllScriptsInCategory(String categoryName) throws PlayOnLinuxError {
-        for(CategoryDTO categoryDTO: categoriesDTO) {
-            if(categoryName.equals(categoryDTO.getName())) {
-                return getAllScriptsInCategory(categoryDTO);
+    public Iterable<ScriptDTO> getAllScripts(InstallerFilter filter) throws PlayOnLinuxError {
+        Iterable<ScriptDTO> scriptsToFilter;
+        List<ScriptDTO> scripts = new ArrayList<>();
+        if(filter.getCategory() == null){
+            scriptsToFilter = this.getAllScripts();
+        }else{
+            scriptsToFilter = this.getAllScriptsInCategory(filter.getCategory());
+        }
+
+        for(ScriptDTO script : scriptsToFilter){
+            if(filter.apply(script)){
+                scripts.add(script);
             }
         }
-        throw new PlayOnLinuxError(String.format("The category %s does not exist!", categoryName));
+        return () -> scripts.iterator();
     }
 
     @Override
@@ -145,6 +162,15 @@ public class RemoteAvailableInstallersPlayOnLinuxImplementation extends Observab
         remoteAvailableInstallers = new RemoteAvailableInstallers(webserviceUrl);
         remoteAvailableInstallers.addObserver(this);
         playOnLinuxBackgroundServicesManager.register(remoteAvailableInstallers);
+    }
+
+    private Iterable<ScriptDTO> getAllScriptsInCategory(String categoryName) throws PlayOnLinuxError {
+        for(CategoryDTO categoryDTO: categoriesDTO) {
+            if(categoryName.equals(categoryDTO.getName())) {
+                return getAllScriptsInCategory(categoryDTO);
+            }
+        }
+        throw new PlayOnLinuxError(String.format("The category %s does not exist!", categoryName));
     }
 
     private Iterable<ScriptDTO> getAllScriptsInCategory(CategoryDTO categoryDTO) {
--- a/src/main/java/com/playonlinux/services/RemoteAvailableInstallersPlayOnLinuxImplementation.java
+++ b/src/main/java/com/playonlinux/services/RemoteAvailableInstallersPlayOnLinuxImplementation.java
@@ -65,7 +65,7 @@ public class RemoteAvailableInstallersPlayOnLinuxImplementation extends Observab
         downloadEnvelopeDto = (DownloadEnvelopeDTO<AvailableCategoriesDTO>) arg;
 
         try {
-            if(downloadEnvelopeDto.getEnvelopeContent() != null) {
+            if (downloadEnvelopeDto.getEnvelopeContent() != null) {
                 List<CategoryDTO> availableCategories = new ArrayList<>(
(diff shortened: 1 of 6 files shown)


src/main/java/com/playonlinux/ui/api/InstallerFilter.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Test the given script against the filter rules.
+     * <p>
+     * Note: At the moment, this method does not filter categories, since Scripts don't know about their own category.
+     * </p>
+     *
+     * @param script Script which should be tested against the filter.
+     * @return True if this Script matches the filter criteria, false otherwise.
+     */
+    public boolean apply(ScriptDTO script) {
```
seijikun: This method is where the filtering is done. It is used by the `RemoteAvailableInstallersPlayOnLinuxImplementation` and thus the central place to change the behavior if needed.

src/main/java/com/playonlinux/ui/impl/javafx/installwindow/InstallWindow.java
```diff
@@ -242,10 +243,10 @@ private void setUpEvents() throws PlayOnLinuxError {
                 playOnLinuxError.printStackTrace();
             }
         });
```
seijikun: Using the KeyReleased event instead of the KeyPressed event fixes that the Title-Filter was hanging behind by one char.

src/main/java/com/playonlinux/ui/impl/javafx/installwindow/InstallWindow.java
```diff
@@ -242,10 +243,10 @@ private void setUpEvents() throws PlayOnLinuxError {
...
-        searchWidget.setOnKeyPressed(event -> availableInstallerListWidget.setSearchFilter(searchWidget.getText()));
-        testingCheck.setOnAction(event -> availableInstallerListWidget.setIncludeTesting(testingCheck.isSelected()));
-        noCdNeededCheck.setOnAction(event -> availableInstallerListWidget.setIncludeNoCDNeeded(noCdNeededCheck.isSelected()));
-        commercialCheck.setOnAction(event -> availableInstallerListWidget.setIncludeCommercial(commercialCheck.isSelected()));
+        searchWidget.setOnKeyReleased(event -> availableInstallerListWidget.getFilter().setTitle(searchWidget.getText()));
+        testingCheck.setOnAction(event -> availableInstallerListWidget.getFilter().setShowTesting(testingCheck.isSelected()));
+        noCdNeededCheck.setOnAction(event -> availableInstallerListWidget.getFilter().setShowNoCd(noCdNeededCheck.isSelected()));
+        commercialCheck.setOnAction(event -> availableInstallerListWidget.getFilter().setShowCommercial(commercialCheck.isSelected()));
```
seijikun: All filter-parameter are now set in the `AvailableInstallerList's` Filter-Property.
The `AvailableFilterList` observes the `InstallerFilter` and thus updates after a filter-parameter has changed.

src/main/java/com/playonlinux/ui/api/InstallerFilter.java
```diff
@@ -0,0 +1,142 @@
...
+            return false;
+        }
+        if (StringUtils.isNotBlank(title)) {
+            if (!script.getName().contains(title)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
```
seijikun: This method is where the filtering is done. It is used by the `RemoteAvailableInstallersPlayOnLinuxImplementation` and thus the central place to change the behavior if needed.

src/main/java/com/playonlinux/ui/api/InstallerFilter.java
```diff
@@ -0,0 +1,142 @@
...
+     */
+    public boolean apply(ScriptDTO script) {
+        //TODO: write testcases for this method.
+        ScriptInformationsDTO scriptInfo = script.getScriptInformations();
+        boolean isTesting = scriptInfo.getTestingOperatingSystems().contains(hostOs);
+
+        if (isTesting && !showTesting) {
+            return false;
+        }
+        if (scriptInfo.isRequiresNoCD() && !showNoCd) {
```
seijikun: This method is where the filtering is done. It is used by the `RemoteAvailableInstallersPlayOnLinuxImplementation` and thus the central place to change the behavior if needed.

src/main/java/com/playonlinux/ui/impl/javafx/installwindow/InstallWindow.java
```diff
@@ -242,10 +243,10 @@ private void setUpEvents() throws PlayOnLinuxError {
                 playOnLinuxError.printStackTrace();
             }
         });
-        searchWidget.setOnKeyPressed(event -> availableInstallerListWidget.setSearchFilter(searchWidget.getText()));
-        testingCheck.setOnAction(event -> availableInstallerListWidget.setIncludeTesting(testingCheck.isSelected()));
-        noCdNeededCheck.setOnAction(event -> availableInstallerListWidget.setIncludeNoCDNeeded(noCdNeededCheck.isSelected()));
-        commercialCheck.setOnAction(event -> availableInstallerListWidget.setIncludeCommercial(commercialCheck.isSelected()));
+        searchWidget.setOnKeyReleased(event -> availableInstallerListWidget.getFilter().setTitle(searchWidget.getText()));
```
seijikun: Using the `KeyReleased` event instead of the `KeyPressed` event fixes that the Title-Filter was hanging behind by one char.

src/main/java/com/playonlinux/ui/api/InstallerFilter.java
```diff
@@ -0,0 +1,142 @@
...
+    private final OperatingSystem hostOs;
+
+
+    public InstallerFilter() {
+        //TODO: Pre-define filters.
+        OperatingSystem os;
+        try {
+            os = OperatingSystem.fetchCurrentOperationSystem();
+        } catch (PlayOnLinuxError err) {
+            os = null;
```
qparis: I don't really like doing this. Couldn't we manage the exception in a different way? Maybe adding a default operating system policy? At least we need to add a todo here

src/main/java/com/playonlinux/ui/impl/javafx/installwindow/AvailableInstallerListWidget.java
```diff
@@ -35,62 +35,35 @@
...
-                    alert.setContentText(String.format("The error was: %s", playOnLinuxError));
-                    playOnLinuxError.printStackTrace();
+            try {
+                for (ScriptDTO script : remoteAvailableInstallers.getAllScripts(filter)) {
+                    this.addItem(script.getName());
                 }
+            } catch (PlayOnLinuxError playOnLinuxError) {
+                Alert alert = new Alert(Alert.AlertType.ERROR);
+                alert.setTitle(translate("Error while trying to show installer list."));
+                alert.setContentText(String.format("The error was: %s", playOnLinuxError));
```
qparis: We have forgotten alert.show() here

qparis:
I'm comitting your code but I've made few comment.
Thanks

src/main/java/com/playonlinux/services/RemoteAvailableInstallersPlayOnLinuxImplementation.java
```diff
@@ -23,6 +23,7 @@
 import com.playonlinux.domain.PlayOnLinuxError;
 import com.playonlinux.injection.Scan;
 import com.playonlinux.injection.Inject;
+import com.playonlinux.ui.api.InstallerFilter;
```
qparis: You need to put this class in the same package than RemoteAvailableObserver

src/main/java/com/playonlinux/ui/impl/javafx/installwindow/AvailableInstallerListWidget.java
```diff
@@ -35,62 +35,35 @@
...
-    private String searchFilter = "";
 
     AvailableInstallerListWidget(InstallWindowEventHandler installWindowEventHandler) throws PlayOnLinuxError {
         super();
         this.installWindowEventHandler = installWindowEventHandler;
         this.installWindowEventHandler.getRemoteAvailableInstallers().addObserver(this);
         remoteAvailableInstallers = this.installWindowEventHandler.getRemoteAvailableInstallers();
+
+        AvailableInstallerListWidget me = this;
+        this.filter.addObserver((observable, o) -> me.update());
```
qparis: Just use this.filter.addObserver((observable, o) -> update()); here

src/main/java/com/playonlinux/ui/impl/javafx/installwindow/InstallWindow.java
```diff
@@ -283,7 +283,7 @@ public void update(Observable o, Object arg) {
 
     public void clearSearch() {
         searchWidget.clear();
-        availableInstallerListWidget.setSearchFilter("");
+        availableInstallerListWidget.getFilter().setTitle(null);
```
qparis: I guess a blank title is perfectly valid here. No need to introduce a null pointer

PR #34, opened 2015-05-24
[WIP] Filter standardization and refactoring for Installers. Version 2

seijikun:
A standardization for the UI how the Filtering of DTO's should be handled.
I've refactored the filtering for Installer-Scripts displayed in the InstallWindow.
This is a fix for #24.

Abstract:
The interface Filterable<DTO> defines how a container of DTO's should look for being filterable. The Filter<DTO> interface defines how the filter (doing the actual filtering) should look and work like.

Every container of DTO's then needs an own implementation of Filter<DTO> to be able to export all filter-parameters to the callers. In the case of Installers, this is done with the abstract class within `RemoteAvailableInstallers`, only defining the Filter's getters and setters.

### DIFF
--- a/src/main/java/com/playonlinux/ui/api/RemoteAvailableInstallers.java
+++ b/src/main/java/com/playonlinux/ui/api/RemoteAvailableInstallers.java
@@ -20,12 +20,14 @@ package com.playonlinux.ui.api;
 
 import com.playonlinux.common.dtos.CategoryDTO;
 import com.playonlinux.common.dtos.ScriptDTO;
+import com.playonlinux.common.dtos.ScriptInformationsDTO;
 import com.playonlinux.domain.PlayOnLinuxError;
+import com.playonlinux.utils.OperatingSystem;
 
-import java.util.Iterator;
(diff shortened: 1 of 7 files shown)


src/main/java/com/playonlinux/ui/api/RemoteAvailableInstallers.java
```diff
@@ -34,13 +36,82 @@
...
+    abstract class InstallerFilter extends Observable implements Filter<ScriptDTO> {
+
+        protected String title;
+        protected String category;
+        protected boolean showTesting;
+        protected boolean showNoCd;
+        protected boolean showCommercial;
+
+        public String getTitle() {
+            return title;
```
seijikun: This abstract class is needed within RemoteAvailableInstallers, so that we can overload `Filterable<DTO>`'s getFilter(), with the more specific filter for this DTO-Container. Without this, all users of the `getFilter()` function would have to cast the returned `Filter<DTO>` to the more concrete `InstallerFilter` to be able to set filter-parameters.

PR #39, opened 2015-05-24
Optimize findAllFilesByExtension (less memory usage)

Tutul-:
- Use Array instead of ArrayList  (less memory usage)
- Can be faster when EXCLUDES_FILES is bigger, now it's a little bit slower (~0.030s)
BONUS : change .gitignore for Eclipse stuff and *.class


qparis:
Ok to test

qparis:
Test this please

qparis:
Ok to test

src/main/java/com/playonlinux/wine/WinePrefix.java
```diff
@@ -41,6 +44,7 @@
     private static final String EXECUTABLE_EXTENSION = "exe";
     private static final String DRIVE_C = "drive_c";
 
+    private static String[] EXCLUDES_FILES = new String[1]; // One element for now
```
qparis: It is better to directly set here the list of exclusion:
private static String[] EXCLUDES_FILES = new String[] { "iexplore.exe", "notepad.exe" }, etc...

Tutul-:
Done

qparis:
Ok to test

PR #47, opened 2015-05-25
Add confirm popup when exit

Tutul-:
We must confirm when exiting (like POLv4)
Now, closing the mainWindows close all the program

### DIFF
--- a/src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/MainWindow.java
+++ b/src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/MainWindow.java
@@ -26,13 +26,11 @@ import com.playonlinux.ui.api.PlayOnLinuxWindow;
 import com.playonlinux.ui.impl.javafx.common.PlayOnLinuxScene;
 
 import javafx.application.Platform;
-import javafx.event.EventHandler;
 import javafx.scene.Scene;
 import javafx.scene.control.Alert;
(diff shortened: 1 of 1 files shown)


src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/MainWindow.java
```diff
@@ -84,6 +90,22 @@ public void setUpWindow() {
         this.setScene(scene);
         this.setTitle(translate("${application.name}"));
         this.show();
+        
+        
+        scene.getWindow().setOnCloseRequest(new EventHandler<WindowEvent>() {
```
qparis: Can you use a lambda expression instead please?

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/MainWindow.java
```diff
@@ -84,6 +90,22 @@ public void setUpWindow() {
...
+            @Override
+            public void handle(WindowEvent event) {
+                Alert alert = new Alert(Alert.AlertType.CONFIRMATION);
+                alert.setTitle(translate("Would you like to exit?"));
+                alert.setHeaderText(translate("Are You Sure?"));
```
qparis: Can you try to use the same text that exist in PlayOnLinux 4?

PR #68, opened 2015-05-30
Add message on ScriptFailureException + replace CancelException

Tutul-:


src/main/java/com/playonlinux/framework/WinePrefix.java
```diff
@@ -63,7 +64,7 @@ public WinePrefix create(String version) throws ScriptFailureException {
         try {
             return this.create(version, Architecture.fetchCurrentArchitecture().name());
         } catch (PlayOnLinuxException e) {
-            throw new ScriptFailureException(e);
+            throw new ScriptFailureException("Unable to create the wineprefix", e);
```
qparis: Maybe putting a more accurate message here?

PR #79, opened 2015-05-31
Create an interface for classes that need instant cleanup.

seijikun:
The `dispose()` method adds the functionality for directly telling a class to clean itself up if necessary.
This is the point where resources like streams should be closed.


qparis:
ok to test

PR #90, opened 2015-06-07
Use PO files for translation like gettext

yawkat:
Not entirely happy with injecting a static field but the basic implementation is there. Tested with German language in the existing interface.

- I'm not sure how to handle the zh_CN ; zh_TW pair since neither specify the top-level zh language.
- en_GB also isn't used as default but instead strings aren't translated as fallback, is this okay?

### DIFF
--- a/src/main/java/com/playonlinux/domain/Localisation.java
+++ b/src/main/java/com/playonlinux/domain/Localisation.java
@@ -19,14 +19,15 @@
 package com.playonlinux.domain;
 
 import com.playonlinux.app.PlayOnLinuxContext;
+import com.playonlinux.domain.lang.LanguageBundle;
 import com.playonlinux.injection.Inject;
 import com.playonlinux.injection.Scan;
 import com.playonlinux.utils.ReplacableProperties;
 
-
-// TODO
 @Scan
 public final class Localisation {
+    @Inject
+    private static LanguageBundle bundle;
     @Inject
     private static PlayOnLinuxContext playOnLinuxContext;
 
@@ -39,6 +40,6 @@ public final class Localisation {
         ReplacableProperties properties;
         properties = playOnLinuxContext.loadProperties();
 
-        return properties.replaceAllVariables(stringToTranslate);
+        return properties.replaceAllVariables(bundle.translate(stringToTranslate));
     }
 }
--- a/src/main/java/com/playonlinux/domain/lang/CatalogLanguageBundle.java
+++ b/src/main/java/com/playonlinux/domain/lang/CatalogLanguageBundle.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (C) 2015 Jonas Konrad
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package com.playonlinux.domain.lang;
+
+import com.google.common.base.Preconditions;
+import org.fedorahosted.tennera.jgettext.Catalog;
+import org.fedorahosted.tennera.jgettext.Message;
+
+/**
+ * Language bundle from gnu gettext PO files using jgettext.
+ */
+public class CatalogLanguageBundle implements LanguageBundle {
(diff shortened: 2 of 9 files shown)


qparis:
Can you use your injection method?

src/main/java/com/playonlinux/domain/Localisation.java
```diff
@@ -19,14 +19,21 @@
...
 import com.playonlinux.injection.Scan;
 import com.playonlinux.utils.ReplacableProperties;
+import java.util.Locale;
 
-
-// TODO
 @Scan
 public final class Localisation {
+    private static final LanguageBundle DEFAULT_BUNDLE;
+
```
qparis: I think you can declare this bean in PlayOnLinuxConfig class file and use @Inject annotation

qparis:
ok to test

yawkat:
Okay I will move that to the config

src/main/java/com/playonlinux/domain/lang/CatalogLanguageBundle.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public CatalogLanguageBundle(Catalog catalog) {
+        this.catalog = catalog;
+    }
+
+    public static CatalogLanguageBundle fromCatalog(Catalog catalog) {
+        return new CatalogLanguageBundle(catalog);
+    }
+
+    @Override
```
qparis: Why do we need these two methods?

src/main/java/com/playonlinux/domain/lang/LanguageBundleSelector.java
```diff
@@ -0,0 +1,71 @@
...
+            return parseCatalogBundle(poResourceUrl);
+        }
+    }
+
+    private static CatalogLanguageBundle parseCatalogBundle(URL poResourceUrl) {
+        PoParser parser = new PoParser();
+        try (InputStream inputStream = poResourceUrl.openStream()) {
+            return CatalogLanguageBundle.fromCatalog(parser.parseCatalog(inputStream, false));
+        } catch (IOException e) {
+            // this should only happen if a PO file is broken so we throw an Error
```
qparis: Can we put the interface in english and log the error instead of crashing the entire app :)

yawkat:
Tested and amended

src/main/java/com/playonlinux/domain/lang/LanguageBundleSelector.java
yawkat: Well it is a compile-time error unless the POs are outdated, you sure?

yawkat: To clarify, users can't break it at run time, it won't error on compile

qparis:
The build is failing :) Can you have a look?
Many thanks

src/main/java/com/playonlinux/domain/lang/LanguageBundleSelector.java
qparis: What do you mean by It is a compile time error?

yawkat:
Fixed, was missing the bean in testing

qparis:
Cool! This is exactly the point of this @Inject annotation, it allows to mock the bean up for testing purposes

yawkat:
Yep, I'm just not too comfortable with injecting static fields but I suppose it is useful for often-used features like translation

qparis:
Personally I'm not confortable with static fields at all. We need to inject Static field in PlayOnLinux because it is the easiest way to have our beans accessible from Jython and JavaFX. (I haven't found any better method so far)

src/main/java/com/playonlinux/domain/lang/LanguageBundleSelector.java
qparis: Is the fact that the .po is broken completely prevent us from using the app and then, requires to throw an error?

src/main/java/com/playonlinux/domain/lang/CatalogLanguageBundle.java
yawkat: Oh, forgot about the fromCatalog one, the constructor used to be private

src/main/java/com/playonlinux/domain/lang/LanguageBundleSelector.java
yawkat: Actually, it should only happen when theres an FS / zip / other classloader issue:

- Missing PO is checked above in `poResourceUrl == null`
- Broken PO will throw a `org.fedorahosted.tennera.jgettext.catalog.parse.ParseException` which will escalate anyway

yawkat:
Oh, you're right, javafx doesn't really do that. ThreadLocals may work but I'm not sure how javafx manages threads.

src/main/java/com/playonlinux/domain/lang/LanguageBundleSelector.java
qparis: In this case we could still use the failback right?

qparis: What do you think?

qparis:
Ok I think it is fine. I'm merging now

qparis:
Many thanks :)

PR #91, opened 2015-06-08
Refactor filtering Version 3

seijikun:
Refactored the process of filtering, which then can easily be applied for the other sections (Shortcuts, Engine-Versions, ...).
We should talk about how and where to handle the flag-filtering (testing / NoCd / Commercial) - which I left unimplemented for now.

__Abstract__:
`Filter<T>` and `Filterable<T>` are both observable. Whenever something changes (e.g. filter-parameters), the observers are notified.
I combined both within `FilterPromise<T>`, which listens to Filter and Filterable and updates, after either one of them changed - directly providing the new filtered list.
I outsourced the list of categories into an own widget (`CategoryView`) to avoid unnecessary repaints and potential memory leaks. The CategoryView is passed an `ObservableList<T>` which fires an update after its list has changed and therefore causing the CategoryView to repaint the list of CategoryButtons.
The CategoryView is rendered nearly the same as before, however I'm not sure if that's not a memory-leak - since the observers of the CategoryButtons are never removed. We should probably have an eye on that one.

### DIFF
--- a/src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/center/CategoryView.java
+++ b/src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/center/CategoryView.java
@@ -72,9 +72,7 @@ public class CategoryView extends VBox implements Observer {
 
     private void fireCategorySelection(String categoryName){
         for(CategorySelectionObserver o : observers){
-            try{
-                o.update(this, categoryName);
-            }catch(Exception ex){}
+            o.update(this, categoryName);
         }
     }
 
--- a/src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/center/CategoryView.java
+++ b/src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/center/CategoryView.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2015 Markus Ebner
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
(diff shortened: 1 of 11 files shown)


qparis:
ok to test

qparis:
Can you fix it? :)

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/center/CategoryView.java
```diff
@@ -0,0 +1,89 @@
...
+import javafx.scene.layout.VBox;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Observable;
+import java.util.Observer;
+
+import static com.playonlinux.domain.Localisation.translate;
+
+public class CategoryView extends VBox implements Observer {
```
qparis: Can you put this class final?

qparis: Can you put this class final?

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/center/CategoryView.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public void deleteObserver(CategorySelectionObserver o){
+        this.observers.remove(o);
+    }
+
+    private void fireCategorySelection(String categoryName){
+        for(CategorySelectionObserver o : observers){
+            try{
+                o.update(this, categoryName);
+            }catch(Exception ex){}
```
qparis: Can you remove this catch Exception?

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/center/ViewApps.java
```diff
@@ -42,17 +42,19 @@
 public class ViewApps extends HBox implements Observer {
     private VBox failurePanel;
     private Button retryButton;
-    private List<CenterCategoryDTO> categories;
-    private List<CenterItemDTO> centerItems;
+    private ObservableArrayList<CenterCategoryDTO> categories;
+    private FilterPromise<CenterItemDTO> centerItems = null;
```
qparis: Do we need to put = null here?

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/center/ViewApps.java
```diff
@@ -65,6 +67,9 @@ public ViewApps(MainWindow parent) {
         this.initWait();
         this.initFailure();
 
+        categories = new ObservableArrayList<>();
+        centerItems = new FilterPromise<CenterItemDTO>(eventHandlerCenter.getRemoteAvailableInstallers(), this.filter);
```
qparis: You can replace <CenterItemDTO> with <>

PR #100, opened 2015-06-13
Main window search

Veske:
MainWindow search bar now has filtering applied to it:
   * Rename ShortcutDTO to InstalledApplicationDTO for better clarification.
   * Create a new filter for InstalledApplicationDTO.
   * Make ApplicationListWidget apply the new filter for user keypress events on the shown data.

Notes:
   * Left the filter apply() so that it would show all applications on start unlike it is done in ViewApps.
   * ApplicationListWidget update() is called twice on startup. I was not able to tacle the problem but it is currently in the main branch too. Could use some looking into.

### DIFF
--- a/src/main/java/com/playonlinux/common/filter/InstalledApplicationFilter.java
+++ b/src/main/java/com/playonlinux/common/filter/InstalledApplicationFilter.java
@@ -47,16 +47,7 @@ public class InstalledApplicationFilter extends Observable implements Filter<Ins
 
     @Override
     public boolean apply(InstalledApplicationDTO item) {
-        if(StringUtils.isBlank(name)) {
-            return false;
-        }
-
-        if(StringUtils.isNotBlank(name)){
-            if(!item.getName().toLowerCase().contains(name)) {
-                return false;
-            }
-        }
-        return false;
+        return !StringUtils.isNotBlank(name) || item.getName().toLowerCase().contains(name);
     }
 
     private void fireUpdate() {
--- a/src/main/java/com/playonlinux/common/filter/InstalledApplicationFilter.java
+++ b/src/main/java/com/playonlinux/common/filter/InstalledApplicationFilter.java
@@ -47,6 +47,7 @@ public class InstalledApplicationFilter extends Observable implements Filter<Ins
 
     @Override
     public boolean apply(InstalledApplicationDTO item) {
+        // We want to return the whole list for empty search string. Otherwise compare strings.
         return !StringUtils.isNotBlank(name) || item.getName().toLowerCase().contains(name);
     }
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/playonlinux/common/services/InstalledApplicationsPlayOnLinuxImplementation.java
```diff
@@ -42,7 +44,9 @@
     static BackgroundServiceManager playOnLinuxBackgroundServicesManager;
 
     ShortcutSet shortcutSet;
-    private Iterator<ShortcutDTO> shortcutDtoIterator;
+    private Iterator<InstalledApplicationDTO> shortcutDtoIterator;
+    private List<InstalledApplicationDTO> cache = null;
```
qparis: No need to initialize this field to null

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/library/ApplicationListWidget.java
```diff
@@ -37,16 +41,24 @@
...
+
     public ApplicationListWidget(ViewLibrary parent) {
         this.parent = parent;
         this.rootItem = new TreeItem<>();
         this.setRoot(rootItem);
         this.setShowRoot(false);
+        try {
+            installedApplications = new FilterPromise<>(this.parent.getEventHandler().getInstalledApplications(), this.filter);
+        } catch (PlayOnLinuxException e) {
+            e.printStackTrace();
```
qparis: You need to log this exception and to popup an alert

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/library/ViewLibrary.java
```diff
@@ -62,7 +61,10 @@ private void drawSideBar() {
 
         this.getChildren().add(leftContent);
 
-        TextField searchBar = new TextField();
+        searchBar = new TextField();
+        searchBar.setOnKeyReleased(event -> {
```
qparis: Just remove the {}, as you have only one instruction in this lambda

src/main/java/com/playonlinux/common/filter/InstalledApplicationFilter.java
```diff
@@ -0,0 +1,78 @@
...
+import com.playonlinux.common.dto.ui.InstalledApplicationDTO;
+import org.apache.commons.lang.StringUtils;
+
+import java.net.URL;
+import java.util.Observable;
+
+/**
+ * Filter for installed applications in the MainWindow
+ */
+public class InstalledApplicationFilter extends Observable implements Filter<InstalledApplicationDTO> {
```
qparis: This class should be tested

qparis:
ok to test

PR #101, opened 2015-06-14
Add default presentation system

Tutul-:
presentation() have two methode now, one for free presentation and one with default text format. The second way support translation system (but may be updated for use the new variable)
+ Some correction for javadoc ;)

### DIFF
--- a/src/main/java/com/playonlinux/framework/SetupWizard.java
+++ b/src/main/java/com/playonlinux/framework/SetupWizard.java
@@ -109,7 +109,7 @@ public class SetupWizard {
      * @throws CancelException
      */
     public void presentation(String programName, String programEditor, String editorURL, String scriptorName, String prefixName) throws CancelException {
-        String textToShow = String.format(translate("This wizard will help you install %1$s on your computer.\n\n"
+        final String textToShow = String.format(translate("This wizard will help you install %1$s on your computer.\n\n"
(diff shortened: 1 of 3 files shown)


src/main/java/com/playonlinux/framework/SetupWizard.java
```diff
@@ -99,6 +100,27 @@ public void execute(Message message) {
...
+     * @param programEditor the editor of the program
+     * @param editorIRL the editor website URL
+     * @param scriptorName the scriptor name
+     * @param prefixName the name of the prefix for that program
+     * @throws CancelException
+     */
+    public void presentation(String programName, String programEditor, String editorURL, String scriptorName, String prefixName) throws CancelException {
+        StringBuffer buffer = new StringBuffer();
+        Formatter formatter = new Formatter(buffer);
+        formatter.format("This wizard will help you install %1$s on your computer.\n\nThis program was created by: %2$s\n%3$s\n\nThis installation program is provided by: %4$s\n\n%1$s will be installed in: ${application.user.wineprefix}%5$s\n\n${application.name} is not responsible for anything that might happen as a result of using these scripts.\n\nClick Next to start", programName, programEditor, editorURL, scriptorName, prefixName);
```
qparis: Simply use String.format() here.
Also, try to make shorter lines

PR #119, opened 2015-06-26
Fix some SonarQube issues

Tutul-:
[WinePrefix] rename to comply with Java format and log missing error
[AppPanel] rename to comply with Java format
           + Change HTMLAnchorElementImpl to use Apache instead of sun
           + fix some conversion on log missing error

### DIFF
--- a/src/main/java/com/playonlinux/domain/ShortcutSet.java
+++ b/src/main/java/com/playonlinux/domain/ShortcutSet.java
@@ -18,7 +18,7 @@
 
 package com.playonlinux.domain;
 
-import com.playonlinux.installer.Script;
+import com.playonlinux.installer.ScriptFactoryDefaultImplementation;
 import com.playonlinux.utils.ObservableDirectoryFiles;
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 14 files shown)


qparis:
Does the install bouton still work?

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/apps/AppPanel.java
```diff
@@ -18,33 +18,28 @@
...
-import javafx.event.EventHandler;
 import javafx.scene.control.ScrollPane;
-import javafx.scene.layout.FlowPane;
 import javafx.scene.layout.HBox;
 import javafx.scene.layout.VBox;
-import javafx.scene.web.WebEvent;
 import javafx.scene.web.WebView;
+
 import org.apache.log4j.Logger;
+import org.python.apache.html.dom.HTMLAnchorElementImpl;
```
qparis: Why is it needed?

Tutul-: use a sun import is deprecated

pom.xml
```diff
@@ -118,8 +118,6 @@
         </dependency>
     </dependencies>
 
-
-
     <build>
         <resources>
             <resource>
```
Tutul-: Sorry for the change, test some dependencies for the import but delete all my test

PR #121, opened 2015-06-27
Correct some Java convention

Tutul-:
like "private static final Logger LOGGER = Logger.getLogger(<className>.class);" !

(not important or urgent so take your time)


yawkat:
Loggers are actually often exempt to capitalization :) I guess it doesn't make too much of a difference though.

Tutul-:
it's just a convention, if a declaration is static and final (http://docs.oracle.com/javase/tutorial/java/nutsandbolts/variables.html) but it's only the appearance, no performance impact so that is why it is neither important nor urgent need :)

qparis:
In general, we do not code "just to please" sonarqube 

There must be a valid reason that will improve the code quality.

Tutul-:
for me, respect the langage convention is a bit important. But you decide :)

qparis:
The convention is not defined at a language level but at a project level

qparis:
(But as long as we keep the same convention (ie LOGGER) for the whole project, it is fine)

PR #128, opened 2015-07-04
[MainWindow] add icon on gnome menubar

Tutul-:
Don't merge now, I will try to correct name into menubar too (see https://github.com/PlayOnLinux/POL-POM-5/issues/107)


Tutul-:
No solution for the name so far so merge this and I will open a new Pull Request when solve :)

PR #134, opened 2015-07-07
[StepRepresentationLicence] Modify disposition

Tutul-:
+ Add missing properties for POL

It's ok now ? :)
capture d ecran de 2015-07-08 00 41 25

### DIFF
--- a/src/main/java/com/playonlinux/ui/impl/javafx/setupwindow/StepRepresentationLicence.java
+++ b/src/main/java/com/playonlinux/ui/impl/javafx/setupwindow/StepRepresentationLicence.java
@@ -19,40 +19,33 @@
 package com.playonlinux.ui.impl.javafx.setupwindow;
 
 import static com.playonlinux.lang.Localisation.translate;
-import javafx.scene.control.CheckBox;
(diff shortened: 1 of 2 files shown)


qparis:
No, please have a look at TextField. It is not presented this way

Tutul-:
capture d ecran de 2015-07-08 01 01 08

src/main/java/com/playonlinux/ui/impl/javafx/setupwindow/StepRepresentationLicence.java
```diff
@@ -73,7 +65,7 @@ protected void drawStepContent() {
     @Override
     protected void setStepEvents() {
```
qparis: You do not need to overwrite this method, do you?

Tutul-:
It's better to put it for all methode from interface or other parent

qparis:
No it is not. This is the whole point of inheritance: avoid duplicate code

Tutul-:
I can remove it but why keep it on other StepRepresentation*

qparis:
Because it needs to be defined at least once

PR #139, opened 2015-07-11
Use the same Executor for all ScriptFactory

Tutul-:


src/test/java/com/playonlinux/domain/LegacyWrapperTest.java
```diff
@@ -62,7 +64,9 @@ public void testLegacyWrapper() throws Exception {
             return mockFuture;
         });
 
-        Script testScriptWrapper = new ScriptFactoryDefaultImplementation()
```
Tutul-: Not working actually, try to fix it but no result :/

PR #187, opened 2015-07-30
Outsourcing History and saving cursorPosition.

seijikun:
The History is now handled in the small helper class `CommandHistory`.
Now not only the command itself, but also the cursorPosition when hitting enter is saved and restored.

The `CommandHistory` helper should probably be moved out of impl.javafx, since it can at least be shared with the gtk gui.

### DIFF
--- a/src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/console/CommandHistory.java
+++ b/src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/console/CommandHistory.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2015 Markus Ebner
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package com.playonlinux.ui.impl.javafx.mainwindow.console;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Small helper intended for keeping track of commands that have been entered into the console window.
+ */
+public class CommandHistory {
+
+    private List<Item> history = new ArrayList<>();
+    private int historyPosition = 0;
+
+    public void add(Item item){
+        this.history.add(item);
+        historyPosition = this.history.size();
+    }
+
+    public Item up(){
+        if(historyPosition > 0){
+            historyPosition--;
+        }
+        return current();
+    }
+
+    public Item down(){
+        historyPosition = (historyPosition < history.size()) ? (historyPosition + 1) : history.size();
+        return current();
+    }
+
+    public Item current(){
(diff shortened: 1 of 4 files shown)


src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/console/CommandHistory.java
```diff
@@ -0,0 +1,74 @@
...
+
+package com.playonlinux.ui.impl.javafx.mainwindow.console;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Small helper intended for keeping track of commands that have been entered into the console window.
+ */
+public class CommandHistory {
```
qparis: This class does not seem to be specific of javafx. Can it be moved?

qparis: Can we test this class?

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/console/CommandHistory.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private List<Item> history = new ArrayList<>();
```
qparis: Can you mark this field final?

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/console/CommandHistory.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public Item current(){
+        if(history.size() > 0 && historyPosition < history.size()){
+            return history.get(historyPosition);
+        }
+        return Item.empty;
+    }
+
+
+    public static class Item {
```
qparis: Is it immutable? If yes can you protect fhe fields?

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/console/CommandHistory.java
```diff
@@ -0,0 +1,74 @@
...
+
+        public static Item empty = new Item("", 0);
```
qparis: This can be static final

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/console/CommandHistory.java
seijikun: `CommandHistory` can be moved, yes (last sentence of the pull-request message). If you tell me where to move it, I will do so and add the tests.

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/console/ConsoleTab.java
```diff
@@ -102,21 +101,22 @@ public ConsoleTab() throws CommandInterpreterException {
...
-                historyIndex++;
-                if (historyIndex == commandHistory.size()) {
-                    command.setText("");
-                } else if (historyIndex < commandHistory.size()) {
-                    command.setText(commandHistory.get(historyIndex));
-                } else {
-                    historyIndex = commandHistory.size();
-                }
+                CommandHistory.Item historyItem = commandHistory.down();
+                Platform.runLater(() -> {
```
qparis: Why do you need to use runLater?

seijikun: Without runLater, the TextField's cursor is not correctly set.

qparis: Even if you use the released event?

src/main/java/com/playonlinux/ui/impl/javafx/mainwindow/console/CommandHistory.java
qparis: com.playonlinux.ui.common for the moment should be fine

qparis:
ok to test

PR #194, opened 2015-08-05
[Qt] First implementation, still mostly unfunctional

seijikun:
A first implementation of a Qt-Gui for POL which is still mostly unfunctional.


qparis:
ok to test

src/main/java/com/playonlinux/ui/impl/qt/common/QMemoryStream.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public QMemoryStream(InputStream innerStream){
+        this.innerStream = innerStream;
+    }
+
+    @Override
+    public int readData(byte[] bytes) {
+        try{
+            return innerStream.read(bytes);
+        }catch(IOException e){
```
qparis: Can you set your code formatter so that it follows the following style:

try {
    return innerStream.read(bytes);
} catch(IOException e) {
...

seijikun:
A screenshot of the UI:
Screenshot

qparis:
Amazing!

PR #200, opened 2015-08-11
[Qt] Split MainWindow into components, don't use autogenerated code

seijikun:
The autogenerated code does not have too many benefits. Splitting the single components up and combining them within the MainWindow is probably the best way.


qparis:
ok to test

PR #212, opened 2015-08-20
Fix for external dependencies installation in mvn clean goal

mariok:
Make sure both dependencies get installed when executing mvn clean. Before, only the qtjambi dependency would be installed.


qparis:
ok to test

qparis:
ok to test

qparis:
Thank you :)

PR #226, opened 2015-09-08
[Qt] Implement ActionSidebar and various restructuring

seijikun:
Implemented the ActionSideBar.
Moved icons that are shared across components into the resource-folder of the ResourceHelper.
Make use of qss for simple styling.


qparis:
ok to test

PR #228, opened 2015-09-10
Fix compilation issue

qparis:


qparis:
ok to test

qparis:
ok to test

PR #257, opened 2015-10-01
Remove the exit confirmation message

RobLoach:
Just quit the application, and all PlayOnLinux windows, automatically. We don't really need a confirmation.

### DIFF
--- a/src/main/java/com/playonlinux/ui/impl/qt/mainwindow/MainWindowEventHandler.java
+++ b/src/main/java/com/playonlinux/ui/impl/qt/mainwindow/MainWindowEventHandler.java
@@ -78,19 +78,6 @@ public class MainWindowEventHandler implements UIEventHandler {
      * @return False when the user aborted the application exiting, True otherwise.
      */
     public boolean exit() {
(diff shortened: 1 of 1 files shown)


qparis:
In the current behavior, PlayOnLinux will exit all wine application when closed.
I would at least put a parameter to keep this option

RobLoach:
As an option in the PlayOnLinux settings?

qparis:
Maybe in the property file in a first step?

RobLoach:
Not quite sure where that would go. Do you have an example of using a property?

qparis:
You can use PlayOnLinuxContext as done is this class: https://github.com/PlayOnLinux/POL-POM-5/blob/6d4e9b5aa1b07e222a7a33354a50f88f25afacf3/src/main/java/com/playonlinux/core/log/ScriptLogger.java

Tutul-:
Maybe we can ask only if wine application is runnig. No popup when POL can close normally :)

samrocketman:
I agree the behavior is annoying.

PR #260, opened 2015-11-09
Theme-switch-fx: Implement a way to switch themes in JavaFX client

Veske:
* Create the settings menu
   * Add new combobox to settings menu that can switch JavaFx themes.

Themes can be changed on the fly. Change can be seen here: http://oi66.tinypic.com/2wogab8.jpg

Was not sure about the current config manager implementation so for now the changes are not persisted.

Also as I do not own a eye for beautiful design, I did not change the dark theme much.


qparis:
ok to test

Veske:
Updated the theme too a bit. Could really use a proffessional eye tough. I am not much of a designer.

poldarktheme

PR #279, opened 2015-11-17
Fix CRLF problem

### LINKED ISSUE #278: [Test Failure] ScriptLegacyTestFailure
Fails because of \r\n not anymore present in files

Tomahna:
Should Fix #278


qparis:
Yeah indeed it fixes the test but it just workaround the problem. We really need to test that we can safely extract signature from CLRF scripts, it is important. 
Also, can you set up your code formatter so that the long code stay readable? :)

PR #281, opened 2015-11-17
[Warning Correction] Resource Leak

Tomahna:
Correct a bunch of resource Leak warnings. Most of them was in test files.

The recommended way of handling autocloseable resources is through try with resource since Java 7. Might be a good idea to enforce this in the futur.


qparis:
Can you set up your IDE to use spaces instead of tabs?

Tomahna:
Yeah you're right sonar don't like tabs. Should be good.

qparis:
ok to test

PR #282, opened 2015-11-18
Package renames

Tomahna:
Renaming of packages.

For readability packages should indicate in which subproject they are located and be as concise as possible. Thus : 
com.playonlinux.ui.impl.javafx.mainwindow.container -> com.playonlinux.javafx.container
com.playonlinux.ui.impl.qt -> com.playonlinux.qt
com.playonlinux.ui.impl.javafx -> com.playonlinux.javafx
com.playonlinux.core.scripts -> com.playonlinux.scripts
com.playonlinux.core.injection -> com.playonlinux.injection

Renaming Subproject phoenicis-bash-support in phoenicis-scripts

phoenicis-wine should have com.playonlinux.wine as base package but it would cause it to have a com.playonlinux.wine.wine subpackage. 
phoenicis-core should have com.playonlinux.core as base package but it would cause it to have a com.playonlinux.core.core subpackage


qparis:
ok to test

PR #283, opened 2015-11-21
Fix warning serial version uid

Tomahna:
Fixing Missing SerialVersionUID Warnings by adding generated SerialVersionUID

### DIFF
--- a/phoenicis-core/src/main/java/com/playonlinux/app/PlayOnLinuxException.java
+++ b/phoenicis-core/src/main/java/com/playonlinux/app/PlayOnLinuxException.java
@@ -22,7 +22,7 @@ package com.playonlinux.app;
  * PlayOnLinux exception main class
  */
 public class PlayOnLinuxException extends Exception {
-    private static final long serialVersionUID = 5585485206915024227L;
+    private static final long serialVersionUID = 1L;
 
     public PlayOnLinuxException(String message) {
         this(message, null);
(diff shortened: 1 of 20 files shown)


qparis:
Is it really useful for exceptions? It makes the code harder to understand.

Tomahna:
The exception is activated by default in eclipse and is corrected in one line. In addition it concerns mostly exceptions which are really simple class.

So in my opinion it is far worse to let this warning be as it might distract developers from important warnings.

qparis:
I don't have any warnings afaik. I don't see why it would be required as exceptions are not serializable. It would also distract developers to have useless code.

Tomahna:
According to the [javadoc](http://docs.oracle.com/javase/8/docs/api/java/lang/Exception.html), exception are serializable.

I don't know what the default IntelliJ warning profile is, but atm i'm using the default eclipse configuration and these warnings are enabled. It represent 20 warnings, 19 of them are Exceptions. To solve this warning there are 3 solutions : 
- @SuppressWarnings : which should of course be avoided
- Ask developers using eclipse to disable this warning, which is a lot of work for just that
- Add 20 lines of code in the whole project, 19 of them being in overly simple classes (exceptions)

I'm of course for the third solution as the benefits (not having warnings and no efforts for new devs) surpass very largely the cost (1 line added in a real class, 19 in exceptions).

Veske:
I think it's fine. Exception classes are not the ones new devs would go and read into anyway.

qparis:
ok to test

PR #286, opened 2015-11-21
Fix unused

Tomahna:
Removing unused variables, imports and methods.

### DIFF
--- a/phoenicis-core/src/main/java/com/playonlinux/apps/DefaultInstallerDownloaderEntityProvider.java
+++ b/phoenicis-core/src/main/java/com/playonlinux/apps/DefaultInstallerDownloaderEntityProvider.java
@@ -22,7 +22,6 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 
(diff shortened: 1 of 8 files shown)


phoenicis-core/src/main/java/com/playonlinux/apps/DefaultInstallerDownloaderEntityProvider.java
```diff
@@ -80,7 +77,7 @@ public void getScript() {
         downloadManager.submit(httpDownloader, bytes -> {
             success(bytes);
             return null;
-        }, e -> {
+        } , e -> {
```
Veske: Is this kind of syntax changing actually necessary?

Tomahna:
I'm mostly using the eclipse format profile named Java Convention, the one i've shared in the repository in the setting folder. The only difference with the default version is that i force him to use space so that sonar does not complain about that.

Besides the fact that i strongly recommend that we all use the same formatter, as it helps to limit the amount of changes in git due to format. I have no opinion on which format is the best so if you want to make modifications, feel free to do it.

PR #287, opened 2015-11-21
Fix raw type

Tomahna:
Fix some of the raw type warnings. Mostly just adding a missing type parameter.


qparis:
InstallerSource can have a populate() method, 
WebService can have a populate() method

I don't see any problem here

qparis:
ok to test

PR #290, opened 2015-11-29
[Qt] Add File-Extension filter to runLocalScript-dialog

seijikun:
Add a fileExtension-filter to the dialog for running local scripts.


qparis:
ok to test

PR #295, opened 2015-12-22
Java 8 update

Tomahna:
Changing a bunch of methods by using streams instead of for/if loops

Replacing custom Filter interface by java.util.Predicate

Returning a List<File> instead of File[] when possible


phoenicis-core/src/main/java/com/playonlinux/containers/DefaultContainersManager.java
```diff
@@ -50,17 +50,18 @@
...
+    public List<Container<? extends ContainerEntity>> getContainers() {
         return new ArrayList<>(containers);
     }
 
-    public void update(File[] files) {
+    public void update(List<File> files) {
         containers.clear();
+        
         for(File file: files) {
-            final ConfigFile containerConfigFile = new CompatibleConfigFileFormat(new File(file, "playonlinux.cfg"));
```
qparis: Please keep those final

phoenicis-core/src/main/java/com/playonlinux/containers/DefaultContainersManager.java
```diff
@@ -80,10 +81,8 @@ public void shutdown() {
 
     @Override
     public void init() throws ServiceInitializationException {
-        final File containersDirectory = playOnLinuxContext.makeContainersPath();
+        File containersDirectory = playOnLinuxContext.makeContainersPath();
```
qparis: Please keep those final

phoenicis-core/src/test/java/com/playonlinux/filesystem/DirectoryWatcherFilesTest.java
```diff
@@ -63,13 +65,13 @@ public void testObservableDirectory_dontChangeAnything_ObservableOnlyNotifiedOnc
 
 
         try (DirectoryWatcherFiles directoryWatcherFiles = new DirectoryWatcherFiles(mockExecutorService, temporaryDirectory, CHECK_INTERVAL)) {
-            final Consumer<File[]> mockConsumer = mock(Consumer.class);
+            Consumer<List<File>> mockConsumer = mock(Consumer.class);
```
qparis: Please keep those final

Tomahna:
The keyword final is useless in all of these examples. I'll put it back but final should still be used only when necessary.

qparis:
It is a matter of convention. Please keep the same conventions
In POL code, we've chosen to always use final, unless we need to modify the fields later. 

http://stackoverflow.com/questions/154314/when-should-one-use-final-for-method-parameters-and-local-variables

qparis:
ok to test

Veske:
Looks good, streams are nice.

PR #296, opened 2015-12-22
HTTPDownloader as Consumer

Tomahna:
Use consumer instead of observer in HTTPDownloader.

It has impact on ProgressStateEntity and classes using it. ProgressStateEntity was renamed ProgressEntity and internal enum state was externalized in ProgressState.

### DIFF
--- a/phoenicis-core/src/main/java/com/playonlinux/core/webservice/HTTPDownloader.java
+++ b/phoenicis-core/src/main/java/com/playonlinux/core/webservice/HTTPDownloader.java
@@ -36,7 +36,6 @@ public class HTTPDownloader extends ObservableDefaultImplementation<ProgressStat
 
     private static final int BLOCK_SIZE = 1024;
     private final URL url;
-    private State state;
     private float percentage;
 
     public enum State {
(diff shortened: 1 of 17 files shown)


phoenicis-core/src/main/java/com/playonlinux/core/webservice/HTTPDownloader.java
```diff
@@ -27,106 +27,99 @@
...
+                BufferedOutputStream bufferedOutputStream = new BufferedOutputStream(outputStream, BLOCK_SIZE)) {
+            byte[] data = new byte[BLOCK_SIZE];
+            int i;
+            float totalDataRead = 0.0F;
+            while ((i = inputStream.read(data, 0, BLOCK_SIZE)) >= 0) {
+                totalDataRead += i;
+                bufferedOutputStream.write(data, 0, i);
+
+                this.percentage = totalDataRead * 100 / fileSize;
+                changeState(ProgressState.PROGRESSING);
```
qparis: I think percentage could also be passed to changeState, the same way you've done for the state

qparis:
ok to test

PR #306, opened 2015-12-24
Filterable

Tomahna:
Merging Filterable to its only implementer EntitiesProvider

Renaming applyFilter in filter to follow Stream.filter convention


qparis:
ok to test

qparis:
ok to test

PR #308, opened 2015-12-24
ArchiveException as Runtime

### LINKED ISSUE #301: Remove checked exception
At the moment, we are using a lot of home made checked exception: 
* ScriptFailureException
* ArchiveException
* SignatureException
.... 

These exception usually catches and rethrows Java exceptions like IOException, which make the code heavier to read. 
We should replace those exception with unckecked exceptions like IllegalStateException, IllegalArgumentException, etc..

Tomahna:
Starting to fix #301


qparis:
ok to test

PR #311, opened 2015-12-24
Warnings

Tomahna:
Mostly removing not throwned exception declaration
Correcting type of DownloadEnvelop in DefaultWineVersionsManager


qparis:
ok to test

PR #317, opened 2016-02-23
Core file system

Tomahna:
Using java.nio.file.WatchService instead of custom watcher


phoenicis-wine/src/main/java/com/playonlinux/framework/Wine.java
```diff
@@ -639,10 +682,10 @@ public Wine deletePrefix() throws CancelException {
             final long endSize = 0L;
 
             try (DirectoryWatcherSize observableDirectorySize = new DirectoryWatcherSize(executorService,
-                    prefix.getWinePrefixDirectory(), 10)) {
+                    prefix.getWinePrefixDirectory().toPath())) {
```
qparis: Try to keep the same interface as possible. String is not a good type to represent pathes. File is much better

Tomahna:
I'm not using Strings, i'm using java.nio.Path.

qparis:
ok to test

PR #320, opened 2016-02-24
Removing Name, Nameable and AlphabeticalOrderComparator

Tomahna:
AlphabeticalOrderComparator was never used.
Nameable was only used to enforce the presence of a getName method which pointlessly make the type hierarchy heavier
NameableEntity, same as Nameable, was only used as a cross interface between Nameable and Entity.


qparis:
ok to test

Issue #321, opened 2016-02-26
[Code Quality] Use ResourceBundles for internationalization


Tomahna:
Use [ResourceBundle](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/ResourceBundle.html) instead of custom implementation (cf phoenicis-core/com.playonlinux.core.lang) which, in addition of relying on static state, makes translation and overall code more complex.

PR #322, opened 2016-02-26
I18n

Tomahna:
Use ResourceBundle in phoenicis-wine
See #321


phoenicis-wine/src/main/java/com/playonlinux/framework/Wine.java
```diff
@@ -69,12 +70,46 @@
 @ScriptClass
 public class Wine implements SetupWizardComponent {
     private static final Logger LOGGER = Logger.getLogger(Wine.class);
-    private static final Architecture DEFAULT_ARCHITECTURE = Architecture.I386;
+
+    private static final String EXCEPTION_PREFIX_INCOMPATIBLE_ARCH = "A 32bit wineprefix cannot execute 64bits executables!"; //$NON-NLS-1$
```
qparis: I'm not a fan of using constant for log strings. It makes the code really hard to read.

Tomahna:
I agree that the big series of constant definitions make the class looks scary, but the problem is more about the class itself which is probably too big. 
Whether it is more readable to read constants or string is subjective and i'm pretty much unopinionated on that matter.
But if i revert theses changes i'll have a lot of //$NON-NLS-1$ inside the code which might not help readability.

qparis:
Agreed but I still think that string constant should not be used for log messages

Tomahna:
Removed constants for log messages and exception messages

qparis:
ok to test

Veske:
Merge this one also?

qparis:
The PR needs to be adapted with the new injection system

Issue #262, opened 2015-11-13
[Suggestion] Use lombok


Tomahna:
https://projectlombok.org/

Using Lombok would be a good idea to reduce the amount boilerplate code.

Veske:
Looks like too much magic in my opinion.

Tomahna:
I kinda doubt that the high verbosity of Java was the reason of this choice. I guess it was because it's cross platform, typed and well spread.

Reducing the amount of code like this usually improve it's readability and overall quality. But i guess it can disturb unused devs.

Anyway it's not a choice for me to make, that's just a suggestion. You can close this issue if you consider it's not adapted to your project.

qparis:
Could it be easily integrated to maven? 
I think we should think about dividing PlayOnLinux into submodules

Veske:
Do you want to use maven module system for it or wait for Jigsaw project?

qparis:
I guess we can use maven for the moment

Tomahna:
That would definitely help a lot for code comprehension and unit testing.

PR #323, opened 2016-02-26
Lombok

Tomahna:
#262

This remove ~500-600 lines of boilerplate code. I've not optimized absolutely everything so we might still gain a few lines.


Veske:
Needs conflicts solved

PR #330, opened 2016-11-18
Ensure we are running Java 8

ProfessorKaos64:
- Ensure path of Java 8 takes precedence on systems with Java 7 and 8 both installed
- This is non-destructive and does not force users to set the default java env to Java


Veske:
I would merge it. Havent seen admins around for a while tough.

Veske:
Well, there you have it ;)

PR #340, opened 2016-12-26
use phoenicis-javafx as default assembly

plata:


phoenicis-dist/src/assembly/distribution.xml
```diff
@@ -22,7 +22,7 @@
 
 			<!-- Now, select which projects to include in this module-set. -->
 			<includes>
-				<include>com.playonlinux:phoenicis-app</include>
+				<include>com.playonlinux:phoenicis-javafx</include>
```
qparis: Can you also include phoenicis-cli?

plata: If I do that, I get:

`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.6:single (distribution) on project phoenicis-dist: Failed to create assembly: Artifact: com.playonlinux:phoenicis-cli:jar:5.0-SNAPSHOT (included by module) does not have an artifact with a file. Please ensure the package phase is run before the assembly is generated.`

qparis: Ok you can remove it for now :)

PR #349, opened 2016-12-27
inject theme CSS

plata:


phoenicis-javafx/src/main/java/com/playonlinux/javafx/controller/ControllerConfiguration.java
```diff
@@ -36,8 +37,9 @@
     private ContainersConfiguration containersConfiguration;
 
     @Bean
-    public MainController mainController() {
+    public MainController mainController(@Value("${application.theme:defaultTheme.css}") String theme) {
```
qparis: Just to be consistant with the rest of the project, can you create a private field theme instead?

plata: Done.

PR #350, opened 2016-12-27
enabled user specific settings

plata:
Properties stored in the file given by application.user.settings will
override the system settings (e.g. from Linux.properties).


qparis:
ok to test

phoenicis-configuration/src/main/resources/Linux.properties
```diff
@@ -28,6 +28,7 @@ application.user.engines.wine               =           ${application.user.engin
 application.user.engines.wine.gecko         =           ${application.user.engines}/wine/gecko/
 application.user.engines.wine.mono          =           ${application.user.engines}/wine/mono/
 application.user.resources                  =           ${application.user.root}/resources/
+application.user.settings                   =           ${application.user.root}/config.properties
```
qparis: Does it prevent the program from running is  ${application.user.root} does not exist?

plata: Good point! I fixed that. It will now run with or without ${application.user.settings} set and also if the specified file does not exist.

Issue #345, opened 2016-12-26
Highlight currently selected tab in JavaFX


plata:
It is not possible to see which tab (library, apps, containers, engines, settings) you have currently selected.

PR #354, opened 2016-12-28
use TabPane for menu

plata:
fixes: Highlight currently selected tab in JavaFX #345


qparis:
Does it work with the default CSS?

plata:
The contrast is not that great, but...

defaultTheme

Issue #355, opened 2016-12-28
Fix scaling of content when window is scaled


plata:
![grafik](https://cloud.githubusercontent.com/assets/3973260/21522919/eb02c5e2-cd09-11e6-9cef-b81d6990cad7.png)

PR #358, opened 2016-12-28
replaced header with TabPane, fixed scaling of content

plata:
tabs

This pull request  also fixes: Fix scaling of content when window is scaled #355


qparis:
Can you show with the default theme please ? :)

plata:
defaultTheme

qparis:
Thanks :)

We need to leave the default (gray) tabbar style here, it does not look good otherwise 

Can you add a class to the header tab pane ?

plata:
You mean for the "My applications" TabPane?

qparis:
No the other one, so that the look of the others tab panes (include the "My Application" one) does not change

plata:
Ok. That was what I meant ;)

qparis:
Even in the CSS file?

plata:
Well, I don't really now how to set e.g. `.tab-header-background`. I would like to set the style for the complete pane but I don't think that's possible.

qparis:
Let me try

qparis:
Fixed in pull request #361

plata:
Weird. I had tried pretty much the same...

PR #361, opened 2016-12-28
Fixed CSS

qparis:


plata:
Can you also fix the darkTheme?

Issue #386, opened 2016-12-29
[Easy] Write unittest for TeeApplicationsSource


plata:
See: Implement TeeApplicationsSource #332

The test shall show that:

- categories which occur in only one of the two ApplicationSource occur also in the merged result
- if an application occurs in both ApplicationSource, the one from the rightApplicationSource
- merged applications should contain resources, scripts and miniatures from both ApplicationSource

plata:
@qparis can this be closed?

qparis:
So far yes

PR #388, opened 2016-12-29
TeeApplicationSource

qparis:


qparis:
@plata What do you think of this implementation? 
Tests remain to be written though

qparis:
There are few advantages : 
* The DTOs can be built really easily, and so the sources
* The DTOs are truely immutable, which is safer (especially concerning the scirpts)
* The data is serializable and unserializable, a webservice could easily implement a source
* Resources and scripts are properly merged

phoenicis-apps/src/test/java/com/playonlinux/apps/TeeApplicationsSourceTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+        final ApplicationsSource teeSource = new TeeApplicationsSource(leftSource, rightSource);
+        assertEquals(4, teeSource.fetchInstallableApplications().size());
+    }
+
+
+    @Test
+    public void testFetchInstallableApplications_categoriesAndAppsCollapse_numberOfResultIsCorrect() {
+        final ApplicationsSource leftSource = () -> Arrays.asList(
+                new CategoryDTO.Builder()
```
plata: categories 1, 4, 5 are not used in this test and can be removed

phoenicis-apps/src/main/java/com/playonlinux/apps/TeeApplicationsSource.java
```diff
@@ -22,20 +25,94 @@ public TeeApplicationsSource(ApplicationsSource leftApplicationSource,
...
+        applications.sort(ApplicationDTO.nameComparator());
+        return new CategoryDTO.Builder()
+                .withApplications(applications)
+                .withType(leftCategory.getType())
+                .withIcon(leftCategory.getIcon())
+                .withName(leftCategory.getName())
+                .build();
+    }
+
+    private ApplicationDTO mergeApplications(ApplicationDTO leftApplication,
```
plata: Could you enhance the unit test to check the merge of miniatures and DTOs?

phoenicis-apps/src/main/java/com/playonlinux/apps/dto/ApplicationDTO.java
```diff
@@ -70,14 +71,18 @@ public String getDescription() {
         return scripts;
     }
 
+    public static Comparator<ApplicationDTO> nameComparator() {
```
plata: This is the same for all DTO's. Can it be merged?

phoenicis-apps/src/main/java/com/playonlinux/apps/TeeApplicationsSource.java
qparis: Will do in the issue #386

phoenicis-apps/src/main/java/com/playonlinux/apps/dto/ApplicationDTO.java
qparis: Yes I guess it could be

phoenicis-apps/src/test/java/com/playonlinux/apps/TeeApplicationsSourceTest.java
qparis: Will do as part of the issue #386

qparis:
Yeah, but it behaves like debian repositories for example. I agree that other sources should be sorted.

plata:
Ok. We will found out anyways. If it turns out to be to confusing, we can still revert it with no problem.

qparis:
Or implement another sort of TeeSource that behaves differently. I think that this system is quite flexible anyway

phoenicis-apps/src/main/java/com/playonlinux/apps/dto/ApplicationDTO.java
qparis: Hum they are not exactly the same. If I want to merge them, I need to create something like a NameableDTO interface. Not sure it worth it

plata: Just an idea. If it is more effort and introduced complexity, leave it as it is.

PR #412, opened 2017-01-01
implemented FileUtilities.getSize()

plata:


qparis:
ok to test

phoenicis-tools/src/main/java/com/playonlinux/tools/files/FileUtilities.java
```diff
@@ -80,6 +82,14 @@ public String getFileContent(File file) throws IOException {
         return FileUtils.readFileToString(file, "UTF-8");
     }
 
+    public long getSize(File file) throws IOException {
+        Path folder = Paths.get(file.getAbsolutePath());
```
qparis: Can you add assertInDirectory(file); ?

plata: Wouldn't it be better to use `isDirectory()` instead of an assertion?

qparis: assertI**n**Directory will throw an exception if the file is not in PlayOnLinux root

qparis: It is just some security measure to avoid what we had with POLv4

plata: Okay.

qparis: By the way, does your code works with recursive directories?

phoenicis-tools/src/main/java/com/playonlinux/tools/files/FileUtilities.java
```diff
@@ -80,6 +82,16 @@ public String getFileContent(File file) throws IOException {
         return FileUtils.readFileToString(file, "UTF-8");
     }
 
+    public long getSize(File file) throws IOException {
+        assertInDirectory(file);
+
+        Path folder = Paths.get(file.getAbsolutePath());
+        return Files.walk(folder)
```
qparis: I think you can directly use Files.size(file), can't you?

plata: I followed this: <http://www.baeldung.com/java-folder-size>

qparis: So far so good

qparis:
Thanks :-)

Issue #434, opened 2017-01-06
JavaFx: SetupWizard doesn't scale


plata:

qparis:
This will be very hard to scale it because of the left image

plata:
Do we really need the left image? It doesn't add any value in my opinion.

qparis:
Yes I think we do. It allows the customize the scripts, and we have a lot of already made left images/

Also, if you remove it, the presentation screen will feel very empty

plata:
To be honest, creating the left image to me always was one of the most annoying things with PoL 4. Moreover, I believe that not using custom left images would make installing applications feel more like a PoL install rather than PoL calling some script.

I agree with the empty feeling though. But if we use the current playonlinux image all the time, we could replace the image by some JavaFX Text.

qparis:
All right, I agree on that

Issue #399, opened 2016-12-31
Setup wizard cannot handle long titles


plata:
The setup wizard must insert a line break in long titles (e.g. Dr. Langeskov, The Tiger, and The Terribly Cursed Emerald: A Whirlwind Heist).

PR #438, opened 2017-01-07
Scale SetupWizard

plata:
fixes:
- JavaFx: SetupWizard doesn't scale #434
- Setup wizard cannot handle long titles #399

### DIFF
--- a/phoenicis-apps/src/main/java/com/playonlinux/apps/LocalApplicationsSource.java
+++ b/phoenicis-apps/src/main/java/com/playonlinux/apps/LocalApplicationsSource.java
@@ -83,6 +83,7 @@ class LocalApplicationsSource implements ApplicationsSource {
             }
         }
 
+        Collections.sort(results, Comparator.comparing(CategoryDTO::getName));
         return results;
     }
 
@@ -121,6 +122,7 @@ class LocalApplicationsSource implements ApplicationsSource {
             }
         }
 
+        Collections.sort(results, Comparator.comparing(ApplicationDTO::getName));
         return results;
     }
 
--- a/phoenicis-cli/src/main/java/com/playonlinux/cli/setupwindow/SetupWindowCLIImplementation.java
+++ b/phoenicis-cli/src/main/java/com/playonlinux/cli/setupwindow/SetupWindowCLIImplementation.java
@@ -32,17 +32,12 @@ class SetupWindowCLIImplementation implements SetupWindow {
     }
 
     @Override
-    public void setLeftImage(File leftImage) throws IOException {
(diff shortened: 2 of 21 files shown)


plata:
Not yet finished. Just wanted to know if this approach is ok.

qparis:
Do you have any screenshot ? :)

plata:
The left image does not scale yet and the buttons are on a fixed position, but you can see that the whole size is adjusted.
scaled

qparis:
The approach is good

plata:
@qparis which font is used for "playonlinux"?

qparis:
maven pro

plata:
scaled

qparis:
Awesome, 
The text needs to be rescaled

plata:
Which text?

qparis:
The presentation text. Does it need to be rescaled

qparis:
Do you want to merge it now or do you want to also test all the steps?

plata:
Like this?

small:
small

large:
large

plata:
I checked several of the steps. Should be ok in most cases.

qparis:
Yep

qparis:
Have you checked browse?

plata:
Yes. I checked browse, wait, menu, textbox, message. Maybe there is a small error hidden somewhere, but in general it should be fine.

Just let me check one small thing before merge: What happens if the window is too small to show the text?

qparis:
Don't know. Also, don't forget to check progressbar ? :)

plata:
No ;). It will change the width but keep the height.

qparis:
Tell me when we can merge

plata:
You can. The text will not be readable anymore if the window is too small. I tried using a ScrollPane but then the text is not wrapped anymore...

qparis:
Why not enforcing a minimal window size?

plata:
No. I know applications which do that. Only causes trouble in special cases (e.g. if the enforced size is greater than your screen resolution). But I just found the property to fix the wrapping. Give me one more try :P

plata:
I have a ScrollPane now. But it's not white...

ScrollPane

plata:
Now:

ScrollPane

plata:
What's wrong with the build check? Works for me...

Brainzyy:
@plata It started failing after this: https://github.com/PlayOnLinux/POL-POM-5/pull/440
(Not sure if this is related)

plata:
@Brainzyy yes, that's the reason. Thank you.

plata:
Can I somehow trigger a new build? (without a new commit)

qparis:
Just merge the master to your PR branch and it should work

Issue #430, opened 2017-01-06
Regression: apps are not sorted


plata:
All ApplicationsSource must sort the result of `fetch`.

PR #439, opened 2017-01-07
sort apps

plata:
fixes: Regression: apps are not sorted #430


qparis:
Shouldn't we create a SortedApplicationSource?

plata:
I don't think that's necessary. We can just define that an ApplicationSource always fetches a sorted list. Unfortunately, this cannot be expressed in Java directly.

PR #440, opened 2017-01-07
Hide scrollbar

qparis:
Hide scrollbar


phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/common/widget/MiniatureListWidget.java
```diff
@@ -55,6 +53,7 @@ private MiniatureListWidget(Pane content) {
 
         this.content.prefWidthProperty().bind(this.widthProperty());
 
+        this.setHbarPolicy(ScrollBarPolicy.NEVER);q
```
Brainzyy: The 'q' after the ';' shouldn't be there. @qparis @plata

PR #441, opened 2017-01-07
Revert "Hide scrollbar"

plata:
Reverts PlayOnLinux/POL-POM-5#440

Doesn't build.


Brainzyy:
@plata Don't revert the whole commit. Just remove the 'q' after the ';' on line 56.

plata:
@Brainzyy I cannot change the pull request. @qparis has to do it.

qparis:
My mistake sorry.

qparis:
Can you just remove the scroll bar as I did before? It is a really little change

plata:
see #442

Issue #449, opened 2017-01-08
Text needs to be wrapped and scaled


Brainzyy:
![bug](https://cloud.githubusercontent.com/assets/8126341/21746323/0738e784-d541-11e6-900a-5bd88dc1226c.png)
You can reproduce this by installing any Steam game.
(In case it isn't clear, I'm talking about the text)

Brainzyy:
Changing bug name as discussed in: https://github.com/PlayOnLinux/Scripts/pull/105

plata:
@Brainzyy should be fixed.

Brainzyy:
@plata It is fixed. But maybe one suggestion. In the message itself, maybe remove the \n so that you can see the full message. Because now you can't see it, no matter the dimensions.

plata:
@Brainzyy the problem is that the scroll pane is not resizing vertically. Can you try to add
```java
VBox.setVgrow(scrollPane, Priority.ALWAYS);
```
to StepRepresentationMessage `drawStepContent()` and see if you prefer that?

Brainzyy:
@plata That is perfect. The ScrollPane is not there anymore and the text scales perfectly with the resolution of the window.

plata:
@Brainzyy there's only one thing I don't like about this: The spinning wheel is no longer centered. But I cannot find a way to fix that.

Can you create a pull request?

Brainzyy:
@plata I can make a PR for the suggestion you made. But I'm not experienced enough with Java to figure out the spinning wheel center thing. I'll open a PR and see what @qparis his opinion is.

PR #454, opened 2017-01-08
wrap text in StepRepresentationMessage

plata:
fixes: Text needs to be wrapped and scaled #449

### DIFF
--- a/phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/setupwindow/StepRepresentationMessage.java
+++ b/phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/setupwindow/StepRepresentationMessage.java
@@ -19,7 +19,12 @@
 package com.playonlinux.javafx.views.setupwindow;
 
 import com.playonlinux.scripts.ui.Message;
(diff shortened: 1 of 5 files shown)


phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/setupwindow/StepRepresentationMessage.java
```diff
@@ -32,7 +37,15 @@ public StepRepresentationMessage(SetupWindowJavaFXImplementation parent, Message
     @Override
     protected void drawStepContent() {
         Text textWidget = new Text(textToShow);
-        this.addToContentPane(textWidget);
+
+        ScrollPane scrollPane = new ScrollPane();
+        scrollPane.setBackground(new Background(new BackgroundFill(Color.WHITE, CornerRadii.EMPTY, Insets.EMPTY)));
+        scrollPane.setStyle("-fx-background: #F4F4F4; -fx-border-color: #F4F4F4;");
```
qparis: This need to be set in the CSS

Issue #470, opened 2017-01-14
App hangs when switching from and to containers


qparis:

qparis:
@plata It seems to be caused by your commit

plata:
What exactly is the problem (how do I reproduce)?

qparis:
Click on the "Container" button. It does not appear directly and the interface lags

plata:
Sorry but I do not have that behavior.

qparis:
You do not have enough containers

plata:
I think it is because WinePrefixesManager `fetchContainers()` accesses the file system to get the container information. Therefore it reads the cfg in all prefixes.

qparis:
And all the registry files yes. This is why it should not be synchroneous

plata:
How is it done for the apps? They have to be fetched as well.

qparis:
This is already asynchroneous (See BackgroundApplicationSource)

plata:
Ok. So the task here is: Implement BackgroundContainersManager for asynchronous fetching of containers (see BackgroundApplicationSource).

qparis:
I guess it is yes

plata:
I started but I'm stuck. Do you mind having a look?

qparis:
No problem

plata:
Think I found it already :P

plata:
@qparis I think this can be merged for now, right?

qparis:
Uzs

qparis:
Yes *

Issue #469, opened 2017-01-14
Container main window doesn't get refreshed after uninstall


Brainzyy:
![This](https://cloud.githubusercontent.com/assets/8126341/21955491/91836b1e-da6c-11e6-9769-b48aace57c3a.png)
instead of:
![this](https://cloud.githubusercontent.com/assets/8126341/21955494/9d600f78-da6c-11e6-99e7-627af2673111.png)

plata:
@Brainzyy does it work if you add
```java
initSelectContainerPane();
```
to `populate()` in ViewContainers.java?

Brainzyy:
@plata That does not work.

plata:
@Brainzyy can you try if it's fixed?

Brainzyy:
@plata Confirmed fixed.

PR #471, opened 2017-01-14
asynchronous update for containers

plata:
fixes: 
- App hangs when switching from and to containers #470
- Container main window doesn't get refreshed after uninstall #469

### DIFF
--- a/phoenicis-containers/src/main/java/com/playonlinux/containers/BackgroundContainersManager.java
+++ b/phoenicis-containers/src/main/java/com/playonlinux/containers/BackgroundContainersManager.java
@@ -1,0 +1,22 @@
+package com.playonlinux.containers;
+
+import com.playonlinux.containers.dto.ContainerDTO;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Consumer;
+
+class BackgroundContainersManager implements ContainersManager {
+    private final ContainersManager delegatedContainersManager;
+    private final ExecutorService executorService;
+
+    BackgroundContainersManager(ContainersManager delegatedContainersManager,
+                                 ExecutorService executorService) {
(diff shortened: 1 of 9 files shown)


qparis:
I still have the problem

plata:
Do you have any explanation? I did the same as for apps but couldn't really test as I do not have the problem.

qparis:
It comes from the fact that fetchContainers() is called multiple times at once. 
However, you configured _containersExecutorService_ so that its queue has a size of one. Therefore, this executor service will lock the sender thread until the queue is released.

Few solutions: 
- Increase the size of the queue
- Investigate why fetchContainers() is called multiple times at once

plata:
Strange. I had outputs in `fetchContainers()` and I got the output only once when PoL was started and then once when I opened the "Containers" tab.
Feel free to resize the queue. I don't know which number might reasonable here.

qparis:
Try to switch quickly from one tab to another

plata:
Ok. Well in this case what's happening is that your fetch is slower than you are switching tabs. How many prefixes do you have? Do you have a SSD?

I don't know if it is possible to fetch only when switching to "Containers" (currently it's also done when leaving "Containers").

qparis:
I have 15 prefixes with potentially large registry files. Yes I do have a SSD

plata:
Is it necessary to read the registry?

qparis:
Yes, if you want to know the current settings value (like GLSL, windows version, etc...)

plata:
But that would be enough once you select a container.

qparis:
Yes but much more complicated to implement. You would need to define two different types of DTO

plata:
True. But like this you're potentially reading a lot of files you never need.

plata:
#473 should at least reduce the load a bit.

PR #473, opened 2017-01-14
update containers only when "Containers" tab is opened

plata:
Before, the update was also done when the "Containers" tab was closed.


qparis:
Understood! loadContainers() is also called on startup. So my UI will also lag if I directly click on "Containers". 
I think we should increase a little bit the size of the queue (up to 5 should be enough)

plata:
Yes. As I said, I cannot really try. Please play around a bit with the queue sizes and adjust it.

Can we cancel the fetch when you leave the "Containers" tab (in case it's still running)?

qparis:
It is really hard to cancel such a task. I think your PR is the best we can do for the moment

PR #475, opened 2017-01-15
Sort engines

plata:


plata:
Please merge #473 first.

Brainzyy:
@plata I think you branched off the wrong branch, because the commits of #473 are in this PR as well.

plata:
@Brainzyy I know... I did #473 on my master yesterday because it's really small and I thought it would be merged immediately. That's why I said that #473 should be merged first.

Brainzyy:
@plata Ah okay, that's fine, just wanted to let you know :)

plata:
@Brainzyy always good to have someone reviewing ;)

qparis:
I don't understand the purpose of this

Can you explain?

plata:
The engines are not sorted. If you look at the tabs you get something like 2.0rc5, 2.0rc3, 2.0rc4. 

If you look at the old code, it's trying to sort the versions (`WineVersionDTO`) by `WineVersionDistributionDTO`. That's not working.

qparis:
Then we need to enhance version comparator I think

plata:
Don't know. For me the simple alphabetical sorting would be enough. At least it's obvious to the user how the sorting is done.

qparis:
If you do so, wine-1.9.10 will be before wine 1.9.9 and not after

plata:
You're right.

I don't know how 2.0-rc4 is handled with regards to intermediate/low number.

qparis:
We can add a unit test asking 2.0-rc4 to be higher than 1.9.10
This is the way it should be done I think

plata:
The problem is that "-" is used to denote a description so "rc4" is seen as description instead low number.

plata:
Should be ok now.

Brainzyy:
@plata The staging engines now sort correctly. But the upstream engines have Wine1.5.27-NeedforSpeed as the first engine, which is not correct.

plata:
@Brainzyy that's because the engine is named 15.27 (the . between 1 and 5 is missing). Therefore, the big version is 15 which is > 2. So the sorting is correct, the mistake is in the engine name.

@qparis can you fix it in <http://phoenicis.playonlinux.com/index.php/wine?os=linux>?

PR #489, opened 2017-01-21
settings: use list for repositories

### LINKED ISSUE #486: Use list for repository parameter in settings view

plata:
fixes #486 

list

add


qparis:
I think we could font awesome icons instead of "Add" or "Remove" text (http://www.jensd.de/wordpress/?cat=25) 

What do you think?

plata:
Don't know. Would have to match the whole look and feel.

qparis:
Yes sure, but that would be a good start :)

PR #492, opened 2017-01-21
fixed highlighting of active tab in menu

plata:
Before, the tab lost its highlighting once it lost focus (e.g. if you clicked any button).

highlighting

### DIFF
--- a/phoenicis-javafx/src/main/resources/com/playonlinux/javafx/views/common/defaultTheme.css
+++ b/phoenicis-javafx/src/main/resources/com/playonlinux/javafx/views/common/defaultTheme.css
@@ -123,9 +123,9 @@
 }
 
 #menuPane > .tab-header-area .tab:selected {
-    -fx-border-width: 2;
-    -fx-border-style: hidden solid hidden solid;
-    -fx-border-color: white;
+    -fx-border-width: 4;
+    -fx-border-style: hidden hidden solid hidden;
+    -fx-border-color: #529bda;
     -fx-focus-color: transparent;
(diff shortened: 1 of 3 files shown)


phoenicis-javafx/src/main/resources/com/playonlinux/javafx/views/common/defaultTheme.css
```diff
@@ -122,6 +122,14 @@
     -fx-max-height: 2.5em;
 }
 
+#menuPane > .tab-header-area .tab:selected {
```
qparis: Can you try 
#menuPane > .tab-header-area .tab:selected {
    -fx-background-color: linear-gradient(#529bda, #2173A5);
    -fx-focus-color: transparent;
    -fx-faint-focus-color: transparent;
}
instead?

plata: Looks like this:
gradient

qparis: Do you like it?

plata: Fine for me. What colors would you use for the dark one?

Or how about something like KDE:
KDE like

Btw: Do you see the ugly frame around the left panel with the dark theme?

qparis: I like your idea. The frame problem has been fixed in one of my last PR

PR #499, opened 2017-01-22
Theme selection

qparis:


phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/common/Themes.java
```diff
@@ -19,24 +19,25 @@
...
 
 public enum Themes {
-
-    DEFAULT("Default theme"),
-    DARK("Dark theme"),
-    BREEZE_DARK("Breeze Dark theme"),
-    HIDPI("HiDPI theme");
+    DEFAULT("Default theme", "default"),
+    DARK("Dark theme", "dark"),
+    BREEZE_DARK("Breeze Dark theme", "breeze"),
```
plata: Please name it breeze_dark. It's very well possible that someone implements the "normal" Breeze.

qparis: Done. I think we should automatically detect installed skins by the way

phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/mainwindow/settings/ViewSettings.java
```diff
@@ -125,15 +125,9 @@ public String fromString(String string) {
...
-        if (settings.get(Setting.THEME).equals("dark")) {
-            themes.setValue(Themes.DARK);
-        } else if (settings.get(Setting.THEME).equals("breeze")) {
-            themes.setValue(Themes.BREEZE_DARK);
-        } else if (settings.get(Setting.THEME).equals("hidpi")) {
-            themes.setValue(Themes.HIDPI);
-        } else {
-            themes.setValue(Themes.DEFAULT);
-        }
+        final Themes setTheme = Themes.fromShortName(settings.get(Setting.THEME));
```
plata: Could we use enum values instead of Strings? e.g.
```java
settings.getTheme(Setting.THEME)
```
which would return an enum.

Could also be useful for other data types (e.g. `settings.getDouble()`).

qparis: It is really hard to detect what kind of data we have because the view directly queries the model.
We should use a DTO pattern instead

PR #507, opened 2017-01-22
Breeze Dark: dark background for installation

### LINKED ISSUE #496: Small white line below left image in step representation
white line

plata:
fixes #496

Somehow, I have problems changing the text color. Could need some help there.
text-color

### DIFF
--- a/phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/setupwindow/StepRepresentationMessage.java
+++ b/phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/setupwindow/StepRepresentationMessage.java
@@ -37,6 +37,7 @@ public class StepRepresentationMessage extends AbstractStepRepresentationWithHea
     @Override
     protected void drawStepContent() {
         Text textWidget = new Text(textToShow);
+        textWidget.setId("stepText");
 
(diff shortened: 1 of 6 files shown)


qparis:
Does the text has a className?

plata:
No. I would think that it should use the default text color of `.text {` but it doesn't.

qparis:
Well you may try to add a class then

plata:
Don't know if it's a good idea to have so many classes. In the end, you have to fine tune everything when you just want to change the text color...

qparis:
You can add -fx-fill: #FFF; to the root pane I guess then

plata:
I tried that already. Doesn't work.

qparis:
I think we could replace
 
.leftPaneCheckBox .text {
    -fx-fill: #eff0f1;
}

with

.text {
    -fx-fill: #eff0f1;
}

And then, add the text class in StepRepresentationMessage

plata:
Ok. I think we should consider specifying colors and fonts on a higher level than the individual node in general. Currently, there are also different fonts used throughout the application etc.

qparis:
Oops forgot to squash

plata:
That's why I said change the settings ;)

PR #508, opened 2017-01-22
use plain image for app description

plata:
It doesn't make sense to use the MiniatureListWidget.Element because
- there is no text
- there is no tooltip
- you cannot click it, so hover is useless
- the hover highlight looks strange because it also highlights the part where the text would be usually

description image


qparis:
Screenshot?

plata:
Updated with screenshot.

PR #509, opened 2017-01-22
implemented fallback to default theme

plata:


qparis:
We need to have a SkinManager component if we want to handle this properly

plata:
Well at least it does not crash if you have a non existing theme in your user settings. Instead it shows an error message and falls back to the default theme.

qparis:
It will still crash when the application description is shown which is even worse because you won't understand why.
This is why I suggest to create a bean responsible of providing the theme resources

plata:
#510

PR #534, opened 2017-01-28
added Unity theme

### LINKED ISSUE #529: Ubuntu theme needed
Incorrect theming of the scrollbar on the right, it should look like this (in Ubuntu's default theme):
screenshot from 2017-01-23 18-03-42

not this:
screenshot from 2017-01-23 18-04-14

Incorrect theming of closing screen, should look like this (POL 4):
screenshot from 2017-01-24 10-23-04

not this (POL 5):
screenshot from 2017-01-24 10-23-11

plata:
fixes #529 

library

apps

containers

engines

settings

@Ads20000 what do you think?

### DIFF
--- a/phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/common/Themes.java
+++ b/phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/common/Themes.java
@@ -22,6 +22,7 @@ public enum Themes {
     DEFAULT("Default theme", "default"),
     DARK("Dark theme", "dark"),
     BREEZE_DARK("Breeze Dark theme", "breezeDark"),
+    UNITY("Unity theme", "unity"),
     HIDPI("HiDPI theme", "hidpi");
 
     private final String name;
--- a/phoenicis-javafx/src/main/resources/com/playonlinux/javafx/themes/unity/descriptionTemplate.html
+++ b/phoenicis-javafx/src/main/resources/com/playonlinux/javafx/themes/unity/descriptionTemplate.html
@@ -1,0 +1,126 @@
+<!--
+  ~ Copyright (C) 2015-2017 PÂRIS Quentin
+  ~
+  ~ This program is free software; you can redistribute it and/or modify
+  ~ it under the terms of the GNU General Public License as published by
+  ~ the Free Software Foundation; either version 2 of the License, or
+  ~ (at your option) any later version.
+  ~
+  ~ This program is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  ~ GNU General Public License for more details.
+  ~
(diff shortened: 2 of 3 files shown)


qparis:
Waho!

plata:
@qparis I would have liked the miniatures to have a rounded shape. Doesn't seem to be possible in the CSS directly.

Ads20000:
@plata wow thanks!! The screenshots look great! I'll `git pull` and give this a try later :)

Ads20000:
OK I've tested it and I think it's good. Not quite as good as a 'native' GTK theme (I don't think?) but good enough given that it's not a GTK app.
Also @qparis could there be some way of getting POL to automatically use this theme if POL detects that it's running on Ubuntu? Someone on Ubuntu would expect the Ubuntu theme to be used, the default one looks very out-of-place.

plata:
@Ads20000 if you have any particular suggestion what could be improved in the theme, let me know.

Ads20000:
(First screenshot is current, second is what it looks like elsewhere on Ubuntu)

All merely my suggestions of course, feel free to ignore, or they may be impossible to implement.

I think it's really good, I just think that it may look slightly and unavoidably out-of-place in Unity 7 because it's not a GTK app. Also it **might need a bit of revision when Unity 8** becomes default (no idea when that will be, it's taking a very long time to mature), but since it isn't at the moment I guess we **should stick with trying to match it with Unity 7-style theming**.

When hovering over **items on the left sidebar**, they go a slightly darker grey. Perhaps they should go orange like in LibreOffice (it's slightly lighter than that actually but I can't screenshot the unselected version - and yes, it should go darker orange when selected rather than just hovered over - i.e. when you've got the Games which can be installed displayed). However I think it would be better if you went a similar approach to Files (Nautilus) where it's white unselected and doesn't change colour when hovered over, but goes a really nice orange when selected (so when you've got what you've clicked on displayed).
screenshot from 2017-01-28 18-13-22
screenshot from 2017-01-28 18-13-04
screenshot from 2017-01-28 18-13-52

I'm not sure this is the right way to do the **tabs** the top, maybe it would be better to do it like the Global Menu ones. I can't screenshot it properly, when using the Global Menu (the menu bar at the top), when clicking on an item like 'File' it looks a certain way (with the Ubuntu Ambiance theme). Make it look like that.
screenshot from 2017-01-28 18-15-00
screenshot from 2017-01-28 18-26-42



Is this what you mean by a **miniature**? Because yes, **it should be rounded**:
screenshot from 2017-01-28 18-11-55
screenshot from 2017-01-28 18-15-25

**Checkboxes** probably need some improvement too:
screenshot from 2017-01-28 18-31-07
screenshot from 2017-01-28 18-30-56

qparis:
> Also @qparis could there be some way of getting POL to automatically use this theme if POL detects that it's running on the Unity desktop environment (or just Ubuntu with Unity)? Someone on Ubuntu would expect the Ubuntu theme to be used, the default one looks very out-of-place.

As steam does ;-) Anyway, it will be up to Ubuntu packagers to set up the default configuration file properly

PR #535, opened 2017-01-28
Merging

qparis:


plata:
What about the build failure?

qparis:
Nothing has changed, it will continue to build properly

Issue #520, opened 2017-01-22
[Suggestion] Show which category is selected in "Apps"


Brainzyy:
I think this would be useful for the user.

qparis:
There is no concept of Selected category because the content can be filtered without category

Brainzyy:
@qparis Ah, I see, alright.

PR #538, opened 2017-01-28
Unity refinements

plata:
@Ads20000 like this?

checkbox

button and combobox

I tried to change the tabs but it doesn't look good.

### DIFF
--- a/phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/mainwindow/ui/LeftBarTitle.java
+++ b/phoenicis-javafx/src/main/java/com/playonlinux/javafx/views/mainwindow/ui/LeftBarTitle.java
@@ -18,9 +18,9 @@
 
 package com.playonlinux.javafx.views.mainwindow.ui;
 
-import javafx.scene.text.Text;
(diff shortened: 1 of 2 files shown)


qparis:
There is a weird white bar above left pane title. Is it expected?

plata:
@qparis that's the search field.

Ads20000:
Wow, that's amazing! Yeah the tabs would be difficult to get right, it's pretty good as it is!

Ads20000:
@plata One more thing, would it be possible to get this:
screenshot from 2017-01-29 01-28-57
to also work like this:
screenshot from 2017-01-29 01-29-11
Getting it to go orange when selected?

Ads20000:
@plata and another thing... The dialogue screen for adding a repository needs Ubuntufying
screenshot from 2017-01-29 01-32-50

plata:
Regarding toggle buttons for categories: this has been discussed in #520. 

I know about the dialog issue and I have tried to theme that for Breeze Dark already. However, I could not get it working. If I can figure it out, the theme can be enhanced of course.

Issue #540, opened 2017-01-29
Easy-to-use manual installation screen needed (missing POL 4 feature)


Ads20000:
Like in POL 4, would be nice if there was an option somewhere to manually install an app with the Wine version you want to use and with your own Winetricks options. Should be possible to change the Wine version of (and Winetricks used in) a manually installed app as well (you could keep it uneditable for an app installed with a script so that a user doesn't try to bump the Wine version number thinking it will help them).

plata:
Actually, we could provide simple wrappers for OnlineInstallerScript, LocalInstallerScript, SteamScript and QuickScript in a "Custom" category. Every time user input is needed, they would just show e.g. a file dialog.

Ads20000:
The text box to type in the name of the program is invisible (on the Ubuntu theme at least), I can see it when I hover over it but at first, because I couldn't see it, I thought I had to type the name in the search box top-left!

![screenshot from 2017-02-01 14-05-11](https://cloud.githubusercontent.com/assets/3178485/22509836/9149a784-e887-11e6-89da-e45092b17f4a.png)

Ads20000:
Oh and instead of producing an error, if I put no name in it allows me to continue anyway which is probably a problem?
![screenshot from 2017-02-01 14-07-22](https://cloud.githubusercontent.com/assets/3178485/22509897/c8dbf472-e887-11e6-9d0f-3c52de1d9046.png)

Ads20000:
Then, when I click Cancel, it says 'the script ended unexpectedly' in a popup, which is annoying because I told it to exit. Further, when I click show details it doesn't expand the window nicely to fit in the details box and when I maximise and then hide the details box it doesn't automatically reminimize like, say, the Software Updater on Ubuntu would. If I expand the details box again it looks really nasty. I don't think a box should appear in this case anyway, but presumably the same problems apply when a script genuinely crashes.

![screenshot from 2017-02-01 14-08-27](https://cloud.githubusercontent.com/assets/3178485/22509968/22f87c96-e888-11e6-8565-b8f9fe43e20a.png)
![screenshot from 2017-02-01 14-08-34](https://cloud.githubusercontent.com/assets/3178485/22509969/2319fdc6-e888-11e6-8d1d-a7f4c030cb45.png)
[...]

PR #550, opened 2017-02-01
disable next button if no text entered in wizard

plata:
see #540


qparis:
What if I want to ask something that is not mandatory?

plata:
Might happen. However, I do not really see a use case right now. If it is required at some point, we could just add another `textbox()` function specifying that it's optional.

Like
```javascript
var userText = setupWizard.textbox(message, default, optional);
```

qparis:
I like the first solution

qparis:
Why not allowing to pass a regex?

plata:
We could also pass a function which return true/false when it's given a string (thus determining whether or not the "Next" button is enabled).
Nevertheless, none of those really belongs to this PR so I would like if you could merge this such that the original error is fixed. We can then have a new issue to implement the optional textbox (if you believe that it is needed).

qparis:
Actually I'm pretty sure that we will need it, but ok.

Issue #551, opened 2017-02-04
Create distributable package


plata:
One package containing
- javafx
- cli
- libs

Think about the role of phoenicis-dist.

Consider using [jdeb](https://github.com/tcurdt/jdeb) to build .deb's in maven directly. 

Before people start distributing PoL, rename it to "Phoenicis" (changing the name afterwards will be too confusing).

plata:
With #552 a first approach has been implemented, still with several issues:
- uses phoenicis-dist, probably better in top level pom?
- installs to /usr/shar, should be split to /usr/local/lib and /usr/local/bin
- PlayOnLinux.sh is not executable after installation
- control file needs to be improved (e.g. dependencies)

PR #552, opened 2017-02-04
first approach to create .deb

plata:
see #551 

Just a first approach, still with several issues:
- uses phoenicis-dist, probably better in top level pom?
- installs to /usr/shar, should be split to /usr/local/lib and /usr/local/bin
- PlayOnLinux.sh is not executable after installation
- control file needs to be improved

### DIFF
--- a/phoenicis-dist/pom.xml
+++ b/phoenicis-dist/pom.xml
@@ -26,6 +26,14 @@
 	</parent>
 	<artifactId>phoenicis-dist</artifactId>
 
+    <dependencies>
+        <dependency>
+            <groupId>com.playonlinux</groupId>
+            <artifactId>phoenicis-javafx</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+    </dependencies>
+
 	<build>
 		<plugins>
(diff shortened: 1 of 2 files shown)


qparis:
Also dependencies should be included (cabextract, ...)

plata:
Do we have a list somewhere?

qparis:
So far, no. 
We should keep basically the same dependencies than POLv4, except for python stuffs

plata:
Ok.

Regarding phoenicis-dist: I'm trying to create the deb in the top level. However, I need to place the control file somewhere. Do you think it makes sense to put it to top level or should we rather keep phoenicis-dist?

qparis:
No I'm not really sure

qparis:
In POLv4, packaging files where put in a separate repository

plata:
Where can I find the control file for PoL4?

qparis:
https://github.com/PlayOnLinux/PlayOnLinux_Tools/tree/master/PlayOnLinux_Packager

plata:
Thanks.

PR #556, opened 2017-02-04
renamed PlayOnLinux to Phoenicis

plata:


qparis:
Hum I don't really like the idea to fully name "PlayOnLinux" "Phoenicis" for the moment. We need to give time to the community to adapt

qparis:
This won't be a problem. We cannot really do it now as we would need to create a new website, etc...

plata:
I think it will. As a matter of fact, the impact of the renaming will directly relate to the number of people using PoL. Package maintainers will have to rename their packages, people using PoL will suddenly have to switch to a "different" application etc.

Therefore my opinion is that the best approach would be to do the renaming now. Now only very few people use PoL5 and those who do are technically involved so it's no problem for them.
I would then announce Phoenicis as the successor of PoL4. This would give you a super clean transition as people would have to migrate anyways (from PoL4 to PoL5).

Also, I disagree that this would require a website right away. As long as PoL4 is the official latest release, no one will care if a Phoenicis website exists or not. You are right, however, that a new website will be needed in the future. But: this is the case anyways because playonlinux.com does not match anything of Phoenicis/PoL5 (most stuff on the website relates to scripts and is therefore outdated).
As we now have the scripts on GitHub, there is also no need to discuss them in a forum anymore. People can just use GitHub directly to propose scripts or file issues.
So, in my opinion, the only website which is really needed describes Phoenicis, provides download links and maybe a news blog. All this stuff could be handled with GitHub Pages.

qparis:
Unfortunately, github will never be enough. We have a big non-developer community (about 10k visits a day) talking about their wine / gaming experience we don't want to loose. We need some forums or equivalent for these people and we need these people to upgrade to POLv5 / Phoenicis. 

I agree that disitribution packages should be named phoenicis thought. But I'm not convinced yet that we should change the brand so quickly.

plata:
Ok. Understand the point regarding forums. On the other hand, maintaining them is quite some work. What do you think about something like [Discourse](https://github.com/discourse/discourse)?

Regarding changing the brand: Actually this is one of the reasons why I would prefer to change the name fast. Up to now, there is no PoL5 brand at all. So if we would rename now, there is no brand which could suffer from it. On the contrary people will directly get to know the Phoenicis brand instead of first getting used to PoL5 and then a real brand change to Phoenicis.

Just to get one thing right: I don't by any means want to push it. I'm just trying to explain my opinion.

qparis:
I think we should not talk about PlayOnLinux 5 but just "PlayOnLinux Phoenicis" just as "Mozilla Firefox" did before.

Regarding the forum, I think we need to discuss this later. But we will clearly need an equivalent of https://www.playonlinux.com/en/supported_apps.html feature.

plata:
Mh. If we use "PlayOnLinux Phoenicis", we still have the problem that it does not include PlayOnMac/PlayOnAndroid which is in the end weakening the brand.

qparis:
I think we will have to develop a simplified github user interface so that users can comment about scripts, modify them, sign them, etc...

qparis:
PlayOnMac Phoenicis 
PlayOnLinux Phoenicis

For new platforms, it shall be directly Phoenicis Mobile I guess (Like Firefox Mobile)

plata:
If you do it like this, people will be confused if Phoenicis is the version or the program name.

I don't believe that it's possible to create something simpler than what GitHub offers without loosing the necessary features.

qparis:
End users do not need to have all the features, ... and they need some others (Like a button to ask administrators to push and sign a script, icon displaying, etc...). Just a question of user interface.

Phoenicis is not a version, it is part of the name of the program

plata:
Of course we know that. But imaging someone just reading:
PlayOnMac Phoenicis
Phoenicis Mobile
I bet that he will think the following: There is a program called PlayOnMac which got updated from version 4 to a version called Phoenicis. And then, there is another program called Phoenicis Mobile which is probably completely unrelated.

Compare this to:
Phoenicis (for) Mac
Phoenicis (for) Mobile
Here it's immediately clear that there is ONE Phoenicis program which is available for Mac and for Mobile. Hence, the brand is much stronger.

qparis:
What about Phoenicis PlayOnLinux? 
The message I want to convey is that the future of PlayOnLinux 4 is Phoenicis

qparis:
Just to mention, Mozilla did it very well before with Firefox. Now, everybody calls FF Firefox

plata:
I agree with that message. And yes, I believe that Phoenicis PlayOnLinux would be good. It combines the idea of unifying PoL and PoM and at the same time it is still very clear that Phoenicis is related to PoL.
Just as you said we can have the same transition like Mozilla in the future (PoL → Phoenicis PoL → Phoenicis).

I think we should go that way.

Nevertheless, the code can transition from playonlinux to phoenicis already (as long as the distributed software is called Phoenicis PlayOnLinux / Phoenicis PlayOnMac, no one will care how the source code is called).

qparis:
The source code should definitely be called pheonicis. However, we should let the app titles intact for the moment

plata:
Fine for me. I will change that. How about the application.user.root?

qparis:
We're going to use phoenicis

qparis:
Can I merge it?

plata:
I think so, yes. I can later on change the naming for the .deb as well if you like.

qparis:
I think it will be org.phoenicis and not com.phoenicis

plata:
I didn't change anything about org/com. Must have been like this before already.

qparis:
Yes because the package was com.playonlinux because the adress of the website is playonlinux.com.
The address of the future website will be http://www.phoenicis.org/

plata:
Ok. Do you want to change this in this PR?

qparis:
Yeah I think it is better to do everything in the same time :)

plata:
I will try to change it.

qparis:
Maven modules + package names

plata:
I'm quite amazed. My IDE somehow changed everything once I renamed one package.

PR #569, opened 2017-02-06
Fixed again, previous link was pointing to phoenicis not PlayOnLinux

Ads20000:
Apologies @qparis

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,7 @@ This is the Phoenicis and PlayOnMac 5 repository.
 ## Scripts
 * Please make your pull request on this repository: https://github.com/Phoenicis/scripts to add scripts
 
-## Build instructions
-[Wiki: Building](https://github.com/Phoenicis/POL-POM-5/wiki/Building)
+## [Build instructions](https://github.com/Phoenicis/POL-POM-5/wiki/Build)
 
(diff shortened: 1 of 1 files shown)


Ads20000:
Oh, no, ignore this, sorry, looks like @plata is shaking things up

Ads20000:
I'm really sorry, I wasn't keeping an eye on what was going on, though @plata 's link is currently a dead one because the Phoenicis GitHub doesn't exist yet. Also what's the reasoning behind the name change?

plata:
The reasoning behind the name change is the following:
Currently, we have PlayOnLinux and PlayOnMac. In the future, we might also get PlayOnAndroid/PlayOnMobile. However, all these programs are in fact the same software. So we want one unified brand including everything.

Regarding the Phoenicis GitHub: I changed that by mistake (batch rename). I'm not sure if @qparis is planning to move POL-POM-5 to a Phoenicis repository (maybe in a Phoenicis organization). It would make totally sense in my opinion. If not, we should merge your fix. @Ads20000 please fix the link to the build instructions in wiki.

@qparis reopened for you to take a decision ;)

judge2020:
fixed with #578

PR #571, opened 2017-02-06
use StyleManager to change theme

### LINKED ISSUE #536: [Themes] Theme doesn't change
Theme doesn't change immediately after selecting a theme and pressing `Save`. I would expect it to change immediately.

Unity theme...but it's not
Dark theme...but it's not

### LINKED ISSUE #542: [Themes] Theme dialogs
Dialogs create a new stage, so we cannot just assign a style class.

See: <https://stackoverflow.com/questions/28417140/styling-default-javafx-dialogs>

plata:
- fixes #542 
dialog

- fixes #536

There is an issue with tabs:
tabs

@qparis do you have an idea what I need to do to fix this?

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/PhoenicisScene.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/PhoenicisScene.java
@@ -18,6 +18,8 @@
 
 package org.phoenicis.javafx.views.common;
 
+import com.sun.javafx.css.StyleManager;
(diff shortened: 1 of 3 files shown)


qparis:
Non idea. Isn't it a regression?

plata:
Yes it is. But I don't know why it behaves differently than before.

qparis:
It must be related with your CSS refactoring commit

plata:
No. It's related to the SceneManger somehow (did not change Breeze Dark in the refactoring commit).

PR #577, opened 2017-02-07
use WebView only for app description text

plata:
Simplifies theme handling as only one CSS is needed.

description


qparis:
No, we really want to keep rich text for script

plata:
Fine. But why the complete description including buttons etc? Should be enough to have only the text in a WebView. Then we could still do the theme in one CSS.

qparis:
First it could not because you'all always need to setup the background color and the text.
Second, you cannot make an HTML widget that has automatically the size of its content afaik

plata:
I guess you're right. So: What kind of content do you plan for the description that we need the rich text?

qparis:
The website front-end will be a richtext field (exactly the same thing we have in POLv4). 
See Photofiltre for an example

plata:
I will enhance the Scripts wiki page to describe that it's possible to use HTML.

Regarding WebView only for the description, I think we can do it:
WebView
Just a first approach and styling is missing, but I think it shows that the WebView can be directly integrated into the VBox.

qparis:
Hum I've been thinking about it but the scroll bar is in te wrong place

plata:
Depends. Suppose you have a very long description but you are not really interested in the description and rather want to install the application directly.

qparis:
I don't really see the advantage of doing this in fact

plata:
The advantage would be that the description could be styled with a simple CSS describing only the description itself. The titles, buttons and the layout could still be done in JavaFX.

qparis:
But why styling the button in pute JavaFX? We have much mor possibilities with html5

plata:
In my opinion, the button should integrate as much as possible with the look and feel of the whole application. The best way to achieve this is using a JavaFX button. Of course, we can do many fancy things with HTML5 but this will rather cause the impression of looking at a browser inside Phoenicis.

plata:
Current look:
photofiltre

qparis:
There is a large space between the title and the description

plata:
Fixed space:
fixed space

PR #580, opened 2017-02-09
Make icons part of the theme

### LINKED ISSUE #501: [Themes] Make icons part of the theme
We should use the system icon scheme to better integrate with the look and feel.

plata:
fixes #501 

I don't quite understand when the JavaFX objects (e.g. ViewLibrary) are created. Currently it seems to happen before the ThemeManager is initialized. Therefore, always the default icons are used. Could use some help here.


qparis:
ThemeManager should not be a singleton. The singleton pattern should be prohibited

plata:
Yes. I just did it as a quick solution until we have the bean. It's not even new in this PR.

qparis:
Can you fix it? It will be a good start to solve your problem

plata:
I would like to. But I'm not sure if I can (I don't really know spring, beans etc.). Maybe I'll give it a try.

I was thinking about something else: Wouldn't it be better to specify the icon path for every icon in the CSS? Even if I get the ThemeManager bean working, the icon would never be reloaded. So if the theme is changed, the icons would still stay the same. If the icons are specified in the CSS, this should work automatically.

qparis:
Any screenshot?

Anyway, I insist on the fact that the theme manager should be a bean and support multiple files. We cannot workaround this problem by trying to have a single file

plata:
themed icons

I agree with the bean. However, I would prefer not do do it with this PR. It will be big enough already (I must change some other stuff before this can be merged.

qparis:
What other stuff? This should really be a priority

plata:
E.g. fix the category icons (they are missing currently because they are not yet part of the theme).

qparis:
They should not be directly handled in CSS because we must be able to provide a default icon in the repository side. 

You really need to make ThemeManager a bean. Therefore, you'll rely on a strong architecture and be able to do much more stuff than you can do now

plata:
If the categories are can not have a theme, it will look like crap because two different icon styles are mixed up.

qparis:
Yep but on the other hand, the categories cannot be fixed

plata:
I don't see a problem. It's not like they change all the time.

plata:
themed app categories

qparis:
An non official repository should be able to add a category, this is very important

plata:
Ok. We could think about a fallback for situations like this.

qparis:
This is a question of consistency. The concept of category is on an ApplicationSource. Therefore, this component should be able to provide all the necessary data

plata:
We cannot provide the theme in the ApplicationSource. That doesn't make any sense.

qparis:
In this case we should go even further and make application miniatures part of the theme ;-)

The icon should be in the AppSource and optionally overwritten by the theme IMHO (exactly as it is done for gnome / kde categories)

plata:
The miniatures are not really icons but pictures so that's ok.

I tried to figure out in the code whether or not the icon is available in the theme. But that doesn't seem to work somehow.

plata:
Not exactly what I want but it has the same effect...

In this picture I removed the category "Custom" from the list of themed categories:
fallback

qparis:
Are you sure that the category icons are well aligned?

qparis:
I really think that we need to handle this differently :-/ I find the current approach not really clean

plata:
Well, if you have a better idea, fell free to propose.

qparis:
I have identified two responsibilities 

First, the theme manager should be able to provide any resource on demand.
Second, it should be able to notify some components when some resource are changed.

So we need 
1- a fetchResource(String name) method
2- a refresh method
2- a set of event handlers called when method is called

This component can be used basically anywhere, and should be initialized before the views. It needs to be a Spring bean so that it is instanciated in the right time

What do you think?

plata:
I'm not sure about the second responsibility: This would require that either the ThemeManager knows everything which could be affected by a theme change or components have to subscribe somehow.

Also, I couldn't find anything similar on the web. So I'm not sure if that's the way JavaFX is supposed to be used (I found quite some examples defining icons in the CSS).

plata:
Maybe the easiest thing would be: Store the icon path (not a `byte[]`) in the CategoryDTO. Then we can just set `-fx-background-image` to that path and we're done.

qparis:
don't use the path, it will no longer work for webservice or class path implementation. However, you can try to use base64 in the CSS file

plata:
How do I use base64?

qparis:
In HTML, it looks like this:
http://stackoverflow.com/questions/17090571/is-there-a-way-to-set-background-image-as-a-base64-encoded-image-in-javascript

plata:
It says: `WARNING: unknown protocol: data`

plata:
Can't the ApplicationSource just give a proper URL?

qparis:
Well I'm not 100% that javafx css supports base64 background images.
If it is not the case, we should not use CSS at it is a solution that does not fit our need.
Better to redesign a solution than twisting something that does not work for us

qparis:
The problem of handling the image in another URL is that we should also handle all the asynchronous stuff required to access it

plata:
Shouldn't the default theme handling do this already? I mean after all the images you specify in the CSS must be loaded as well somehow.

qparis:
In the CSS yep, but javafx is doing it for you. If you go from a byte[] to a URL, you need to implement it

qparis:
Oh sorry my bad!

qparis:
Can you try with a HTTP URL?

plata:
I tried. Doesn't seem to be any problem at all.
The also do it here: <https://stackoverflow.com/questions/10518458/javafx-create-custom-button-with-image>

qparis:
Ok. We just need to ensure that we can do it with classpath

plata:
Doesn't it use classpath by default?

qparis:
It uses git (local repository) by default

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -78,10 +84,28 @@ public void showAvailableApps() {
      */
     public void populate(List<CategoryDTO> categories) {
         Platform.runLater(() -> {
+            Set<String> themedCategories = Sets.newHashSet(
+                    "custom",
```
qparis: Why not generalizing it?

plata:
I was talking about the URL resolution. Or does it look in the jar only?

qparis:
I don't understand what you mean?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
plata: How?

plata:
Forget about it. I don't know regarding classpath.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
plata: Done.

Issue #581, opened 2017-02-10
PlayOnLinux 5 suggestions which wine should be used.


pyroxar:
When i install application non-script i see too much wine version (PlayOnLinux 4). 
New user does not know, nothing about wine.

In my opinion must be divide, something of a kind.
- stable (recommend)
- devel (not recommend)

I am only user and it is only suggestion.
Please, you must add this feature :)

plata:
There is already a division in
- upstream (what you call stable)
- staging (what you call devel)

qparis:
This is not really the same things

pyroxar:
Ok, staging or not. Now it is not clear. Wine in PlayOnLinux5 must be better described.

plata:
We could probably select the current stable version by default if that would help you.

pyroxar:
And should be written "Last stable" or "recommended for most users". More user experience and be ok.

plata:
For the "recommended" I see two options:
1. We just add it to the text in the ListView and remove it from the selected result before passing it to wine. Easy, but quite a hack. Might also cause problems with translations.
2. We implement a custom cell factory with an additional description text. Would probably require to populate the ListView with beans from javascript. Make the interface too heavy in my opinion.

PR #584, opened 2017-02-11
added option to specify default value for menu in wizard

plata:
see #581


phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/SetupWindow.java
```diff
@@ -67,8 +67,9 @@
      * Show a menu step and block until a response is given
      * @param textToShow The text to show to the user
      * @param menuItems The items in the menu
+     * @param defaultValue The item which is selected by default
      */
-    void showMenuStep(Message<String> doneCallback, String textToShow, List<String> menuItems);
+    void showMenuStep(Message<String> doneCallback, String textToShow, List<String> menuItems, String defaultValue);
```
qparis: The previous interface should also be available

plata: It is. Compare to the text box interface. It's done the same way.

PR #586, opened 2017-02-11
made ThemeManager a spring bean

### LINKED ISSUE #583: ThemeManager should be a spring bean instead of a singleton

plata:
fixes #583

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/ViewsConfiguration.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/ViewsConfiguration.java
@@ -19,6 +19,8 @@
 package org.phoenicis.javafx.views;
 
 import org.phoenicis.containers.dto.WinePrefixDTO;
+import org.phoenicis.javafx.views.common.ThemeManager;
+import org.phoenicis.javafx.views.common.Themes;
 import org.phoenicis.javafx.views.common.widget.PhoenicisLogo;
 import org.phoenicis.javafx.views.mainwindow.apps.ViewApps;
 import org.phoenicis.javafx.views.mainwindow.console.ConsoleTabFactory;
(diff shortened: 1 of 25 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/ViewsConfiguration.java
```diff
@@ -41,27 +43,46 @@
...
+
     @Autowired
     private ViewsConfigurationLibrary viewsConfigurationLibrary;
 
     @Bean
+    public ThemeManager themeManager() {
+        switch (theme) {
+            case "breezeDark":
+                return new ThemeManager(Themes.BREEZE_DARK);
+            case "dark":
```
qparis: This logic should be implemented in Themes enum (which name should be singular btw)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/ViewsConfigurationLibrary.java
```diff
@@ -28,9 +30,12 @@
     @Value("${application.name}")
     private String applicationName;
 
+    @Autowired
```
qparis: Configurations can be autowired but not beans

plata: How should it be done?

qparis: Just inject the configuration holding the bean, and access it then

plata: Like this?

qparis: Theoritically yes, except that you have created a cyclic dependency which is forbidenn. 
The right way to solve this problem in to create an extra module with a new configuration file

qparis: ViewsConfiguration and ViewsLibraryConfiguration will depend on this new configuration

plata: It's now parallel to the ViewsConfiguration. Is that ok (because the theme belongs to the views somehow)?

plata:
Sorry, I don't understand your remark. The ThemeManager constructor takes a Theme argument already and both classes are in `org.phoenicis.javafx.views.common`.

qparis:
My mistake

PR #587, opened 2017-02-11
fixed themed category icons

plata:
The previous solution did not work properly because setStyle always overrides the CSS.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
@@ -81,25 +81,13 @@ public class ViewApps extends MainWindowView {
      */
     public void populate(List<CategoryDTO> categories) {
         Platform.runLater(() -> {
-            Set<String> themedCategories = Sets.newHashSet(
-                    "custom",
-                    "development",
-                    "games",
-                    "graphics",
-                    "internet",
-                    "multimedia",
-                    "office",
-                    "other",
-                    "science"
-            );
-
             final List<LeftButton> leftButtonList = new ArrayList<>();
             for (CategoryDTO category : categories) {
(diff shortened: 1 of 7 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -79,13 +81,28 @@ public void showAvailableApps() {
      */
     public void populate(List<CategoryDTO> categories) {
         Platform.runLater(() -> {
+            Set<String> themedCategories = Sets.newHashSet(
+                    "custom",
```
qparis: I think that all categories can be themed

plata: Yes can. But are not necessarily (e.g. from some custom repository). I need to figure out for which ones the CSS can be used. I tried several things but could not find a better solution.

qparis: I think that any categories can not be themed by a given theme. This is important not to make particular cases in the code

plata: I guess we can expect that the "default" categories are contained in the theme (even if they link to the default icons).

I agree that it would be better to load the themed icon whenever it is available but, as I said, I don't know how this can be done.

qparis: I don't understand what is your blocker here:
What not asking ThemeManager if a resource with the name of the icon exists. If yes, then we can use it. 
It not, use the repository one

plata: You are right.

PR #588, opened 2017-02-11
use ThemeManager to load button icons

plata:


plata:
I did it to be consistent with the category icons. However, I'm not sure if this is really a good idea: If the icons are not specified in CSS, they also do not change when the theme changes.

plata:
Ok. So if we do it like this, how do we update the icons when the theme is changed? Or do we just say that you have to restart?

qparis:
If you want this, you will need an event handling mechanism.
You could create a ThemeResource class with two attributes: 
- resourceUrl
- Set<Runnable> onUrlChange

plata:
To be honest, I don't really need it. I was just wondering if people would expect it.

qparis:
I think indeed it is not a must to have

plata:
Fine. Should we add a label to the settings page stating "Please restart to apply the new icon theme?"

qparis:
Yes good idea

plata:
Ok. Will do that.

PR #592, opened 2017-02-12
wizard menu step: return index of selected element

### LINKED ISSUE #589: Wizard menu step: return index
The wizard menu step should return the index of the selection instead of the String.

If the menu is used to show an enumeration and the values are translated, it will not work as expected. When an index is returned, you could show any text and then map the index to the correct value. This would also allow us to show something like `2.0 "recommended"` as requested in #581.

plata:
fixes #589

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/MenuItem.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/MenuItem.java
@@ -16,14 +16,22 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-package org.phoenicis.scripts.wizard;
+package org.phoenicis.scripts.ui;
 
-import org.phoenicis.scripts.ui.MenuItem;
+public class MenuItem {
+    private String text;
+    private int index;
 
-import java.util.List;
+    public MenuItem(String text, int index) {
+        this.text = text;
+        this.index = index;
+    }
 
-public interface ChoiceWizard {
-    MenuItem menu(String textToShow, List<String> menuItems);
+    public String text() {
+        return text;
+    }
 
-    MenuItem menu(String textToShow, List<String> menuItems, String defaultValue);
+    public int index() {
+        return index;
+    }
 }
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/MenuItem.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/MenuItem.java
@@ -19,8 +19,8 @@
 package org.phoenicis.scripts.ui;
 
 public class MenuItem {
-    private String text;
-    private int index;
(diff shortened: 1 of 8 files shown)


qparis:
In POLv4, we needed to have both of these options. In fact, sometimes, you want to process directly the menu value (A good example is the menu that is asking you how much VRAM you have)

plata:
The index is more general. It is really easy to use it if you want the menu value directly (see PlayOnLinux/Scripts#165). Therefore I would prefer one interface instead of two.

qparis:
I find that it really makes the script more difficult to read. A better solution would be to return an object holding the index and the value. With this approach, we could later add any extra information we need

phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/MenuItem.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.scripts.ui;
+
+public class MenuItem {
+    private String text;
```
qparis: This should be immutable

phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/MenuItem.java
```diff
@@ -0,0 +1,37 @@
...
+    private int index;
+
+    public MenuItem(String text, int index) {
+        this.text = text;
+        this.index = index;
+    }
+
+    public String text() {
+        return text;
```
qparis: Getters should start by get (Otherwise, sonar will not be happy). 
However, I think that attributes can be public in that particular case

phoenicis-tests/src/main/java/org/phoenicis/tests/TestSetupWindow.java
```diff
@@ -59,8 +60,8 @@ public void showTextBoxStep(Message<String> doneCallback, String textToShow, Str
     }
 
     @Override
-    public void showMenuStep(Message<Integer> doneCallback, String textToShow, List<String> menuItems, String defaultValue) {
-        doneCallback.send(0);
+    public void showMenuStep(Message<MenuItem> doneCallback, String textToShow, List<String> menuItems, String defaultValue) {
+        doneCallback.send(new MenuItem("", 0));
```
qparis: Can we send null instead?

phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/MenuItem.java
plata: I called them `getText` first. But that looked out of place in js. Then I will go for the public members.

PR #593, opened 2017-02-12
enable a rich text presentation in the setup wizard

### LINKED ISSUE #428: StepRepresentationBrowse clickable URL
Allow to click a URL in the StepRepresentationBrowse.

plata:
Now, e.g. links can actually be links and not just plain text.

fixes #428 

rich text presentation

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/wizard/SetupWizard.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/wizard/SetupWizard.java
@@ -124,7 +124,7 @@ public class SetupWizard implements CompleteWizard {
      */
     @Override
     public Void presentationHtml(String htmlToShow) {
-        return messageSender.runAndWait(message -> setupWindow.showPresentationStep(message, htmlToShow));
+        return messageSender.runAndWait(message -> setupWindow.showHtmlPresentationStep(message, htmlToShow));
     }
 
     /**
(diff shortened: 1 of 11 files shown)


qparis:
Great Idea. I think it would ve preferable to create a generic html step

plata:
Just trying this. I think it will make a problem with the CLI.

qparis:
For CLI, we can strip tags

plata:
Is there any way to do that automatically or do I have to list them all?

phoenicis-scripts/src/main/java/org/phoenicis/scripts/wizard/SetupWizard.java
```diff
@@ -118,6 +118,16 @@ public Void message(String textToShow) {
     }
 
     /**
+     * Show a free script presentation
+     *
+     * @param htmlToShow the free presentation HTML to showRightView
+     */
+    @Override
+    public Void presentationHtml(String htmlToShow) {
+        return messageSender.runAndWait(message -> setupWindow.showPresentationStep(message, htmlToShow));
```
qparis: What is the difference between message and htmlToShow?

plata: message is the callback. It's the same as any other method in this class.

qparis:
My mistake

PR #597, opened 2017-02-15
Use SVG icons

### LINKED ISSUE #582: SVG icons
Replace PNG icons with SVG icons. We will be in trouble with high resolutions otherwise.

plata:
fixes #582 

It works, **but** the icons are blurry. Compare this:
SVG
svg
PNG
png

I have no idea why this is happening, but if we cannot figure it out we should probably stick to png.

Maybe it is this issue: <https://stackoverflow.com/questions/6837666/poor-image-quality-using-the-batik-svgconverter-from-svg-to-png>


qparis:
Does it work now?

plata:
I couldn't find the reason for this yet. As you can see, I opened an issue in the javafxsvg repository but also no update from there.
Could you try if you get the same effect?
I suggest to leave it open for now. Maybe we get some update. If not, we just stick to png.

plata:
As javafxsvg seems to be dead, I close this.

PR #600, opened 2017-02-17
added possibility to delete a wine prefix

### LINKED ISSUE #539: No option to delete containers
There should be an option to delete containers/apps, but there isn't.
(This might just be missing functionality, but it's functionality a user would expect to be there so I'm filing it anyway as a to-do task)

plata:
fixes #539

### DIFF
--- a/phoenicis-containers/pom.xml
+++ b/phoenicis-containers/pom.xml
@@ -64,5 +64,10 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.phoenicis</groupId>
+            <artifactId>phoenicis-library</artifactId>
+            <version>5.0-SNAPSHOT</version>
+        </dependency>
     </dependencies>
 </project>
--- a/phoenicis-containers/pom.xml
+++ b/phoenicis-containers/pom.xml
@@ -67,7 +67,7 @@
         <dependency>
             <groupId>org.phoenicis</groupId>
             <artifactId>phoenicis-library</artifactId>
-            <version>5.0-SNAPSHOT</version>
+            <version>${project.version}</version>
         </dependency>
     </dependencies>
 </project>
--- a/phoenicis-containers/src/main/java/org/phoenicis/containers/wine/WineContainerController.java
+++ b/phoenicis-containers/src/main/java/org/phoenicis/containers/wine/WineContainerController.java
@@ -22,6 +22,7 @@ import org.phoenicis.containers.dto.WinePrefixDTO;
 import org.phoenicis.containers.wine.parameters.RegistryParameter;
 import org.phoenicis.scripts.interpreter.InteractiveScriptSession;
 import org.phoenicis.scripts.interpreter.ScriptInterpreter;
+import org.phoenicis.tools.files.FileUtilities;
 import org.phoenicis.tools.system.OperatingSystemFetcher;
 import org.phoenicis.tools.system.terminal.TerminalOpener;
 import org.phoenicis.win32.registry.RegistryWriter;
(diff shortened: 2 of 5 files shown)


qparis:
When a prefix is deleted, it should also remove the corresponding shortcuts

plata:
How can I know them?

qparis:
By parsing all of them

plata:
I get the available shortcuts from the LibraryManager. But how do I know which prefix they use?

qparis:
This information is held in the shortcut, but the behavior you are looking for is only implemented in JS

plata:
I see. It's a bit tricky to figure out what is in Java and what in JS. Is there any particular logic behind this?

qparis:
What is in JS is what is customizable. I think we should find a way to script the container panels

plata:
Why?

In my case I want to get the prefix for a given container. For me, this would make more sense in ShortcutManager than in JS.

qparis:
Wine is something that is pluginable. This make no sense to have some files in the core if we want to be flexible enough to add as many engine as we want :-)

The ShortcutManager is just here to manage the existence of the shortcut. All the wine behavior are scripted because the core is engine agnostic as possible

plata:
Okay. So I shall add a function `winePrefix()` in JS?

qparis:
What it is going to do?

plata:
Return the wine prefix the shortcut uses so I can delete it if it is in the deleted prefix.

qparis:
I think we already have it

plata:
Ah ok. I was looking the Reader...

qparis:
Keep in mind that a shortcut is not necessary related to wine

plata:
The Shortcut seems to be only for the creation. So I think I have to use the Reader and add the function there.

qparis:
The reader is generic, then you have wineShortcutReader.

This component is quite poorly designed

plata:
I don't get it. It gives me `ShortcutManager is not a function`.

plata:
Ah. I cannot reuse the session.

phoenicis-containers/pom.xml
```diff
@@ -64,5 +64,10 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.phoenicis</groupId>
+            <artifactId>phoenicis-library</artifactId>
+            <version>5.0-SNAPSHOT</version>
```
qparis: Can you use ${project.version} ?

plata:
Should be ok now. The ShortcutReader determines the container of the shortcut (which is the wine prefix in case of a wine shortcut). This is then used to delete the shortcut if the container has been deleted.

phoenicis-containers/pom.xml
qparis: We will have a circular dependency problem here, since later, we'll also want to add a configure button on the shortcut

phoenicis-containers/src/main/java/org/phoenicis/containers/wine/WineContainerController.java
```diff
@@ -139,6 +156,34 @@ public void runInPrefix(WinePrefixDTO winePrefix,
         );
     }
 
+    public void deletePrefix(WinePrefixDTO winePrefix,
+                            Consumer<Exception> errorCallback) {
+        try {
+            fileUtilities.remove(new File(winePrefix.getPath()));
+        } catch (IOException e) {
+            errorCallback.accept(e);
```
qparis: Can you log this exception?

phoenicis-containers/pom.xml
plata: I guess you do not want to duplicate the GUI for this? So it probably would be enough to jump to the Containers tab.

Issue #528, opened 2017-01-23
Nothing happens when clicking on the thumbnail of an engine


Ads20000:
Presumably this is meant to do something, like give me a progress bar whilst it installs or something.

Ubuntu 16.10

plata:
You should see something like:

![container](https://cloud.githubusercontent.com/assets/3973260/22216959/6e576208-e1a1-11e6-942c-f8545cb889e8.png)

Do you get any error output when you run PoL from the console?

plata:
I guess this is also fixed after a pull?

Ads20000:
Urgh sorry about that, I meant *engine. Maybe nothing is supposed to happen? But then why is the screen even there? I can't see anything in Terminal is there a way I can produce more verbose Terminal output?

Brainzyy:
@Ads20000 This is supposed to be like that. The screen is there to help users and script creators know what engines are available for usage.

Ads20000:
That seems very odd to me, because it doesn't look like there's an option to change the Wine version of a container and it doesn't look like it's possible to manually install programs with POL like you could in POL 4. And having that list in the consumer-facing app just for developers seems silly!?

qparis:
This component is not finished yet

plata:
@qparis can you describe what the functionality of the "Engines" tab shall be?

qparis:
Install and remove wineversions basically

plata:
I'm trying to install a wine version from Java but the JS Wine expects a wizard which does not exist.

qparis:
We can create a different adaptation of wizard specially adapted for the engine tab

plata:
Ok. I will open some issue for this.

plata:
Superseded by #614, #615, #617.

PR #601, opened 2017-02-17
added engine panel

plata:
engine panel

In preparation for a fix for #528


qparis:
Not sure we should do it this way: I thing we should be able to différenciate very what versions are installed as in POLv4

plata:
I wanted to mark that on the engine miniature directly (e.g. by showing a green check).
We could also add a filter
- [] only installed
- [] only not installed

qparis:
Great idea

Also the button install should not be shown if the zngine is installed

plata:
Sure. Just wanted to get the basics done first so we can easily add all the other stuff later on.

PR #604, opened 2017-02-18
gray scale for not installed wine versions

### LINKED ISSUE #602: Mark installed engines
E.g. by adding a green check
installed

plata:
fixes #602 

grayscale

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/ControllerConfiguration.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/ControllerConfiguration.java
@@ -43,6 +43,9 @@ public class ControllerConfiguration {
     @Value("${application.name}")
     private String applicationName;
 
+    @Value("${application.user.engines.wine}")
+    private String wineEnginesPath;
+
     @Autowired
     private ThemeConfiguration themeConfiguration;
 
@@ -93,7 +96,11 @@ public class ControllerConfiguration {
 
     @Bean
     public EnginesController enginesController() {
-        return new EnginesController(viewsConfiguration.viewEngines(), enginesConfiguration.wineVersionsFetcher());
+        return new EnginesController(
+                viewsConfiguration.viewEngines(),
+                enginesConfiguration.wineVersionsFetcher(),
+                wineEnginesPath
+        );
     }
 
     @Bean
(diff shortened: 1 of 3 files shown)


plata:
@qparis I'm not quite sure about the way I check if the version is installed. I assume it should be done in JS, right? But then, where do I put the Java method to access this? I would expect WineVersionsManager but this seems to be used for available versions only.

qparis:
We could extend this piece of code

plata:
The WineVersionsManager? If so, should we rename it to WineManager?

qparis:
WineEnginesManager

qparis:
But maybe this should also be scriptable

plata:
What do you have in mind?

qparis:
I don't understand why using WineManager

plata:
How would you do it?

qparis:
I mean, why do you want to remove "Version" from the name?

plata:
I was thinking that we need an interface in Java to access the Wine JS. As this is more than just querying versions, a more general name would be better.

qparis:
On the contrary, I think that everything related to wine should be written in JS as it is more customizable

plata:
Ok, but I need to access this somehow from Java. Or should this be done in the controller directly?

qparis:
Why?

plata:
I need the information whether or not a certain version is installed to show it in ViewEngines. So somehow I must be able to access this data.

qparis:
We are already accessing getAllWineVersions() from Java.

But I really think that in the long run, Java should be wine agnostic and all interfaces should be implemented in JS

plata:
Are you talking about `getAvailableVersions()`? That's not what I need.

I think I get your point. You want something similar to ShortcutReader, right?

qparis:
I'd like to go even further. We should define an complete interface describing what an engine is, what a container is, and what a shortcut is. Then, we could implement it in JS for every kind of engine

plata:
Ok. But that's now a completely separate issue (can you create a new issue for this?). 

For this implementation now: Should I leave it as it is or implement `isEngineInstalled()` in JS?

qparis:
You can implement is installed engine

plata:
Ok. Is it ok to do it later and merge this first? I'm working on the filter as well so it would be easier to merge the current implementation for both first and then fix it with `isEngineInstalled()` together later on.

PR #606, opened 2017-02-18
Added engine filter

### LINKED ISSUE #603: Add engine filter
- [ ] only installed
- [ ] only not installed

plata:
fixes #603

filtered


qparis:
Why not using two checkboxes instead?

plata:
Can also do that if you prefer it.

plata:
checkboxes

PR #619, opened 2017-02-20
implemented engine install wizard

### LINKED ISSUE #617: Implement engines wizard
A new wizard implementation for the engine installation. Required by #614.

plata:
fixes #617

engine wizard
The screenshot shows a Notepad++ installation as the wine installation is not yet implemented.


qparis:
This implementation does not take advantage of the abstraction that has been put in the design on POLv5.

You really should leave the factory as it was. Then, you can create a totally different implementation of SetupWizard, that could be just a small status bar in the engine tab. Then, you can inject this SetupWizard as a paremeter of your JS function and your status bar will be well updated :)

plata:
Sorry, but I don't think that's true. I would have to implement the complete SetupWizard interface which is way more than just a progress bar because otherwise I break the design contract of that function.

qparis:
It could easily be implemented with popups later. Meanwhile you can throw a UnsupportedOperationException

plata:
Yes, it could. But I don't believe it is a good solution. From a design perspective, the installation of an app and the installation of an engine both belong to the concept "installation". Therefore, it doesn't make sense to choose a different presentation for the user.
In general, I believe that the apps tab and the engines tab should be as similar as possible. In many points this is already the case:
- app categories → engine categories (e.g. Wine)
- app miniatures → engine miniatures
- app description → engine description

qparis:
The user certainly does not want to see and extra wizard if this is not required.

As a matter of fact, SetupWizard was designed like this with sub interfaces before and was event implemented in Phoenicis before spring was introduced. 

On iOS, when you install or update an app, it looks like this: http://media02.hongkiat.com/ios-7-tips-tricks-features/apps-automatically-updated.jpg

We really should aim such simplicity for the end user

qparis:
However, I agree with the fact that we should also try to have the same behavior between tabs. Why not injecting a default simplified setup wizard when the script does not create one in this case?

qparis:
capture d ecran 2017-02-20 a 22 25 21

Wine installation only requires ProgressWizard

qparis:
What do you think of this?
capture d ecran 2017-02-20 a 23 35 45

plata:
I like the idea of using only the functionality of the wizard that is really needed. So `_installVersion` of Wine would only get the ProgressWizard like you described. The full blown wizard should then have some method like `getProgressWizard()`.
About the clock symbol I'm not sure. It looks nice but the relation between clicking the install button and this symbol might not be obvious.

I have a general question regarding this whole topic: Currently, the whole engine related stuff is really dependent on the Java implementation (e.g. WineVersionDTO, the way the engines are shown as miniatures etc.). Following the overall Phoenicis concept, I would expect that Java is more agnostic of Wine and treats it only as a special kind of engine.
I'm thinking about the following: In the scripts we have an engines folder containing several types of engines (Wine, ScummVM...). The engines tab would just loop over all engines similar to how the apps tab gets all categories. Every one of those engines implements `fetchAvailableVersions`. The result of this is not stored in an engine specific DTO but in a general EngineDTO. The EngineDTO contains certain fields like category, name etc., which apply to all engines. Additionally, there is a engineData field with information specific to a certain engine type (in JSON). For the installation and deletion, every engine must implement `install` and `delete`. Those functions receive an EngineDTO as parameter.

qparis:
Agreed with your last paragraph.

Concerning the clock, I think that we should find a way to allow the installation directly from the list. We could show the button when the mouse is over for example. 

> The full blown wizard should then have some method like getProgressWizard().

You don't really need to since you already have inheritance principles

plata:
I opened a new issue for the last paragraph topic: #620.

> Concerning the clock, I think that we should find a way to allow the installation directly from the list. We could show the button when the mouse is over for example.

So you want to completely remove the panel which shows the engine details? Don't you think that might contain interesting information?

qparis:
In my mind, the panel has to be kept. If you click on install, it will go back and show the clock

plata:
Understood. I'm a bit afraid that the following might happen: The user clicks "install" and suddenly finds himself on the miniatures list again. 
- In the worst case, he believe that something went terribly wrong and the detail page just crashed. 
- In a slightly better case he assumes that the installation is already finished because he doesn't recognize the clock on one of the many miniatures (especially if the miniature is not visible because the scroll pane scrolled to the top and the miniature is hidden somewhere at the bottom).
- The perfect case would be: The user immediately sees the clock and understands that the installation is still in progress.

qparis:
The transition animation will do everything ;-)

plata:
:D

plata:
Just looking at the interfaces for wizard. I'm a bit confused: Why do we have a SetupWizard which contains a SetupWindow instead of the SetupWindow implementing the CompleteWizard interface directly? If I want to create a ProgressWindowJavaFXImplementation, I need a ui.ProgressWindow interface. So in the end I would be more or less duplicating all interfaces from wizard in ui.

qparis:
a Wizard and a UI are very different concepts. I'm aware that it is hard to represent, but you could use a Wizard when no UI (for example an automated silent installation) and a UI with no wizard (A setupwindow that wouldn't be bound to any script basically). 

Separating them is a way to separate responsibility. For our case, you could imagine implementing a SetupWizard with no SetupWindow at all but a SetupWizard that just modify an existing interface (i.e. the engine tab). 


Don't know if I am clear.

plata:
Partly. I understand that those are different concepts. However, I don't understand why the SetupWizard should not have a SetupWizard and modify an existing interface instead. My idea was that the EngineSetupWizard implements ProgressWizard (through a new ui.ProgressUi interface). So it would not modify an existing interface but own its UI (a progress bar) which is simply embedded into the engine tab (could also be embedded somewhere else if needed).

qparis:
Yes, you can do this, what is the problem? :) SetupWizard needs a SetupWindow, because SetupWizard is an implementation of CompleteSetupWizard which is the interface.

I think we shall rename SetupWizard to SetupWizardSetupWindowImplementation and CompleteSetupWizard to SetupWizard. It will be clearer

plata:
I would also like to rename the SetupWindow because actually it's not a window. If I add the progress bar, it's not even a separate tab. So how about something more general like SetupUi, ProgressUi?

qparis:
SetupWindow *is* a SetupWindow. If you don't want to create a window, you need to implement a SetupWizard directly

plata:
I thought the SetupWindow is used by SetupWindowJavaFXImplementation implementation and shown when you install an app?

qparis:
SetupWindowJavaFXImplementation is an implementation of SetupWindow which is an interface
capture d ecran 2017-02-25 a 12 22 44

The only mistake we should fix is that SetupWizardFac

qparis:
Scripts needs a Wizard, not a Window

plata:
Sure. But in the end, SetupWindowJavaFXImplementation opens a tab which is clearly not a window. My point is that, following this, the progress bar for the engines would be a ProgressWindowJavaFXImplementation which doesn't make any sense in my opinion.

qparis:
The SetupWindow refers to the POLv4 window which was in fact a Window. You could imagine that you would be able to open a new window and do exactly the same thing. If you have a better idea, I'm interested.

Concerning your point, you can just implement ProgressWizard and inject it to your javascript wine object and it will work. It can be an anonymous class

plata:
It's just the naming^^. If we would call it SetupUi, it doesn't imply a window/tab or anything else. It would be perfectly logical to show a SetupUi in a tab, a separate window or simply in a console.

I have to think about how I want to do the engine progress bar in the end.

qparis:
All right, sounds good

plata:
Fine. I'll do the renaming and push it so you can have a look before I continue with the other stuff.

qparis:
Ok so now, I think we should remove the WizardType enumeration as it not abstract enough

plata:
engine progress

The progress UI is created from JS to show how it looks like. Will be injected when the engine install is really implemented but that should be done in a separate PR (this one is quite big already).

Issue #620, opened 2017-02-22
Implement generic engine handling


plata:
I'm thinking about the following: In the scripts we have an engines folder containing several types of engines (Wine, ScummVM...). The engines tab would just loop over all engines similar to how the apps tab gets all categories. Every one of those engines implements `fetchAvailableVersions`. The result of this is not stored in an engine specific DTO but in a general EngineDTO. The EngineDTO contains certain fields like category, name etc., which apply to all engines. Additionally, there is a engineData field with information specific to a certain engine type (in JSON). For the installation and deletion, every engine must implement `install` and `delete`. Those functions receive an EngineDTO as parameter.

### Container tab
We have to move all of this to JS. Without thinking a lot about it, we could do something like:
* Have a generic function `getEngineTools` in the JS engine which returns a list of function names and tool button icons. If you click a button, we call the provided function from Java.
* "Display" and "Input" are moved to a generic "Settings" tab (so we have "Information", "Settings", "Tools").
* Settings must be provided by JS, e.g. `getSettings` function which returns a list of `EngineSetting`. `EngineSetting` has a name, type (string, number, choice) etc.

To avoid having to implement a lot of functions for a new engine, we can have all of this optional. If you don't want to provide any tool, you do not have to implement anything. To do this, we derive all engine types from a base class `Engine` which e.g. returns an empty list for `getEngineTools`.

madoar:
I think #702 should be included here too.

madoar:
@qparis what do you think about this?

qparis:
I think we should think about the way it is done, but I'd like to release a alpha version first

plata:
I agree with @qparis. This is something internal which should not be noticed by the users (unless you provide your own engines of course, but that shouldn't affect alpha).

madoar:
The problem currently is, that we have quite a few issues which target the container tab and the engines (#833 #834 #702 #625). If we do these issues, or at least the issues marked for the alpha release, first and then this issue (#620), we'll need to do a lot of things two times, the first time when we work on the previous issues and then again when we move the features to javascript as part of this issue.

plata:
You are right (didn't see that #833 and #834 are assigned to alpha).

madoar:
I think it would be a good idea to start the whole topic by moving wine engine specific functions from java (the POL 5 repository) to javascript (the Scripts repository). 
When doing this we can also add functions for #833 and #834 at the same time in javascript.
[...]

PR #622, opened 2017-02-26
allow apps in a separate "Apps" directory

plata:


qparis:
I don't understand the purpose of this commit

plata:
Currently, the apps categories and the functions folder are in the same directory. In my opinion this should be separated.

qparis:
Why?

plata:
Because the functions are no apps. They only do not show up in the Apps tab because they do not contain a category.json. I would prefer if you could look at the repository and can tell immediately where the installable applications are.

qparis:
Why not... But I think that your commit is going to break javascript include function

plata:
Don't know. I installed Notepad++ with the configuration and it worked.

plata:
Maybe it would even make sense to move the engines from functions to an "Engines" directory at top level. This would emphasize the similarity between apps and engines as it can also be found in the GUI (at least when #620 is implemented).

qparis:
With this code, fetchCategories() will no longer return functions, and I'm not 100% sure that it is ok

plata:
It will. It only searches the Apps directory additionally. Moreover, I don't understand why it should return functions. After all, Functions is not an app category so why should the ApplicationsSource fetch it?

qparis:
Because this is how include is designed

plata:
Ok. Can you tell me where this is implemented? (just want to understand it)

plata:
Found it. ScriptFetcher.getScript right? Wouldn't it be more logical to call the ApplicationsSource RepositorySource if it fetches applications, engines and functions?

qparis:
This is not strange. You need to imagine that you can call any script for any script. This is the idea of the structure.
What you want to implement is a 4-level namespace instead of a 3-level one. I think it really makes sense, but the whole problem should be thought about. For example, ClassPathApplicationSource also needs to be updated. The DTO set also has to be updated

plata:
Don't get me wrong: I don't think it's strange that there is a class which fetches the scripts in the repository. My concern was that this class is called `ApplicationsSource` which suggests that it fetches only applications.
To be honest, I didn't really realize the implications of this PR. But maybe it would really make sense to switch to 4-level as you say (before we have so many apps that it's too hard to handle this). I don't know about the `ClassPathApplicationsSource` but the DTO should be rather easy: Just a new top level DTO which distinguishes functions, apps etc.

plata:
Seems to work like this. The Functions folder contains another Functions folder. If you know a better naming, let me know ;)

qparis:
Shall we close this Pull request?

plata:
Superseded by  #745.

PR #623, opened 2017-02-27
prepared generic engines (including install/delete)

### LINKED ISSUE #614: Install wine engine

### LINKED ISSUE #615: Delete wine engine

plata:
Preparation for a generic engine handling as described in #620. 

Includes a generic install/delete.
- fixes #614
- fixes #615
- requires PlayOnLinux/Scripts#175

The EngineVersionDTO is not yet generic. It should contain an engine specific userData Map like EngineDTO. However, this requires a change in the Wine JSON format (Wine specific information like Mono/Gecko file should be contained in userData).

### DIFF
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/CombinedEnginesFilter.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/CombinedEnginesFilter.java
@@ -19,8 +19,9 @@
 package org.phoenicis.engines;
 
 
-import org.phoenicis.engines.dto.WineVersionDTO;
-import org.phoenicis.engines.dto.WineVersionDistributionDTO;
+import org.phoenicis.engines.dto.EngineCategoryDTO;
+import org.phoenicis.engines.dto.EngineSubCategoryDTO;
+import org.phoenicis.engines.dto.EngineVersionDTO;
 
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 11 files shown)


qparis:
Do you have the diagram class of the new DTO set?

plata:
No. But it didn't really change much. WineVersionDistributionDTO and WineVersionDTO got renamed to EngineSubCateoryDTO and EngineVersionDTO. EngineCategoryDTO has been added on top to distinguish the engine categories (Wine, ScummVM...).

qparis:
What IDE are you using?

plata:
IntelliJ IDEA (the community version does not have the UML tools).

qparis:
We have free pro license if you are interested

plata:
Unless I shall do more UML, I think the community is enough for me ;)

qparis:
The pro edition allows you to use Maven, and test coverage

plata:
I can use Maven build in the community edition as well. What can I do more in pro? BTW: Could you add the descriptions about code formatter etc in README also for IDEA?

qparis:
The code formatter settings should be directly committed 

Here are the main differences https://www.jetbrains.com/idea/features/editions_comparison_matrix.html

plata:
Oh. JS support would be nice for the scripts.

What do you mean the formatter settings should be committed? Is it possible to add a beautify step to the maven build?

qparis:
This is possible with maven but it would lead to a ping-pong effect if the ide is not well configured.

I suggest integrating the code formatting settings in the source repository (I'm pretty sure that it is already done for at least eclipse)

plata:
I never did code formatting for Java so I cannot tell. For C++ integrating the beautifier into the build system is quite nice as you do not have to care about anything.

qparis:
Well you have to because you are not obliged to use maven to POL. Also, you could have maven de synchronized with the IDE which would lead to pollution commits.

Before thinking about the tool we are going to use, we need to think about the way the convention are described (xml codestyle files ?)

plata:
Let's continue this in  #627.

PR #624, opened 2017-02-27
implemented apps search

### LINKED ISSUE #372: Implement search for application in "Apps"
This could be done with a FilteredApplicationsSource class which applies a filter to the result of a given ApplicationsSource. 
The filter can be an object containing

- testing flag
- no CD needed flag
- commercial flag
- search string

I think a really simple search with `String.contains()` is enough for this purpose.

plata:
fixes #372

apps search

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/ApplicationsSource.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/ApplicationsSource.java
@@ -52,6 +52,8 @@ public interface ApplicationsSource {
         return null;
     }
 
+    default void setFilter(CombinedAppsFilter filter) {}
+
     default void getScript(List<String> path, Consumer<ScriptDTO> callback, Consumer<Exception> errorCallback) {
         callback.accept(getScript(path));
     }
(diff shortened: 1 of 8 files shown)


qparis:
Just a suggestion: why not being more abstract and try to create a FilteredApplicationSource implementation?

plata:
My idea is to enhance the `CombinedAppsFilter` for all possible filter criteria (e.g. commercial, testing...). The `FilteredApplicationsSource` is totally agnostic about the filter implementation. If I would implement the filter directly in a `FilteredApplicationsSource` implementation, I would have to change this object when the filter changes.
If I'm missing something, please tell me.

qparis:
I think that ApplicationSource (which should be renamed as you pointed it out before) should also implement a mechanism so that the result can be refreshed at any time

plata:
I agree. We should try to make a plan how all of this should look like (probably including some UML for the classes and DTOs).

plata:
Nevertheless: Can we merge this and then think about the general structure of ApplicationsSource, the DTOs etc. as part of  #622?

PR #630, opened 2017-03-04
added "About" window

### LINKED ISSUE #629: Show Play on Linux 5 version
Is there a way to see the current Play on Linux 5 version in the javafx gui client?
I believe it would be useful to have a window that shows your current version and maybe also an information if a newer version is available.

plata:
about

fixes  #629


madoar:
Could you please add the build date too? I know that it currently won't be helpful when doing a bug report, but I think it will be helpful for the users to see if they are using an up to date POL 5.

plata:
timestamp

madoar:
Great! I guess the git revision is the git commit?
If yes, it would be nice if you could click on it and a browser is automatically opened that shows the commit, or maybe the commit list?
For the first option a link to https://github.com/PlayOnLinux/POL-POM-5/commit/revisionId would be needed, where revisionId is the the git revision from your screenshot.
For the second option a link to https://github.com/PlayOnLinux/POL-POM-5/commits/master would be needed.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/ViewSettings.java
```diff
@@ -52,6 +52,10 @@
     private static final String CAPTION_TITLE_CSS_CLASS = "captionTitle";
     private static final String CONFIGURATION_PANE_CSS_CLASS = "containerConfigurationPane";
     private static final String TITLE_CSS_CLASS = "title";
+    private String applicationName;
```
qparis: These could be final

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/ViewSettings.java
```diff
@@ -210,6 +227,34 @@ private void initNetworkPane() {
...
+        final GridPane gridPane = new GridPane();
+        gridPane.getStyleClass().add("grid");
+
+        gridPane.add(new TextWithStyle(translate("Name:"), CAPTION_TITLE_CSS_CLASS), 0, 0);
+        gridPane.add(new Label(applicationName), 1, 0);
+
+        gridPane.add(new TextWithStyle(translate("Version:"), CAPTION_TITLE_CSS_CLASS), 0, 1);
+        gridPane.add(new Label(applicationVersion), 1, 1);
+
+        gridPane.add(new TextWithStyle(translate("Git Revision:"), CAPTION_TITLE_CSS_CLASS), 0, 2);
```
qparis: What if we are not running a dev version?

plata: Does it matter? Also release versions must have been built with a certain revision.

madoar: In this case I would suggest to show the git tag of the release. But I'm not sure how to automatically detect this best.
One option would be to analyse the version string. If it ends with RELEASE instead of SNAPSHOT, the client could automatically show the git tag belonging to the version, otherwise it shows the git revision, like it does now.

plata: I have no idea how to do this.

plata:
@madoar to open a browser, I would have to call:
```Java
getHostServices().showDocument("https://github.com/PlayOnLinux/POL-POM-5/commit/" + applicationGitRevision);
```
But this requires access to JavaFXApplication in ViewSettings. I can't currently think of a nice way to achieve that.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/ViewSettings.java
madoar: @plata I think your answer is right. The release versions will be built with a certain revision. Because of this I think it's alright to show the git revision. In addition the version string tells the user that he/she is using a release version.

madoar:
Hmm, how is access to JavaFXApplication given? Do I need to add something in my OS settings? Or is it possible to show a popup that asks for permission?
In addition, is it possible to previously detect if the permission is given? If yes you could prevent the link if no permission was given and only show it if you know you have permission to open the browser.

plata:
What do you mean by "how is access to JavaFXApplication given"? I don't quite understand what kind of permissions you are talking about (and how this relates to this PR).

madoar:
You said that it requires access to JavaFXApplication in ViewSettings. 
I don't know how I give access to JavaFXApplication in ViewSettings. Is this done programatically, or do I need to give access somewhere in my OS settings?

plata:
JavaFXApplication and ViewSettings are both Java classes. JavaFXApplication is derived from Application and ViewSettings is the class which shows the "About" information. For whatever reason, `getHostServices()` is not static in Application so I need a JavaFXApplication instance in ViewSettings.

plata:
@qparis unless you have some clever solution for this, I would suggest to merge it as it is now.

madoar:
I think that the best long term way to solve this would be Dependency Injection. But after a short look over the code, I don't know if and how you want to use Dependency Injection. 
I've found some `@Autowired` annotations in some config classes that indicate dependency injection, but I'm not sure if you're only want to use it for config purposes.

plata:
I'm not really sure that it is a good idea to inject the Application into the Views. It holds the ApplicationContext but for the dependency injection it should be a Bean itself, shouldn't it?

qparis:
We need to have a third component whose responsibility is to hold the data you need.

JavaFXApplication and ViewSettings will depend on this component. One will write data to it, the other will read it.

@madoar, we already have a dependency injection framework set up

plata:
It's not really about data. I just would like to access `getHostServices()` which has to be called on the Application object (of course we could move the data to a separate class but that's not really the problem right now).

qparis:
Can you call getHostServices on startup?

plata:
I don't think so. The way it works is like this: You call it with some URI, it fetches the system application and opens the URI.

qparis:
Then you can register a Supplier<URI> on startup to this intermediate class

plata:
Ok... Honestly, I have no idea how this should be set up. Could you describe it more detailed?

qparis:
Ok my idea is quite simple. You could create a class that has a mutable Runnable attribute. Runnable has no argument so you can use Consumer or Supplier instead.

When POL starts, you can asks JavaFXApplication to inject the behavior you will need to this intermediate class (with a setter)

When you need your behavior later (in ViewSettings), use the getter and apply

plata:
Where do I put the class (in the source tree)? Should I add a Bean to ViewsConfiguration?

qparis:
Is this class only related to configuration or does it has any other added value?

plata:
Currently the only use I can think of is providing the HostsServices.

qparis:
Sorry but I have not followed the whole thread. What is HostsService?

plata:
The return value of `getHostServices` (which is used to open the browser).

plata:
I have implemented something but it's not working. I think it is the same bug as described here: <https://bugs.openjdk.java.net/browse/JDK-8160464>

qparis:
Why not using xdg-open instead?

plata:
Changed it to xdg-open. I also left the code for the HostServices solution because I believe it would be the better way to do this with Java means (e.g. to support Android later on).

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/ViewSettings.java
```diff
@@ -250,7 +251,17 @@ private void initAboutPane() {
         gridPane.add(new TextWithStyle(translate("Git Revision:"), CAPTION_TITLE_CSS_CLASS), 0, 2);
         Hyperlink gitRevisionLink = new Hyperlink(applicationGitRevision);
         gitRevisionLink.setOnAction(event ->
-                hostServicesSupplier.get().showDocument("https://github.com/PlayOnLinux/POL-POM-5/commit/" + applicationGitRevision)
+                {
+                    String url = "https://github.com/PlayOnLinux/POL-POM-5/commit/" + applicationGitRevision;
+                    // use xdg-open until HostServices works
+                    Runtime runtime = Runtime.getRuntime();
+                    try {
+                        runtime.exec("xdg-open " + url);
```
qparis: The right command on OSX is "open". We should create a dedicated class for this

madoar: I agree with @plata on this, I'm not sure creating a dedicated class for this is a good idea, especially if you want to support a large number of platforms which may use different tools for opening a web browser.
To better support multiple platforms I would recommend using the build in features of java like the HostServices class.

qparis: We always needed xdg-open and open command on POLv4 for a lot of reasons (opening documents mainly).
What you are saying is exactly what I have in my mind. If you write this piece of code in a dedicated interface, you'll be able to clean it later without changing all the implementations everywhre. See #637

qparis: TLDR: Ugly code should always be encapsulated

madoar: I know what you mean to say, but I think that this is unneeded in the case of opening the browser, because java itself has a function for that (`HostServices.showDocument`). I would agree to the suggestion to use a dedicated interface for this if it's needed for other tasks that aren't supported directly by java.

qparis: HostServices is a JavaFX class. We need something that is not dependant on JavaFX to be used in the scripts

madoar: You mean, because OpenJDK 8 often doesn't include the JavaFX classes? Sry, I forgot that...
Is it possible/ok for us to use awt classes? If we can use awt classes, then maybe the following code could be a solution?
```java
import java.awt.Desktop;
import java.net.URI;

// ...

if(Desktop.isDesktopSupported())
{
  Desktop.getDesktop().browse(new URI("http://www.example.com"));
}
```
(see also [here](http://stackoverflow.com/questions/5226212/how-to-open-the-default-webbrowser-using-java))

qparis: We need to imagine that POL could be run without any GUI (automated scripts). 
I don't like the idea of loading awt just to open the browser

plata:
Like this?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/LinuxWebBrowser.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.IOException;
+import java.util.function.Supplier;
+
+public class LinuxWebBrowser implements WebBrowser {
+
+    @Override
+    public void openUrl(String url) {
+        Runtime runtime = Runtime.getRuntime();
+        try {
+            runtime.exec("xdg-open " + url);
```
qparis: This implementation is not safe. See my last PR

qparis: (You could potentially inject code here)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/ViewSettings.java
plata: Apart from this, the awt Desktop is not even working all the time (e.g. on my PC it didn't work).

madoar: My problem with the dedicated interfaces approach in this case is, that it's possible, that the approach fails, if no working implementation for the used platform is found, which happens, if xdg-open is not installed on a linux platform. 
But I can't find a better approach, when we want to have a java gui-library-free solution.

qparis: The idea of using an interface is to be allowed to improve the implementation with corner cases later on

plata:
@qparis @madoar I think this should be fine now.

PR #637, opened 2017-03-05
Opener command

qparis:


qparis:
@plata might be useful for you PR

phoenicis-tools/src/main/java/org/phoenicis/tools/system/opener/OpenerProcessImplementation.java
```diff
@@ -0,0 +1,46 @@
...
+        final ProcessBuilder processBuilder = new ProcessBuilder(Arrays.asList(commandName, file));
+        try {
+            processBuilder.start().waitFor();
+        } catch (InterruptedException | IOException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    @Override
+    public void open(URI url) {
```
plata: is this needed? shouldn't it use the default in the interface anyways?

phoenicis-tools/src/main/java/org/phoenicis/tools/system/opener/Opener.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * System Opener
+ */
+@FunctionalInterface
+public interface Opener {
+    /**
+     * Open a file
+     * @param file to be opened
+     */
+    void open(String file);
```
plata: Don't like "file" (it's also used e.g. for URLs). Doesn't it allow to inject code as well?

phoenicis-tools/src/main/java/org/phoenicis/tools/system/opener/OpenerProcessImplementation.java
qparis: The command xdg-open can be used to open a file (a document) or a URL. This is why I want to separate the two behaviors

phoenicis-tools/src/main/java/org/phoenicis/tools/system/opener/Opener.java
qparis: See my other comment :)

phoenicis-tools/src/main/java/org/phoenicis/tools/system/opener/OpenerProcessImplementation.java
plata: Yes. I just was thinking that this method here duplicates the default method in the interface (it's the exact same code).

qparis: Oh yeah sorry, it needs to be removed on this implementation

phoenicis-tools/src/main/java/org/phoenicis/tools/system/opener/Opener.java
plata: Understand the idea. But no one can prevent me from calling this with a String URL directly (which basically leads to the same code injection issues like my implementation).

qparis: Not really, here it is impossible to inject pipes for example

plata: Ok. Anything you want to change? Otherwise I'll merge and update the "About" PR.

plata:
Sorry for not squashing. Didn't realize that it is not the default option. Could you change the settings so this doesn't happen again?

qparis:
Not sure we can do it

plata:
Why?

PR #639, opened 2017-03-06
fixed build if path contains spaces

plata:


phoenicis-tools/src/test/java/org/phoenicis/tools/archive/ExtractorTest.java
```diff
@@ -83,7 +89,12 @@ public void testUncompress_withSymbolicLinks() throws IOException {
     }
```
madoar: I'm not sure the try/catch is needed here, because we could just add URISyntaxException to the throws in the method declaration like this:

```java
     @Test
     public void testUncompress_withSymbolicLinks() throws IOException, URISyntaxException {
         File  inputFile = new File(inputUrl.toURI().getPath(), "tarLink.tar.gz");

         final File temporaryDirectory = Files.createTempDir();
 
         temporaryDirectory.deleteOnExit();
 
         final List<File> extractedFiles = extractor.uncompress(inputFile, temporaryDirectory, e -> {});
 
         final File file1 = new File(temporaryDirectory, "file1.txt");
         final File file2 = new File(temporaryDirectory, "file1_link.txt");
 
         assertTrue(file1.exists());
         assertTrue(file2.exists());
 
         assertEquals("file1content", new String(FileUtils.readFileToByteArray(file1)));
         assertEquals("file1content", new String(FileUtils.readFileToByteArray(file2)));
 
         assertTrue(java.nio.file.Files.isSymbolicLink(Paths.get(file2.getPath())));
     }
```

qparis: A public method should not throw more than one exception

madoar:
What I'm asking myself is, if this problem with the spaces occurs only in the tests or if it occurs in the main program too?

phoenicis-tools/src/test/java/org/phoenicis/tools/archive/ExtractorTest.java
madoar: Oh, ok, I didn't know this guideline.

plata:
I didn't test completely but at a first glance the program seems to work (also could not find any other `getFile()`).

phoenicis-tools/src/test/java/org/phoenicis/tools/archive/ExtractorTest.java
plata: Mh it's only in the test. I don't think it makes sense to throw in some tests and try-catch in others.

qparis: Oh yeah my mistake 
Sorry

phoenicis-win32/src/test/java/org/phoenicis/win32/registry/RegistryParserTest.java
```diff
@@ -60,8 +61,8 @@ public void testRegistryParser_parseSimpleFile_testKeyTree() throws IOException
     }
 
     @Test
-    public void testParse_realRegFile_testObjectPopulated() {
-        File registryFile = new File(this.getClass().getResource("user.reg").getFile());
+    public void testParse_realRegFile_testObjectPopulated() throws URISyntaxException {
```
qparis: This should be avoided if possible

madoar: What exactly do you mean? The method name in snake case?

qparis: Sorry I was thinking that the production (non test) code was actually throwing a checked exception, which is what we are trying to avoid

Issue #634, opened 2017-03-04
Bad color choice for header texts in the Dark theme


madoar:
When using the "Dark theme" the header texts in the left sidebar like "Settings" above "User Interface" in the following screenshot are written in black, which is really hard to read with the dark background color:
![grafik](https://cloud.githubusercontent.com/assets/18488086/23580589/df97e4ce-0104-11e7-8764-9faa5508b417.png)

plata:
If you want to change this, there is a tutorial about themes in the [wiki](https://github.com/PlayOnLinux/POL-POM-5/wiki/Themes). For this particular case, you need to change `leftBarTitle` in phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/dark/main.css.

PR #640, opened 2017-03-07
Fixing #634

madoar:
This should fix issue #634.
What I don't understand is, why `-fx-fill: #FFFFFF;` doesn't change the text color to white for the dark theme, because at least in my eyes it works for the Breeze Dark theme (`-fx-fill: #bdc3c7;`)

### DIFF
--- a/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/dark/main.css
+++ b/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/dark/main.css
@@ -148,7 +148,7 @@
     -fx-font-size: 1.1em;
     -fx-alignment: baseline-left;
     -fx-translate-y: -0.1em;
-    -fx-fill: #FFFFFF;
+    -fx-text-fill: #FFFFFF;
 }
 
 .leftButton {


plata:
Please always add a screenshot when changing something in the GUI. Also use a more descriptive title.

The change itself is good. Probably it does not work for Breeze Dark as well. However, Breeze Dark also sets the default colors:
```CSS
.root {
    -fx-text-fill: #eff0f1;
}
```

plata:
One additional hint: You can write the the commit message such that the issue gets closed automatically when the PR is merged. See: [Closing issues via commit messages](https://help.github.com/articles/closing-issues-via-commit-messages/)

madoar:
This is my first Pull Request, so thanks for the information on how to improve my future Pull Requests.

I didn't close the issue because I first want to have this PR approved, otherwise I wouldn't say the issue is solved.

About the default colors:
Which approach do we want to take? Do we want to use a default text color or should we set it individually for each component type?

plata:
The issue will be closed only after the PR is merged. So if you change "fix issue #634" to "fixes #634", everything is fine.

I think it makes sense to overwrite the default colors because you do not want a black text to show up anywhere in a dark theme.

madoar:
I've just taken a look at the default colors approach.
It seems like I must define the default color for every type of component, because the addition of 
```css
.root {
	-fx-text-fill: #ffffff;
	-fx-text-base-color: #ffffff;
        -fx-text-inner-color: #ffffff;
}
```
in the beginning doesn't seem to be enough. If I only add such a root class and remove the previous added `-fx-text-fill` from `.leftBarTitle` the bar title will be black again. What seems to work is the addition of a `.label` class, but in my eyes this approach is quite similar to the addition of `-fx-text-fill` to every text component class.

plata:
Unless no one uses a "plain" label (without a CSS class assigned), you're probably right. For this PR it doesn't matter anyways, so I would go ahead and push this (unless you want to change something?).

madoar:
No, as long as the current solution works, I don't see the need to change something ;)
It just would be nice to refactor the themes at a later time to make them more uniform to each other, i.e. define default colors etc.

plata:
I think the dark theme is the only one which does not set the defaults (the default/HiDpi theme use the default colors).

PR #646, opened 2017-03-09
Added "clear" buttons for the search textfields

madoar:
Every searchfield contains now a "clear" button, that automatically clears the content of the textfield after it has been clicked:
grafik

This currently looks a little strange on the hidpi theme, because the textfield doesn't scale there.


madoar:
Ok, I've just fixed the problem with the width of the textfield being too small in the hidpi theme.
Now I have discovered a new problem. I've added two different button icons for the clear button, a light one and a dark one. The dark icon is used in the dark breeze theme, while the light icon is used with the other themes. 
The problem is now, that when you change your theme to the dark breeze theme, or away from it, you need to restart POL 5 to see the correct button icon.
I've discovered that the theme change can be detected somehow, because the icons for the existing containers (in the "Containers" tab) change, if you change your theme from, or to, the dark breeze theme.
Can anyone tell me how to detect the theme change, and use this to set the correct clear-button-icon belonging to the new theme?

plata:
The icons are reloaded if they are part of the theme (i.e. specified in the CSS directly). We had that some time ago for all icons. However, we changed it back to specifying the icons in the code because we needed a fallback mechanism for the app categories (as you can have your own app source, we cannot know which categories can occur).

The containers tab is somewhat different. It changes the icons because the GUI is recreated (could be that containers have been added/deleted in the meantime). Probably not the smartest solution of all times but it works for now.

madoar:
A possible solution to listen to theme changes could be, if we wrap the currentTheme in the ThemeManager class inside an ObjectProperty.
Then we can add an `addListener` method to the ThemeManager class that forwards the given listener to the new ObjectProperty.
This enables us to detect theme changes at runtime and react to them.
What do you think?

plata:
Should work. The problem is that either all views must react on a change or the ThemeManager must know all themed resources (we probably don't want that).
I'm wondering how this is done usually. I mean we are not the only ones with themed icons in a JavaFX application so there should be a standard way to do this. I would also like the icons to resize automatically according to the screen resolution.

madoar:
Why must either all views react on a change or the ThemeManager know all themed resources?
With my idea we would only add a way to bind a listener to the currentTheme, which can detect a change of the object. This doesn't mean that anything else needs to be rewritten.

I'm not sure how other JavaFX applications are doing this, but I guess that the normal approach is to set the image paths in the stylesheets. If you do this a simple stylesheet exchange should be enough.

plata:
We need to change the icons in the separate views if the listener is triggered, don't we?

That's also what I found on the Internet. The question is if we can somehow make this work for the app categories as well. You can check #580 for a history of this.

plata:
Anyways let's continue this topic in a new issue (it's not really related to this PR).

plata:
ok to test

plata:
We should not mix the two topics in one PR.

madoar:
The latest commit contains my suggestion.
I've wrapped the `currentTheme` in an ObjectProperty and added an `addListener` method to detect changes to the theme at a later time. 
This new `addListener` method I've called in the SearchBox class, where I've added a listener that changes the clear button icon to the one from the current theme.

madoar:
I didn't mix the PRs. I just fixed the problem for this PR that the wrong clear button icon is shown, when the theme is changed. 
What remains is the question, do we want this solution?

plata:
It's a general issue that icons do not change when the theme changes so it's not really related to this PR.
We would have to add the listener to all places where icons are used. I would really love to somehow do this in CSS directly if we could find a way.

madoar:
It is possible, we would just need to add the url(...) parts directly to the stylesheet. 
This would make the file-paths static. My problem with this approach currently is, that I think this is ugly, if we don't replicate the icons to all theme folders, that every theme folder contains its own icon folder.

madoar:
Where is the app category normally themed, in the theme main css files? How is ApplicationDTO themed? Does it only contain an icon as a byte array?
If we could provide a css string for every ApplicationDTO we could do it the same way as for the second  problem you've described. We just need to load the css string as a stylesheet (which is described [here](http://stackoverflow.com/questions/24704515/in-javafx-8-can-i-provide-a-stylesheet-from-a-string)).

plata:
Currently, the fallback logic is implemented in ViewApps#populate. The ApplicationDTO contains the icon which is given in the repository (e.g. [here](https://github.com/PlayOnLinux/Scripts/blob/master/Games/icon.png)) as byte array. We've discussed using that icon in CSS in #580 without really finding a solution.

plata:
The main problem is that setStyle overrides the CSS. Maybe this is not the case with the link you posted.

madoar:
The Stylesheet solution overrides CSS too, but it should only override the CSS that was defined both in the older and in the newer CSS file, which is what you want, if I understand you correctly.
If you first add the ApplicationDTO stylesheet, then the default stylesheet, then the dedicated stylesheet (e.g. the dark stylesheet), it should work like you want. 
Then, iff an icon is not themed in the dedicated stylesheet, it will automatically fall back to the next level, which is the default stylesheet. If it's not defined there, which can happen with a few category icons, you fall back to the ApplicationDTO stylesheet. At least that is my understanding of CSS stylesheets, newer (later) definitions override older (earlier) ones.

plata:
Maybe we should try that. I would suggest that you undo the ObjectProperty stuff so we can merge this PR.

madoar:
Ok, then I'll revert my last commit

madoar:
The last commit is now reverted.

plata:
Could you open a new issue for the themed icons?

madoar:
I've just opened issue #647 for this, can you add the previous ideas?

Issue #650, opened 2017-03-11
Long waiting time until Applications in the "Apps" tab appear


madoar:
Currently it often takes a long time for me until the loading symbol in the center of the "Apps" tab disappears and the applications appear.

As far as I understand the current process, the content of the "Apps" tab is currently generated by searching the content of the github script repositories and local application sources.
Because I've added no custom application sources (repositories) to my installation, I guess the problem will be somewhere in the communication with the git-repositories.

plata:
Do you know where the problem is caused? As you said correctly, it fetches GitHub by default so we need to know if there's something wrong in the code or e.g. simply your Internet connection is slow.

madoar:
It shouldn't be my connection (100 Mbit/s connection).
I have this phenomenon every time I start POL 5. Is it possible, that the repositories are cloned every startup?

plata:
Just had the same idea. Yes, I think so (need to verify in the code). There is a caching mechanism but it's only per session.

madoar:
Yes I think so too. 
While we are at it, we should check if some modifications were done to the local clone. 
If such modifications were done it should either override the local changes or don't update to local repository. Maybe we can set this behavior in the Repositories Settings?

plata:
I think we should not mix a user checkout with the GitApplicationsSource checkout. So I would just overwrite any local changes.

plata:
Probably also better to use JGit instead of ProcessBuilder.

madoar:
That's ok for me we should just document it. 
Maybe it would be best if we document the whole process around the git script repositories in POL 5, while working on this issue?

plata:
Yes. The only quesion: where? Do we want to have this kind of documentation in the wiki?

madoar:
I would do an approach like they do for [angular-cli](https://github.com/angular/angular-cli).
For angular-cli they have the documentation in their main repository which enables them to use pull requests for docu updates. 
In regular intervals they update the wiki with the newest version from the main repo.

madoar:
How exactly does POL 5 interact with Git?

plata:
Git is just one possible ApplicationsSource. If you specify one GitApplicationsSource in your properties, it will clone the given repository and fetch the available scripts from it. @qparis wants to replace it by a WebApplicationsSource by default for the released version.

madoar:
[...]

PR #652, opened 2017-03-11
Using JGit to fetch GitApplicationsSources and caching them so they still exist after a restart of POL 5

madoar:
This PR switched the ProcessBuilder approach of using git to clone the script repositories with a JGit approach that is written in pure java, as talked about in issue #650. 
In addition this approach clones the repositories in a cacheDirectory, which is currently still tmp, and doesn't delete the clones repository folders after application exit/termination. This enables POL 5 to just load the still existing local repository folders, without the need to clone them again, at the next startup of POL 5.

What's missing here is the correct folder, where the clones repositories are to be kept. If we let them stay in tmp, they are going to be deleted at system shutdown.

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
@@ -18,45 +18,109 @@
 
 package org.phoenicis.apps;
 
-import org.phoenicis.apps.dto.CategoryDTO;
-
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.errors.RepositoryNotFoundException;
+import org.phoenicis.apps.dto.CategoryDTO;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 class GitApplicationsSource implements ApplicationsSource {
-    private final String gitRepositoryURL;
-    private final LocalApplicationsSource.Factory localAppsSourceFactory;
-    private List<CategoryDTO> cache;
+	private final static Logger LOGGER = LoggerFactory.getLogger(GitApplicationsSource.class);
 
-    GitApplicationsSource(String gitRepositoryURL, LocalApplicationsSource.Factory localAppsSourceFactory) {
-        this.gitRepositoryURL = gitRepositoryURL;
-        this.localAppsSourceFactory = localAppsSourceFactory;
-    }
+	private final String gitRepositoryURL;
+	private final LocalApplicationsSource.Factory localAppsSourceFactory;
+	private List<CategoryDTO> cache;
 
-    @Override
-    public synchronized List<CategoryDTO> fetchInstallableApplications() {
-        if (cache != null) {
-            return cache;
-        }
+	private static final String cacheDirectoryPath = System.getProperty("java.io.tmpdir");
 
-        try {
-            final File gitTmp = Files.createTempDirectory("git").toFile();
-            gitTmp.deleteOnExit();
-            new ProcessBuilder(Arrays.asList("git", "clone", gitRepositoryURL, gitTmp.getAbsolutePath()))
-                    .inheritIO()
-                    .start()
-                    .waitFor();
+	GitApplicationsSource(String gitRepositoryURL, LocalApplicationsSource.Factory localAppsSourceFactory) {
+		this.gitRepositoryURL = gitRepositoryURL;
+		this.localAppsSourceFactory = localAppsSourceFactory;
+	}
 
-            cache = localAppsSourceFactory.createInstance(gitTmp.getAbsolutePath()).fetchInstallableApplications();
-            return cache;
-        } catch (IOException | InterruptedException e) {
-            return Collections.emptyList();
-        }
-    }
+	@Override
+	public synchronized List<CategoryDTO> fetchInstallableApplications() {
+		if (cache != null) {
+			return cache;
+		}
 
+		File gitRepositoryLocation = new File(
+				GitApplicationsSource.cacheDirectoryPath + "/git" + this.gitRepositoryURL.hashCode());
 
+		LOGGER.info(String.format("Git-repository '%s' will be saved in '%s'", this.gitRepositoryURL,
+				gitRepositoryLocation.getAbsolutePath()));
+
+		/*
+		 * This value is true if a folder for the git repository exists,
+		 * otherwise it is false
+		 */
+		boolean folderExists = true;
+
+		// check that the repository folder exists
+		if (!gitRepositoryLocation.exists()) {
+			folderExists = false;
+
+			LOGGER.info(String.format("Creating new folder '%s' for git-repository '%s'",
+					gitRepositoryLocation.getAbsolutePath(), this.gitRepositoryURL));
+
+			if (!gitRepositoryLocation.mkdir()) {
+				LOGGER.error(String.format("Couldn't create folder for git repository '%s' at '%s'",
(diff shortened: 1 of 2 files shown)


phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
```diff
@@ -18,45 +18,109 @@
...
+			return cache;
+		}
+
+		File gitRepositoryLocation = new File(
+				GitApplicationsSource.cacheDirectoryPath + "/git" + this.gitRepositoryURL.hashCode());
+
+		LOGGER.info(String.format("Git-repository '%s' will be saved in '%s'", this.gitRepositoryURL,
+				gitRepositoryLocation.getAbsolutePath()));
+
+		/*
```
plata: This comment doesn't make sense (it's obvious from the code already).

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
```diff
@@ -18,45 +18,109 @@
...
+			LOGGER.error(String.format("Folder '%s' is no git-repository", gitRepositoryLocation.getAbsolutePath()), e);
+
+			this.cache = Collections.emptyList();
+		} catch (IOException e) {
+			LOGGER.error(String.format("An unknown error occured.", e));
 
+			this.cache = Collections.emptyList();
+		}
 
+		return this.cache;
```
plata: cache might be null here. It should be an empty list if something failed (I don't think it will work if you try to clone again later).

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
```diff
@@ -18,45 +18,109 @@
...
-            return Collections.emptyList();
-        }
-    }
+	private final static Logger LOGGER = LoggerFactory.getLogger(GitApplicationsSource.class);
+
+	private final String gitRepositoryURL;
+	private final LocalApplicationsSource.Factory localAppsSourceFactory;
+	private List<CategoryDTO> cache;
+
+	private static final String cacheDirectoryPath = System.getProperty("java.io.tmpdir");
```
plata: The cache directory should be set in the properties (Linux.properties, Mac OS X.properties).

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
```diff
@@ -18,45 +18,109 @@
...
+				gitRepositoryLocation.getAbsolutePath()));
+
+		/*
+		 * This value is true if a folder for the git repository exists,
+		 * otherwise it is false
+		 */
+		boolean folderExists = true;
+
+		// check that the repository folder exists
+		if (!gitRepositoryLocation.exists()) {
```
plata: you could write
```Java
boolean folderExists = gitRepositoryLocation.exists();

File: phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java [hunk 5]
```diff
@@ -18,45 +18,109 @@
...
+			folderExists = false;
+
+			LOGGER.info(String.format("Creating new folder '%s' for git-repository '%s'",
+					gitRepositoryLocation.getAbsolutePath(), this.gitRepositoryURL));
+
+			if (!gitRepositoryLocation.mkdir()) {
+				LOGGER.error(String.format("Couldn't create folder for git repository '%s' at '%s'",
+						this.gitRepositoryURL, gitRepositoryLocation.getAbsolutePath()));
+
+				return Collections.emptyList();
```
plata: Better set the cache and return the cache at the end.

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
madoar: Can you tell me how to receive enter the cache property? Or better make a new pull request where you set the correct cache directory?

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
madoar: This is not meant as a comment for the reader of the code, but as a debug message, in case something went wrong.

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
madoar: Not sure we want to do this here. I would suggest to forcefully terminate POL 5 here, because it doesn't have write-access of the caching folder in this case, or something similar.

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
madoar: How can cache be null here? In both the try and the catch clauses cache is set.

plata: You are right. Probably still had the `return Collections.emptyList();` from above in mind...

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
plata: Don't think that's a good idea. Assume you want to start an already installed app but cannot because the apps cannot be fetched. Also, you could have many other ApplicationsSources and not even care about the git one.

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
plata: I'm talking about:
```Java
/*
* This value is true if a folder for the git repository exists,
* otherwise it is false
*/
```

madoar: Oh you're right, this comment seems kind of useless.

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
madoar: The current idea is, that if the folder couldn't be created, simply return at this point an empty collection without setting the cache, because:

- This prevents the potentially git initialization code from being called (the code in the try/catch block).
- Because cache isn't set here it makes it possible for POL 5 to retry the generation of the directory at a later time (more detailed, after the method is called a second time).

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
plata: See #654.

phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
plata: I don't see any reason why it should work the second time.

madoar: Because maybe something was blocked by the underlying OS, which made the method return false, not sure how probable this is.

plata: I don't think it's probable. Anyways it doesn't really matter. It will just try to create the directory and fail again. Shouldn't have any bigger impact.

madoar: I think it depends on the reason why it failed. I can see a benefit for both strategies, to set the cache and don't set a cache, but I'm not sure which one is best.

plata: If it is possible that it works later on, I agree with you. However, I don't think that will happen.

PR #653, opened 2017-03-12
delay app search

### LINKED ISSUE #651: Slow performance of the search function in the "Apps" tab
If you're adding a search keyword in the "Apps" tab, after all applications are loaded (see #650) the input isn't added fluently, often it seems stuttering to stutter when inputting or deleting something from the search-field.

plata:
Delay the search by 0.5 seconds so it does not search for every character you type.

fixes #651


madoar:
I'm not sure if this is the right approach to solve this issue. What happens if you type for longer than 0.5 seconds, won't you still have the same problem?

plata:
The 0.5 seconds always refers to the last time you type. So it starts searching as soon as you stop typing. I tried it with the search string "age of empi". Works really well.

PR #657, opened 2017-03-12
Changing the implementation of MultipleApplicationsSource to be parallel

madoar:
This PR changes the implementation of MultipleApplicationsSource to be parallel through the use of the Java 8 Streaming API, see also #650. 
In addition this PR extracts the merging functionality of two `CategoryDTOs` from the `TeeApplicationsSource` class to a new interface called `MergeableApplicationsSource`, which enables this functionally to be directly used in `MultipleApplicationsSource`.

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/MergeableApplicationsSource.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/MergeableApplicationsSource.java
@@ -55,7 +55,7 @@ public interface MergeableApplicationsSource extends ApplicationsSource {
         leftApplication.getMiniatures().forEach(miniature -> mergeMiniaturesSet.add(ByteBuffer.wrap(miniature)));
         rightApplication.getMiniatures().forEach(miniature -> mergeMiniaturesSet.add(ByteBuffer.wrap(miniature)));
 
-        final List<byte[]> mergeMiniatures = new ArrayList();
+        final List<byte[]> mergeMiniatures = new ArrayList<>();
         mergeMiniaturesSet.forEach(miniature -> mergeMiniatures.add(miniature.array()));
 
         return new ApplicationDTO.Builder()
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/MergeableApplicationsSource.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/MergeableApplicationsSource.java
@@ -5,6 +5,7 @@ package org.phoenicis.apps;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


plata:
ok to test

madoar:
In the context of this PR, we can think about moving the for loop over multiple (two or more) application sources to the new `MergeableApplicationsSource` interface too, because both `MultipleApplicationsSource` and `TeeApplicationsSource` use this code.

plata:
If we have duplicate code, we should put that in a common class.

phoenicis-apps/src/main/java/org/phoenicis/apps/MergeableApplicationsSource.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.function.Function;
+
+import org.phoenicis.apps.dto.ApplicationDTO;
+import org.phoenicis.apps.dto.CategoryDTO;
+import org.phoenicis.apps.dto.ResourceDTO;
+import org.phoenicis.apps.dto.ScriptDTO;
+
+public interface MergeableApplicationsSource extends ApplicationsSource {
```
qparis: It looks like an Abstract class more than an interface

phoenicis-apps/src/main/java/org/phoenicis/apps/MergeableApplicationsSource.java
```diff
@@ -0,0 +1,95 @@
...
+	public default CategoryDTO mergeCategories(CategoryDTO leftCategory, CategoryDTO rightCategory) {
```
qparis: You need to reformat your code here

phoenicis-apps/src/main/java/org/phoenicis/apps/MultipleApplicationsSource.java
```diff
@@ -18,30 +18,71 @@
...
+		 * This step is needed because we need a mapping between the CategoryDTO
+		 * list and its application source, to preserve the order in the
+		 * reduction step
+		 */
+		final Map<ApplicationsSource, List<CategoryDTO>> categoriesMap = this.applicationsSources.parallelStream()
+				.collect(
+						Collectors.toConcurrentMap(source -> source, ApplicationsSource::fetchInstallableApplications));
+
+		if (applicationsSources.isEmpty())
+			return Collections.emptyList();
```
qparis: Please put braces here

phoenicis-apps/src/main/java/org/phoenicis/apps/MergeableApplicationsSource.java
madoar: What do you mean with "reformat" here? Do you mean I'm using the wrong formatter settings? Or that this shouldn't be an interface but a class?

madoar:
The parallelization happens in line 48 with the method call `parallelStream()` this makes java call `ApplicationsSource::fetchInstallableApplications`, which can take a lot of time, for every application source in parallel.
The merging step is still done sequentially in the for loop afterwards.

PR #658, opened 2017-03-12
Added the .gitignore files from the subfolders to the .gitignore file in the rootfolder

madoar:
Every time I start POL 5 in eclipse it generates .gitignore files in the module folders. Because they aren't needed I added them to the .gitignore file in the root project folder


plata:
I don't think we should do that. If you want to add one of those later one, you will hardly find the reason why it doesn't work. You should configure your eclipse in a way that it does not create these files. Maybe [this](https://stackoverflow.com/questions/24348657/are-gitignore-files-being-automatically-created-in-my-project) is helpful.

madoar:
Why should we want to add one of them later? One .gitignore file should be enough in my opinion. 
The problem with the approach in the link is, that it's an eclipse setting, which requires manual changes to eclipse. Alone this is no problem, but combined with other changes to eclipse that need to be done by hand (like the access-rules to deactivate the warning for javafx), it's one more chore I don't want to do, or easily forget, which leads to an accidental commit of these .gitignore files.

plata:
Sometimes it makes sense to add .gitignores not on top level. Also for me it's strange to add .gitignores to .gitignore because your IDE does strange stuff.

madoar:
In the case we sometime later want to add a `.gitignore` to a subfolder, can't we then just remove the line that "blocks" it from the `.gitignore` file in the rootfolder?
Normally I would agree with you, that we shouldn't add options and settings just to fix odd behaviors of an IDE. 
The problem is, that eclipse is one of the major IDEs that, in my eyes, should be well supported by POL 5 for development. In addition you can't completely configure eclipse, with the pom.xml file.
That's the reason why I suggested to add the sub-`.gitignore` files to the main `.gitignore` file as a precaution, in case one or more later developers accidentally adds them to the repository.

On another note:
@plata you said in a previous issue, that you use IntelliJ for development. Can you maybe write a short guide in the wiki how to setup intellij (how to project import, setup important configuration) that a user can start to develop on POL 5?

plata:
<https://github.com/PlayOnLinux/POL-POM-5/wiki/IntelliJ-IDEA>

I don't want to decide about the gitignores. @qparis  should do that.

qparis:
I don't like this idea because it forces us to update .gitignore when we add another submodule

madoar:
Ok then we'll close this, I'm currently switching to IntelliJ.

PR #661, opened 2017-03-15
renamed ApplicationsSource to RepositorySource

plata:
Renamed ApplicationsSource to RepositorySource to clarify that it contains the whole scripts repository including QuickScripts, Engines, Verbs etc.


qparis:
Why RepositorySource and not Repository instead?

madoar:
@qparis wouldn't `ScriptRepository` be more exact than `Repository`?

Issue #656, opened 2017-03-12
Add information about the repository, to which a script belongs


madoar:
Currently a user sees no information about the repository to which a script belongs.
This is currently no big problem, because each script exists only once. 
But this may change, if we have two repositories in the future, one with a stable version of a script, and another with a development version of a script, it is possible that two, or more scripts for the same application, with the same name, exist. 
If such a situation occurs, POL 5 needs to show the source repository of the script. I think the best location to show the source repository is both in the detailed view of the script and directly in the "Apps" tab for each visible script.

plata:
You can add this to the ApplicationDTO. A simple string should be enough (e.g. "git: https://...).

madoar:
You mean as a new variable inside the ApplicationDTO class?
How should this variable be shown best to the user?

plata:
Yes. In the AppPanel. Something like "Source: git: https://...".

madoar:
When implementing this, should we also add a field for the last modification of the script at the same time? As far as I remember we found no better solution that to add a modification date field to the script in https://github.com/PlayOnLinux/Scripts/issues/188.

plata:
I wouldn't do it in the same PR (you have to touch all scripts).

madoar:
This seems to be more work than I initially thought. The reason for this is, that `GitApplicationsSource` internally uses `LocalApplicationsSource` to process the folder after the repository has been locally cloned. 
The problem here is, that `LocalApplicationsSource` doesn't know, and truthfully spoken shouldn't care in my eyes, if the folder, it is processing, is a git repository or not. Therefore it can't set the git repository url as the source. The only thing it can set as the source is the folder name, which we don't want in case of a git repository. 

I would propose, that we move the folder-processing-parts from `LocalApplicationsSource` to a new class/interface. Then we can easily extend `LocalApplicationsSource` and `GitApplicationsSource` from the new class/interface, to mixin the ability to read the corresponding folder. After the categories have been fetched, we set the source information of the `ApplicationDTO`s, which is still easily doable because this is still done in the ApplicationSource `fetchInstallableApplications` method.  
What do you think?

plata:
Can't you just set the git repository URL after
```Java
this.cache = localAppsSourceFactory.createInstance(gitRepositoryLocation.getAbsolutePath())
					.fetchInstallableApplications();
```

madoar:
[...]

PR #664, opened 2017-03-16
Added source information for scripts in the GUI

madoar:
This PR adds source information to the scripts in the GUI, according to #656:
grafik
While doing this, I've also removed the cache variable from `GitApplicationsSource`, because we're using `CachedApplicationsSource` for caching purposes.

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/GitApplicationsSource.java
@@ -35,11 +35,11 @@ class GitApplicationsSource implements ApplicationsSource {
 
 	private final String gitRepositoryURL;
 	private final LocalApplicationsSource.Factory localAppsSourceFactory;
-	private List<CategoryDTO> cache;
 
 	private final String cacheDirectoryPath;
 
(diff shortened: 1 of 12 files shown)


qparis:
I think it makes the end user experience more complexe. I would not display it by default

madoar:
Hmm, do you mean we should add an option in the settings, where we can enable this?

madoar:
Ok, I've justs added an option in the "Settings" tab under "User Interface", where you can select if you want to see the source repository of the scripts or not.
To make this changes on this option visible without the need to restart POL 5, I additionally changed a the `Settings` - `SettingsManager` infrastructure. 
Previously the values inside the `SettingsManager` were initialized by Spring. The `SettingsManager` itself created then a `Settings` objects, which could be modified and read by the settings-GUI function and other functions. If you wanted to persist these `Settings` you needed to call a function in `SettingsManager` which took a `Settings` object that was then persisted. The problem with this approach is, that if you "spawn" two `Settings` objects via the load method in `SettingsManager`, they contain different information, that is not propagated to the other existing `Settings` objects. To solve this problem, the information is now directly queried and changed in `SettingsManager`, without the need to ask for a `Settings` object.
grafik

PR #665, opened 2017-03-17
Bugfix: Added missing default values for the viewsources property

madoar:
Added the missing default value for the new `viewsource` property.
See the comments on commit 4b08e122ae034459e614939d398a2f4a6a774387 for details.


plata:
ok to test

plata:
@qparis we should have a test in Jenkins to find errors like this before the merge.

qparis:
Which test?

plata:
I don't know. Some kind of system test which at least checks that the application does not crash on startup.

PR #668, opened 2017-03-18
Added a test for the initialisation of the AppConfiguration class

madoar:
Ok, this PR introduces a test for the Initialisation of the AppConfiguration.
The problem with this approach is, that it opens a gui-window, and I don't know if jenkins supports this...
If someone has a better approach or understands the API of testfx better than me, I'm open for suggestions :)

### DIFF
--- a/phoenicis-javafx/pom.xml
+++ b/phoenicis-javafx/pom.xml
@@ -28,6 +28,19 @@
 
     <artifactId>phoenicis-javafx</artifactId>
 
+	<build>
+		<plugins>
+			<plugin>
+	            <groupId>org.apache.maven.plugins</groupId>
+	            <artifactId>maven-surefire-plugin</artifactId>
+	            <version>2.19.1</version>
+	            <configuration>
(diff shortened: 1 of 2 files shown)


plata:
Sorry, never used testfx until now so I cannot help. Nevertheless I think it's a useful approach. We will find out in a minute if Jenkins can do this.

plata:
ok to test

plata:
Well, there you have it...

plata:
If you search for "headless testing JavaFX" you find some suggestions how to make it work. 
E.g.
- [Headless testing with JavaFx and TestFx](https://stackoverflow.com/questions/27403410/headless-testing-with-javafx-and-testfx)
- [Headless UI Testing with TestFX and JavaFX 8](https://ustesis.wordpress.com/2014/04/23/headless-ui-testing-with-testfx-and-javafx-8/)

madoar:
Ok, how about now?

plata:
Looks good. Let's @qparis also have a look at this.

madoar:
What I'm a bit sceptical about is, the reset jenkins does between two runs. Are the settings (including the created config files etc.) deleted after a run?

plata:
I have no idea.

Issue #667, opened 2017-03-18
Add indication of repository priority and a method to change the order of the repositories


madoar:
In the repository settings it's currently only possible to add and remove repositories.
![grafik](https://cloud.githubusercontent.com/assets/18488086/24072130/1caa55f8-0be2-11e7-89b5-cff45a69ad69.png)
What's missing are the following features:

- [X] An indicator that shows, what priority the repositories have. Is the first repository in the list the highest one or the last repository in the list?
- [X] In my opinion the "Add" button should add a repository as the one with the highest priority. Currently it seems to be added as the repository with the lowest priority.
- [X] In addition to the remove and add functionality, we need a way to change the priority of existing repositories. In my opinion we should either add an up and down button,that enables the user to move the selected repository up or down, or we could change the `ListView` component through a list component with drag and drop functionality.
- [X] The "Add" and "Remove" buttons should add/remove the applications in the corresponding repository without the need to restart POL 5.
- [X] The same should be true when changing the order of the repositories.

PR #670, opened 2017-03-20
Improvements to the Repository Settings

madoar:
This PR tries to improve some points from #667.
More detailed it adds the following features to the repositories list in the repositories settings:

- When viewing the repositories list in the repositories settings you can now see the priorities of the repositories, via a number in front of the repository
- The order of the repositories in the repository list is now changeable via drag and drop. The current behavior is, that directly after the order was changed the new repository order is saved to disk, maybe we want to change that and add a "Save"-button.

grafik

A still open problem is, that this PR breaks the text-color of the `ListCell`-content. I guess that this is because this PR changes the structure of the `ListCell`-content, which was previously simply a string. Now the text content is wrapped inside Labels, which are inside a GridPane, which is added to the ListCell as its graphic.

grafik


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/DragableRepositoryListCell.java
```diff
@@ -0,0 +1,121 @@
+/**
+ *
```
qparis: Please remove this

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/DragableRepositoryListCell.java
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * @author marc
+ *
+ */
+public class DragableRepositoryListCell extends ListCell<String> implements ChangeListener<Number> {
+
+	/**
+	 * What should be done after this cell has been dragged
+	 */
+	private Runnable onDragDone;
```
qparis: Can this be final

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/DragableRepositoryListCell.java
```diff
@@ -0,0 +1,121 @@
...
+import javafx.scene.layout.GridPane;
+
+/**
+ * @author marc
+ *
+ */
+public class DragableRepositoryListCell extends ListCell<String> implements ChangeListener<Number> {
+
+	/**
+	 * What should be done after this cell has been dragged
```
qparis: This comment is not needed, however, the corresponding javadoc can be written on the constructor

plata:
Why do we have numbers and an order? Isn't this duplicate? Also, I think that it is not clear what the meaning of the numbers is.
I would prefer to have just the order and add an explanatory text describing that you can change the priority by drag and drop.

madoar:
The problem currently is, that the order of the List isn't visible to the user. What is the repository with the highest and the lowest priority?
To visualize this I've added a number that everyone can understand. I agree with you that we should add a short legend for this.

madoar:
I've just fixed the css problem, and removed the center alignment of the priority.

About the priority information in general:
I'm not sure how to handle them in the final version. But I think we need to somehow show the user what priority which repository has. Therefore I would suggest merging this now and postponing the final decision how to do this to a later time.

plata:
I would like to have the legend in this PR already. Should not be a big deal.

madoar:
I was just too lazy to do that^^
grafik

grafik

madoar:
Just as a note: I think somewhere between the file reading and the loading of the repositories the list gets reversed. We should check where that happens (and if it really happens...), because I don't think we should reverse the list, this only makes working with the repository list more complex.

Issue #673, opened 2017-03-21
Build error


feanor12:
I tried to build 0a3edfe694e1d4115e22ba97ea3805d9124faa93 and I get an error, but 8cfc7851e844e62c82a7f447ba583d9a02dfeed5 works.

feanor12:
http://pastebin.com/BfYL1zRg

plata:
The same here. @madoar can you check this (it's most probably caused by  #668)?

plata:
Could be: <https://github.com/TestFX/TestFX/issues/364>

madoar:
@feanor12 which JDK are you using? OpenJDK or OracleJDK?
Can you test the linked project in https://github.com/TestFX/TestFX/issues/364 (https://github.com/Maxoudela/MonocleTest)?
In addition can you try to locally set a newer version for `openjfx-monocle` in the pom.xml in the javafx project, because I've set a quite old one?

feanor12:
java -version
openjdk version "1.8.0_121"
OpenJDK Runtime Environment (build 1.8.0_121-b13)
OpenJDK 64-Bit Server VM (build 25.121-b13, mixed mode)

feanor12:
I tried `<version>8u76-b04</version>`, but it still get an error.

feanor12:
If I run `mvn package` or `mvn test `on https://github.com/Maxoudela/MonocleTest I get an error.
http://pastebin.com/vTzVDr0q
http://pastebin.com/wiWrdf2U

feanor12:
If I change textfx to 4.0.6-alpha (local build of TestFX master) I can build POL5

madoar:
Ok, this means it's a combination of OpenJDK and testfx 4.0.5-alpha.
I suggest we simply comment the test out and comment it back in when testfx 4.0.6-alpha is officially released. What do you think?

feanor12:
Sounds like a solution. I don't know if it's possible in maven to add the git repository as dependency and build a snapshot.

madoar:
It's not possible to directly add a git repository as a dependency to maven.
But what's possible is to use a service, that builds the contents in the repository for you and distributes the result via maven.
For details look [here](http://stackoverflow.com/questions/20161602/loading-maven-dependencies-from-github) and [here](https://jitpack.io/#maven)

But this seems more or less like a hack.

lans98:
How can I build from 8cfc785?

madoar:
[...]

PR #676, opened 2017-03-22
Set the javafx configuration test as ignored

madoar:
This PR should temporarily solve #673 by ignoring the configuration test. This test should be included again at the time when the textfx team releases their next version 4.0.6-alpha.

It would be nice if someone with the problem described in #673 could test this PR.


plata:
ok to test

PR #677, opened 2017-03-22
Less integration at the example of the SearchBox component 

madoar:
This PR shows how to use less stylesheets (instead of normal css stylesheets) with POL 5.
To illustrate some benefits from less I've made the SearchBox component theme-changeable without the need to restart POL 5. At least between the default theme and the dark-breeze theme. 
If you have any questions or other comments please tell me.


phoenicis-javafx/pom.xml
```diff
@@ -31,115 +31,134 @@
...
+				<version>1.2.1</version>
+				<executions>
+					<execution>
+						<phase>process-sources</phase>
+						<goals>
+							<goal>compile</goal>
+						</goals>
+						<configuration>
+							 <sourceDirectory>src/main/resources</sourceDirectory>
+                     		 <outputDirectory>target/classes</outputDirectory>
```
plata: Please fix indentation.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
```diff
@@ -347,32 +357,34 @@
...
-	-fx-background-color: transparent;
+	.searchCleanButton {
+		.graphic("icons/mainwindow/general/edit-clear.png");
+		
+		-fx-border-radius: 0;
+		-fx-border-color: transparent;
+	}
 }
 
-/*******************************************************/
```
plata: Why did you delete this?

plata:
I like that variables are possible and the icons change automatically.

How do we solve the issue with default icons (e.g. for the app categories)? Can this be done with LESS?

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
madoar: By accident

madoar:
In LESS (and SASS/SCSS) it's possible to include other LESS (and SASS/SCSS) files.
If you include a Less file, containing a definition for class `.repository-icon` and define it afterwards in the Less file, that includes the other file, it will override the included definition. 
We can use this to create a file containing the fallback/default icons, and include this in the other theme-files.
I think this would be the most elegant solution.

What I've asked myself while working on this is why do we have these folder structures? Wouldn't it be easier to save all themes-files in the same folder, without using a subfolder for each theme? To identity the themes we simply give them a better name than main.css / main.less.
In addition what use do the `description.css` files have?

madoar:
In addition to this approach, I've also experimented with using SCSS as the stylesheet language. 
SCSS is, in my opinion, better and more clearly structured. 
I'll sometimes later upload my SCSS-approach too in a pull request, which enables us to better compare both approaches.
The only problem with using SCSS currently is, that it generates a `@charset "UTF-8";` at the beginning of the css files. Which is in itself nice to have, but sadly javafx returns an error for this line at the beginning of the file....

To solve this, I've created an issue at the repository of the maven-plugin I used to preprocess the scss files to css.

plata:
For the defaults: This would mean that the repository has to provide a stylesheet, right? I don't think we want that because the repository should only contain data. The only possibility to avoid this would be that we create a stylesheet on the fly based on the repository (after `fetchInstallableApplications()`). I would be really interested, if/how the fallback mechanism works out with LESS.

The theme folders also contain the icons. Therefore having all themes in one folder will become quite a mess.

The description.css is used by the `WebView` which shows the app description (when you click an app). See `AppPanel`.

madoar:
About the default:
What exactly belongs to these default values? Only a repository icon? 
Are these default values valid only for a single repository or for the whole application? 
Where can I find these default values? In what format are they available?
To do this statically and not at runtime, we need them in a css/less format, which we can either import in the less file or via the javafx api with `getStylesheets().add(...)`.

About the icons:
Don't we only have two icon sets? A dark one and a light one? Wouldn't it be more useful and intuitive to save them both at the same folder-level each in a subfolder called `dark` and `light`?

About the description.css:
Here we could think about adding these few definitions from the descriptions.css files to their corresponding main.css files. We simply need to give them a good class-name I guess.
What do you think?

plata:
default:
The repository contains a icon for every category e.g. [office](https://github.com/PlayOnLinux/Scripts/blob/master/Office/icon.png). This must be used whenever this icon is not available in the theme.

icons:
No. There is the default icon set and an icon set for Breeze. Every theme may have it's specific icon set or use the default one.

description:
I don't think that's possible. If you look how `WebView` is styled, you will find that it is not a class inside the theme CSS but a CSS file which is applied to the "browser".

madoar:
default:
Are the categories fixed, or is it possible to create new custom categories?
The point is that we would need a unique identifier for each category, if we want to do this with css, to assign the correct icon to each category.
If we can't ensure this, I think we'll need to add the category-icons programmatically, like we do currently.

icons:
I'm still not sure if we must bundle them in one folder together with the stylesheets, especially if we have a line in each stylesheet describing which icon folder is used.

description:
I agree

plata:
default:
The categories are definitely not fixed. Anyone who provides a repository can decide to add/delete any category he/she likes. If we have to do it in the code, I do not really see the point of changing the current implementation.

icons:
How would you do it.

madoar:
default:
We should still move most icon setting code to the css files to make it possible to change the theme, and its icons, without restarting POL 5. 
This is in most cases straight forward, like you see with the icon for the clear button in the `SearchBox`. The only exception to this would be the category-icons. To support the setting of the category-icons by css, we would need to create a css file for the category-icons for each repository, this file could be created programmatically too. 
This file doesn't only need to contain the default icon, but also the other icons in css format. Here is an example how this would look like for the games categroy(in less format):
```less
#game-repository {
    .graphic(".../game-icon.png");
}
```
I think, that even if we don't support the setting of the category-icons by css/less we would still benefit a lot.

icons:
I would create a folder containing all theme css files. In addition I would create a folder called `icons`, this can be done inside the theme css folder, which contains for each icon set its own subfolder named after the icon set.
I think this structure would be more clear to the viewer, at least it makes a better overview for the different icon sets possible, I think.

plata:
default:
I had implemented setting all icons except the category ones in the CSS already some time ago but we removed that again (you can find it in the PR history if you're interested). I would like to see how the CSS would be generated in the code.

icons:
I don't know. The icons really belong to the theme. Most themes simply do not use custom icons (yet).

madoar:
I think that the css for the category icons would be independent of the approach taken in this PR. The reason for this is, because I think we should use straight css for settings the category icons, because this is easier to pour in code, than to use the way through preprocessed less, which would need an additional preprocessing step made at runtime, if we don't want to add a css file to each of the repositories.
Can you tell me the PR, in which you implemented the CSS generation code?

icons:
I guess it's not really important. The only "bad" point of the current folder structure are the long and complex folder paths.

plata:
It is #587. If you look at the PR history before that, you will find how the whole stuff evolved. I just want to make sure that the concept really will work out in the end. Therefore I would like to see a proof of concept. You are right that this is not directly related to this PR but if it does not work with this PR, I'm not sure we should go for LESS.

plata:
@madoar have a look at #681

qparis:
I really like this PR. However, the only drawback is that we won't be able to run the app before running the maven goal in an IDE as we did before.

Do we have a solution for this?

madoar:
What do you mean? 
You don't need to run the app in an IDE, you just need to build the runnable (e.g. jar file) using maven to preprocess the less files to css files.
This can be done on any computer without IDE using `mvn package`.

qparis:
This is what I mean. Currently, I don't need to use maven to run the program with my IDE

madoar:
That is not entirely true. 
Your IDE uses maven internally, because you've imported the project as a maven project.
In eclipse, this is done using the m2eclipse plugin.
Therefore your IDE should do a lot of the maven lifecycle steps.

As an example, whenever I ran this PR in eclipse it automatically generated new css files for me.

plata:
@qparis I don't quite understand the issue. Are you talking about support for other build systems than maven?

plata:
@qparis can you take a decision if we want use Less or not? Having this PR laying around in an undecided state is a bit pointless. We should either go for it, or decide not to use Less and close it.

qparis:
Let's go for using less

plata:
Fine. @madoar you've heard it. If you fix the conflicts, we will merge this.

madoar:
I'm not sure if we should do this before of after we've addressed #711

plata:
I don't see why this should be related. #711 only changes the way we use the CSS/Less but not the CSS/Less itself.

madoar:
My guess was, that when we work on #711 we'll need to change/move most of the content of our css files (at least for the different resolutions we want to address), therefore it would be interesting to think about if we want to do this before or after the less migration.

plata:
I have fixed the resolutions stuff already (see #746). So I don't think that the CSS files will change a lot.

madoar:
Is it enough to change the font size?
Does this work for icons/images too?

plata:
As long as you specify the sizes in em it works.

plata:
@madoar can we have LESS before starting with issues like #782?

madoar:
We may have a small problem with this PR. 
It seems like IntelliJ is not automatically preprocessing the less file to css files.
Therefore you get a `Theme Default theme is not found!` error, when opening POL 5 in IntelliJ without running a `mvn test` or `mvn package` before.

I see two approaches we can take:

1. A developer, that wants to test POL 5 in his IDE needs to run `mvn test` or `mvn package` beforehand to get it starting
2. We add a fallback mechanism, that preprocesses the less files on the fly, if no css files were found

plata:
I don't see the problem. Before you can run, you have to build anyways. So as long as the build creates the CSS, everything should be fine.

madoar:
@plata This is about the requirement from @qparis: `This is what I mean. Currently, I don't need to use maven to run the program with my IDE`.
To build POL 5 you need to explicitly use maven before you can run the application in your IDE. I don't think this is what @qparis had in mind.

plata:
I don't understand this. How can you run the application without building right now?

madoar:
It seems like the Build process inside IntelliJ and I guess eclipse too (if you remember the issues we had with eclipse and the buildnumber-maven-plugin which still appeared every now and then when using eclipse) doesn't use the full maven build lifecycle.

plata:
Even if this is true, shouldn't we be able to do the generation in one of the build steps which are executed (after all, it must at least build the application somehow)?

qparis:
@madouar : this is what I had in mind indeed.
We can decide to detect and to alert the user if the build cycle is incomplete

madoar:
Ok, I've just pushed a new PR for the less integration, see #788.
This new PR enables us to use less files in addition to css files.
It also renames all `*.css` files we currently have to `*.less` files.

It would be nice if both of you (@plata and @qparis) could try this new PR to see if it fulfills your requirements.

madoar:
I guess there is nothing left open in this PR, right?

PR #678, opened 2017-03-24
SCSS integration

madoar:
In lieu of less, this PR uses SCSS for the javafx stylesheets.

This PR is currently unusable because SCSS adds `@charset "UTF-8";` to the beginning of the generated css files, which javafx seems to be unable to process.
Otherwise I think this approach is more beautiful and syntactically better readable that the less one in PR #677.


plata:
To be honest, I don't see much difference in the readability.

Do you think you can make it work?

madoar:
With readable I mean that some things are more intuitive. As an example the string concatenation in 
```scss
@mixin graphic($relative-url) {
	-fx-background-color: transparent;
	-fx-graphic: url($root-url + $relative-url);
}
```
where I can simply use the `+` operator. Another thing that improves the readability in my eyes is the `@mixin` and `@include` which show the reader that this is no normal css. In less it's easier to mistake a mixin definition for a normal css class definition.

I'm not sure I can make it work.
I've created an issue at https://github.com/GeoDienstenCentrum/sass-maven-plugin/issues/150, where I've asked if it's possible to prevent the generation of the `@charset` definition.
I guess we'll wait and see if it's possible to prevent this, if not we should stick with less I guess.

madoar:
Issue https://github.com/GeoDienstenCentrum/sass-maven-plugin/issues/150 was now closed with the reason, that this is not an option both in SASS itself and for the plugin.

So it seems like we can't use SCSS for the `@charset` reason.

plata:
Ok. Then please close this PR.

Issue #662, opened 2017-03-16
Clean up after a repository has been removed


madoar:
After a repository has been deleted/removed in the Repositories settings it may be needed to do additional cleanup work, depending on the type of removed Repository.
As an example, for git repositories it's needed to delete the cached repository folder, after the repository has been removed, to to free disk space.

PR #680, opened 2017-03-24
Add a delete method to all repositories

madoar:
This PR adds a `delete` method to all repositories.
This method is responsible for all cleanup work required after a repository has been removed.
In the case of a git repository this work consists of a deleted the previously cloned local repository folder.
For more details see #662.

The calling part of the newly created `delete` method is not included in this PR.
Before we can implement the call part of the delete method, we will need to make the loaded repositories modifiable at runtime. This includes both the addition and removal of repositories at runtime.

In case I forgot something don't hesitate to tell me. :)

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/GitRepository.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/GitRepository.java
@@ -18,17 +18,18 @@
 
 package org.phoenicis.apps;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.Collections;
-import java.util.List;
-
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.errors.RepositoryNotFoundException;
 import org.phoenicis.apps.dto.CategoryDTO;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 class GitRepository implements Repository {
 	private final static Logger LOGGER = LoggerFactory.getLogger(GitRepository.class);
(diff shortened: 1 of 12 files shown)


phoenicis-apps/src/main/java/org/phoenicis/apps/GitRepository.java
```diff
@@ -113,4 +114,23 @@
 
 		return result;
 	}
+
+	@Override
+	public void delete() {
+		File gitRepositoryLocation = new File(this.cacheDirectoryPath + "/git" + this.gitRepositoryURL.hashCode());
```
plata: You should store the repository location when it is created.

phoenicis-apps/src/main/java/org/phoenicis/apps/GitRepository.java
```diff
@@ -113,4 +114,23 @@
...
+
+		deleteFile(gitRepositoryLocation);
```
plata: Can you use `FileUtilities#remove`?

phoenicis-apps/src/main/java/org/phoenicis/apps/LocalRepository.java
```diff
@@ -68,6 +68,11 @@ private LocalRepository(String repositoryDirectory, ObjectMapper objectMapper) {
         return fetchCategories(categoryDirectories);
     }
 
+    @Override
+    public void delete() {
```
plata: Maybe the empty `delete()` should be implemented as default in the interface.

plata:
ok to test

qparis:
I don't understand why do we need to delete the content of a repository. 
How will it work with read only implementations like Web services?

madoar:
The `delete` method will be called when a repository is to be deleted.
It should perform cleanup work, if and only if this is necessary and possible.
The only repository that currently needs such clean up work is the git repository. Here we need to delete the local clone of the remote repository to free space.
For read only implementations like web services, I would suggest to use an empty "do nothing" delete method, which is the current default implementation in the `Repository` interface.

qparis:
I think that you want to rename the method by "onDelete" then

PR #681, opened 2017-03-25
default category icons in CSS

plata:
Creates a CSS file with the category icons from the repository. This approach enables icons switching during runtime.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -96,6 +100,30 @@ public void showAvailableApps() {
      */
     public void populate(List<CategoryDTO> categories) {
         Platform.runLater(() -> {
+            // set default category icons
+            try{
+                String css = "";
```
madoar: I think a `StringBuilder` or `StringBuffer` would be better

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -96,6 +100,30 @@ public void showAvailableApps() {
...
+                for (CategoryDTO category : categories) {
+                    css += "." + category.getName().toLowerCase() + "Button{\n";
```
madoar: As far as I understand our categories, there can be at most one category with the same class/name or?
If this is the case I think it's better to use a css-id instead of the class.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -96,6 +100,30 @@ public void showAvailableApps() {
...
+                    css += "-fx-background-image: url('" + category.getIcon() + "');\n";
+                    css += "}\n";
+                }
+                File temp = new File("tempfile.css");
```
madoar: How about using `Files.createTempFile(...)` to create the temp file, I think this would be better?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -96,6 +100,30 @@ public void showAvailableApps() {
...
+                temp.delete();
+                temp.createNewFile();
+                BufferedWriter bw = new BufferedWriter(new FileWriter(temp));
+                bw.write(css);
+                bw.close();
```
madoar: I don't really like Writers and BufferedWriters, how about using `Files::write` instead?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -96,6 +100,30 @@ public void showAvailableApps() {
...
+                // apply current theme again to fix hierarchy
+                final String shortName = themeManager.getCurrentTheme().getShortName();
+                final String url = String.format("/org/phoenicis/javafx/themes/%s/main.css", shortName);
+                final URL style = this.getClass().getResource(url);
+                getTabPane().getScene().getStylesheets().clear();
+                getTabPane().getScene().getStylesheets().addAll(temp.toURI().toString(), style.toExternalForm());
```
madoar: Watch out at which position you add this stylesheet. To make it work as a fallback solution I guess it would be better to insert it in an earlier position.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -96,6 +100,30 @@ public void showAvailableApps() {
      */
     public void populate(List<CategoryDTO> categories) {
         Platform.runLater(() -> {
+            // set default category icons
+            try{
```
madoar: I think it would be better to extract the whole css creation part for a list of categories to a new private method

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
plata: I thought about this. But I'm not sure when the the tmp file will be deleted again (it is also set in SettingsView).

madoar:
Just as a note:
I would first add a class to the css file, which contains the default icon, which will be used in case no custom one was specified. This could look like this:
```css
.default-category-icon {
    -fx-background-image: url('path/to/default/icon');
}
```
In addition I would add for each category that defines its own custom icon an id definition, like this: 
```css
#games-category-icon {
    -fx-background-image: url('path/to/games/icon');
}
```

Afterwards we need to add to each javafx category button the default category icon css-class.
In addition we set the id of every javafx category button, for which a custom icon is known.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
plata: It works like this. What do you mean by "earlier position"?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
plata: I'm don't think it should be done in the view at all.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
madoar: As far as the java documentation says the file is created in the default temporary-file directory. In my case this is `/tmp`. This directory gets deleted at system shutdown.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
madoar: Yes you're right, I think it would fit better in a controller.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
madoar: Normally later stylesheets should "override" earlier ones. With override I mean that duplicate definitions are overwritten by the later definition.
I'm not sure if this css file should have a higher or lower priority that the loaded theme-css file.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
plata: Ok. As long as there is no cleanup during the app runtime it should be fine.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
plata: The default CSS should have lower priority (and in fact has due to the order in `addAll`).

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -111,12 +117,8 @@ public void populate(List<CategoryDTO> categories) {
...
                     final LeftToggleButton categoryButton = new LeftToggleButton(category.getName());
                     categoryButton.setToggleGroup(group);
-                    final String resource = String.format("icons/mainwindow/apps/%s.png", category.getName().toLowerCase());
-                    if (themeManager.resourceExists(resource)) {
-                        categoryButton.setStyle("-fx-background-image: url('" + themeManager.getResourceUrl(resource) + "');");
-                    } else {
-                        categoryButton.setStyle("-fx-background-image: url('" + category.getIcon() + "');");
-                    }
+                    final String themeName = new String(category.getName().toLowerCase() + "Button");
+                    categoryButton.setId(themeName);
```
madoar: Maybe better use a `String.format` here like this:
```java
categoryButton.setId(String.format("%sButton", category.getName().toLowerCase()));
```

madoar:
Would it be possible to bind the ViewApps functionality to an `ObservableList`?
I ask this, because we should plan ahead to the open task left in #680, where we need to implement a way to add and remove repositories at runtime. This includes a refresh of the apps shown in `ViewApps`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
```diff
@@ -85,6 +92,33 @@ public void loadApps() {
...
+                themeManager.setDefaultCategoryIconsCss(defaultCategoryIconsCss);
+
+                // apply current theme again to fix hierarchy
+                final String shortName = themeManager.getCurrentTheme().getShortName();
+                final String url = String.format("/org/phoenicis/javafx/themes/%s/main.css", shortName);
+                final URL style = this.getClass().getResource(url);
+                this.view.getTabPane().getScene().getStylesheets().clear();
+                this.view.getTabPane().getScene().getStylesheets().addAll(defaultCategoryIconsCss, style.toExternalForm());
+            } catch (Exception e) {
+                e.printStackTrace();
```
madoar: I think we should log an exception here.

plata:
I tried that once but didn't manage to get it running with the miniature list properly. Feel free to try yourself.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
```diff
@@ -85,6 +97,38 @@ public void loadApps() {
...
+                Files.write(temp, css.getBytes());
+                String defaultCategoryIconsCss = temp.toUri().toString();
+                themeManager.setDefaultCategoryIconsCss(defaultCategoryIconsCss);
+
+                // apply current theme again to fix hierarchy
+                final String shortName = themeManager.getCurrentTheme().getShortName();
+                final String url = String.format("/org/phoenicis/javafx/themes/%s/main.css", shortName);
+                final URL style = this.getClass().getResource(url);
+                this.view.getTabPane().getScene().getStylesheets().clear();
+                this.view.getTabPane().getScene().getStylesheets().addAll(defaultCategoryIconsCss, style.toExternalForm());
```
madoar: I don't like our current way of interacting with the stylesheets.
Ideally I would like to integrate the handling/interacting in the `ThemeManager` or a new class called `StyleManager` or `CSSManager`.
This class should encapsulate the interaction with stylesheets via an `set()` method, which takes two arguments:

- The first argument is an enum, which tells the method what type of stylesheet you want to set. This value can `DEFAULT_THEME`, `USED_THEME`, `CATEGORY_ICON_THEME` and maybe some other values we'll need.
Depending on the given enum value the method then decides which value inside the stylesheet-list it needs to change.
- The second argument is a File object. This file object is the new stylesheet to be set.

What do you think?

plata: We can certainly discuss this. But please not in this PR as it is not really related to it. As quick reply: I agree that building the stylesheet URL here is not the most beautiful implementation.

plata:
Yes, please do so. I don't want to merge this myself without @qparis having a look on it before, however.

madoar:
@plata I've just added a new PR for the first part of the ObservableList issue, see #684.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
```diff
@@ -85,6 +97,38 @@ public void loadApps() {
...
         this.view.setOnSelectCategory(categoryDTO -> this.view.populateApps(categoryDTO.getApplications(), settingsManager));
+
+        this.view.setOnSetDefaultCategoryIcons(categoryDTOS -> {
+            // set default category icons
+            try {
+                StringBuilder cssBuilder = new StringBuilder();
+                for (CategoryDTO category : categoryDTOS) {
+                    cssBuilder.append("#" + category.getName().toLowerCase() + "Button{\n");
+                    String categoryIcon = category.getIcon();
+                    if (StringUtils.isEmpty(categoryIcon) || ! new File(new URI(categoryIcon)).exists()) {
```
qparis: The file will never exists if the icon is a remote file or if it is packaged inside a .jar fie

plata: That's true. So how should I handle this? Check only if the String is empty?

plata:
@qparis is this ok now?

qparis:
Sorry there are other conflicts. Can you solve it and merge it? 

Thanks

plata:
Why is this failing in Jenkins? I have no issues on my machine.

qparis:
ok to test

qparis:
ok to test

plata:
Merged as #699 has been merged with the same issue. So the failure has not been introduced by this PR.

PR #685, opened 2017-03-28
Changed AppsView to work with ObservableLists

madoar:
This PR connects with PR #684.
It finishes the change of the "Apps" tab to use `ObservableList`s. 

The only interface of AppsView to the outside world is now the populate method, through which you can set new categories. 
When you set new categories through the `populate` method the application will automatically detect this and extract the applications from the categories to set the content of the `MiniatureListWidget`.

The applications are now automatically sorted and filtered through the use of an FilteredList and and a SortedList. This has the additional benefit that the search speed is now much faster, doesn't stutter and all in all doesn't need the 0.5 second pause.

It's now possible to select a category and to add a search term at the same time.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/SearchBox.java
```diff
@@ -1,24 +1,29 @@
...
     private static final String cleanButtonIcon = "icons/mainwindow/general/edit-clear.png";
 
-    public SearchBox(ThemeManager themeManager) {
+    public SearchBox(ThemeManager themeManager, Consumer<String> onSearch, Runnable onClear) {
         this.searchField = new TextField();
 
         this.searchField.getStyleClass().add("searchBar");
         this.searchField.prefHeightProperty().bind(this.prefHeightProperty());
         this.searchField.prefWidthProperty().bind(this.prefWidthProperty());
+        this.searchField.setOnAction(event -> onSearch.accept(getText()));
```
plata: I think we should keep the delayed search. It might not be noticeable when the search is performed with every new/removed character on a reasonably fast computer. However, you might notice it on a slow computer. Moreover it is useless computation.

plata:
ok to test

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/SearchBox.java
madoar: The problem with the delayed search is, that it only delays it, which results in the same amount of work to be done only at a later time. 
The "new" search works different from the old search.
The old search refreshed the complete view after a search term has been entered or a category was selected. The new search does this through bindings. Therefore we set the applications/categories only once. Afterwards the search is done by applying a filter to the observable list, that contains the applications/categories. 
As far as I know this leads to a specific addition or removal of items and not to a rebuilding of the complete view.

plata: The last part of what you said is true. Nevertheless, the filter must be applied.

It is, however, not true that the delay causes the same amount of work. There is only one action after the delay has expired. So if you type and then type again while the delay is not expired, nothing will happen.

madoar: You're right, because we only have a single instance of `PauseTransition`. 

But I still think we don't need it, the new approach should be much more resource friendly than the last one.
In addition, as far as I remember, the reason why we added a delay in the first place was, because the input of longer search terms was stuttering, which should be no longer the case now.

plata: Yes. But as I said: You do not know how it performs on a slow computer and you do unnecessary calculations.

madoar: There you go, I've reinserted the delay.

madoar:
@plata, @qparis is this PR ok, or is there anything I should change? 

If not I would appreciate if this PR could be merged, because I'm already working on the next PR, which adds a new `RepositoryManager` to the project, that enables the addition and removal of repositories at runtime.

PR #688, opened 2017-03-31
Add repository manager

madoar:
This PR enables POL 5 to process the following repository changes in the repository settings at runtime:

- The addition of a new repository through the "Add" button
- The removal of one or more repositories through the "Remove" button
- The permutation of two repositories through drag and drop

In addition this PR removes the `EasyBind` dependency again, because it doesn't remember its mappings. If you remove an element from the source list it creates a new mapping for the removed element and forwards it to its own bindings to be removed. Because the new mapping and the old mappings don't need to be equal (like for the category-`ToggleButton`s as an example), it's impossible to remove the old elements because they are not equal to the to be deleted ones.

The new `MappedList` implementation is based on the one in `EasyBind`, but it has currently the problem that it has a potential memory leak, because its map, that contains all done mappings, is never cleaned up, when one or more elements are to be deleted. 
If someone has a good idea of how to do this please tell me.

In addition to the previous stated points I've added `equals` and `hashCode` implementations to all `Repository` classes. If you have some ideas on how to improve these implementations, I'm open for suggestions.

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/BackgroundRepository.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/BackgroundRepository.java
@@ -18,6 +18,8 @@
 
 package org.phoenicis.apps;
 
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.phoenicis.apps.dto.CategoryDTO;
 import org.phoenicis.apps.dto.ScriptDTO;
 
@@ -62,16 +64,29 @@ class BackgroundRepository implements Repository {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+        if (this == o) {
+            return true;
+        }
+
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
 
         BackgroundRepository that = (BackgroundRepository) o;
 
-        return delegatedRepository != null ? delegatedRepository.equals(that.delegatedRepository) : that.delegatedRepository == null;
+        EqualsBuilder builder = new EqualsBuilder();
+
+        builder.append(delegatedRepository, that.delegatedRepository);
+
+        return builder.isEquals();
     }
 
     @Override
     public int hashCode() {
-        return delegatedRepository != null ? delegatedRepository.hashCode() : 0;
+        HashCodeBuilder builder = new HashCodeBuilder();
+
+        builder.append(delegatedRepository);
+
+        return builder.toHashCode();
     }
 }
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/BackgroundRepository.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/BackgroundRepository.java
@@ -59,4 +59,19 @@ class BackgroundRepository implements Repository {
     public void setFilter(CombinedAppsFilter filter) {
         delegatedRepository.setFilter(filter);
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        BackgroundRepository that = (BackgroundRepository) o;
+
+        return delegatedRepository != null ? delegatedRepository.equals(that.delegatedRepository) : that.delegatedRepository == null;
+    }
+
+    @Override
+    public int hashCode() {
+        return delegatedRepository != null ? delegatedRepository.hashCode() : 0;
+    }
 }
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/CachedRepository.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/CachedRepository.java
@@ -18,6 +18,9 @@
 
 package org.phoenicis.apps;
 
+import com.jcraft.jsch.HASH;
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.phoenicis.apps.dto.CategoryDTO;
 
 import java.util.List;
@@ -50,16 +53,29 @@ class CachedRepository implements Repository {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+        if (this == o) {
+            return true;
+        }
+
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
 
         CachedRepository that = (CachedRepository) o;
 
-        return repository != null ? repository.equals(that.repository) : that.repository == null;
+        EqualsBuilder builder = new EqualsBuilder();
+
+        builder.append(repository, that.repository);
+
+        return builder.isEquals();
     }
 
     @Override
     public int hashCode() {
-        return repository != null ? repository.hashCode() : 0;
+        HashCodeBuilder builder = new HashCodeBuilder();
+
+        builder.append(repository);
+
+        return builder.toHashCode();
     }
 }
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/CachedRepository.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/CachedRepository.java
@@ -47,4 +47,19 @@ class CachedRepository implements Repository {
     public void clearCache() {
         this.cache = null;
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        CachedRepository that = (CachedRepository) o;
+
+        return repository != null ? repository.equals(that.repository) : that.repository == null;
+    }
+
+    @Override
+    public int hashCode() {
+        return repository != null ? repository.hashCode() : 0;
+    }
 }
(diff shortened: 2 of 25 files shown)


madoar:
The latest commit should fix the problem with the memory leak in `MappedList`, by rewriting `MappedList`. At least in my tests it performed as it should. It would be nice if someone else could test the MappedList of this PR. 

As a note, for all who wish to test this:
In this PR `MappedList` is used for a lot of things:

- Inside `MiniatureListWidget` (in the "Library", "Apps" and "Engines" tabs)
- Inside `LeftToggleGroup` (to show the categories at the left side in the "Apps" tab)

phoenicis-apps/src/main/java/org/phoenicis/apps/BackgroundRepository.java
```diff
@@ -59,4 +59,19 @@ public void getScript(List<String> path, Consumer<ScriptDTO> callback, Consumer<
     public void setFilter(CombinedAppsFilter filter) {
         delegatedRepository.setFilter(filter);
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
qparis: Don't forget the {} symbols, and please use EqualsBuilder and HashcodeBuilder as possible

phoenicis-apps/src/main/java/org/phoenicis/apps/FilterRepository.java
```diff
@@ -90,4 +90,27 @@ public void onDelete() {
     public void setFilter(CombinedAppsFilter filter) {
         this.filter = filter;
     }
+
+    @Override
+    public boolean equals(Object o) {
```
qparis: Same here

phoenicis-apps/src/main/java/org/phoenicis/apps/CachedRepository.java
```diff
@@ -43,4 +43,23 @@
     public void onDelete() {
         this.repository.onDelete();
     }
+
+    public void clearCache() {
+        this.cache = null;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
qparis: Same here

phoenicis-apps/src/main/java/org/phoenicis/apps/ClasspathRepository.java
```diff
@@ -151,6 +151,21 @@ private ScriptDTO buildScript(String categoryFileName, String applicationFileNam
                 .build();
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
qparis: Same remark

phoenicis-apps/src/main/java/org/phoenicis/apps/ConfigurableRepository.java
```diff
@@ -81,4 +81,22 @@ private Repository toRepository(String repositoryUrl) {
             return new NullRepository();
         }
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
qparis: Same remark

phoenicis-apps/src/main/java/org/phoenicis/apps/GitRepository.java
```diff
@@ -126,4 +126,19 @@ public void onDelete() {
 			LOGGER.error(String.format("Couldn't delete local git-repository at: '%s'", gitRepositoryLocation.getAbsolutePath()), e);
 		}
 	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) return true;
+		if (o == null || getClass() != o.getClass()) return false;
```
qparis: Same remark

phoenicis-apps/src/main/java/org/phoenicis/apps/LocalRepository.java
```diff
@@ -239,6 +239,21 @@ private ApplicationDTO unSerializeApplication(File jsonFile) {
         }
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
qparis: Sam remark

phoenicis-apps/src/main/java/org/phoenicis/apps/MultipleRepository.java
```diff
@@ -54,4 +55,51 @@
...
+			Repository[] newRepositories = new Repository[this.repositories.length - 1];
+
+			System.arraycopy(this.repositories, 0, newRepositories, 0, index);
+			System.arraycopy(this.repositories, index + 1, newRepositories, index, this.repositories.length - 1 - index);
+		}
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) return true;
```
qparis: Same remark

phoenicis-apps/src/main/java/org/phoenicis/apps/MultipleRepository.java
```diff
@@ -54,4 +55,51 @@
...
+		int oldIndex = newRepositories.indexOf(repository);
+
+		if (oldIndex >= 0 && toIndex >= 0 && toIndex < newRepositories.size()) {
+			Collections.swap(newRepositories, oldIndex, toIndex);
+
+			this.repositories = newRepositories.toArray(new Repository[0]);
+		}
+	}
+
+	public void addRepository(Repository repository) {
```
qparis: I think it would be cleaner to also add a onUpdated({MOVED, DELETE}) event on the repository interface

phoenicis-apps/src/main/java/org/phoenicis/apps/TeeRepository.java
```diff
@@ -55,4 +55,23 @@ public void onDelete() {
         this.leftRepository.onDelete();
         this.rightRepository.onDelete();
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
qparis: Same remark here

phoenicis-apps/src/main/java/org/phoenicis/apps/BackgroundRepository.java
madoar: The equals and hashCode methods were autogenerated by intellij.
I'm not sure how to use EqualsBuilder and HashcodeBuilder do you have an example for them, or can you change them?

madoar:
What do you mean by EventHandler? When and for what should an event be triggered?

phoenicis-apps/src/main/java/org/phoenicis/apps/BackgroundRepository.java
plata: You might want to take a look at this [stackoverflow](https://stackoverflow.com/questions/10912646/hashcodebuilder-and-equalsbuilder-usage-style).

qparis:
I think it is a matter of behavior: A repository could be updated and could notify a subscriber (your view) when it's done.
The pattern we are generally using in POL is to put a setter to a Runnable or a Consumer

phoenicis-apps/src/main/java/org/phoenicis/apps/BackgroundRepository.java
qparis: IntelliJ can also generate method with EqualsBuilder

@Plata, I'm was not talking about the reflective approach which is quite ugly

phoenicis-apps/src/main/java/org/phoenicis/apps/MultipleRepository.java
plata: This is a general question: Do we want objects do something or other classes do something with objects?

madoar:
I'm a friend of splitting different functionality to different classes. That's also the reason why I've set the `setFilter` method in `RepositoryManager` as deprecated. I think that the filter functionality should be handled outside the repository instance abstraction layer itself. For the javafx gui this happens in a FilterList for a console application we should be able to find another type of list that can handle this.
I would suggest to do the same with the repositories. The repository instance itself should only be responsible for the management and retrieval of all applications and scripts inside the repository.
Everything else, from the handling of the repositories at runtime to displaying them in a visual form to the user, should be done outside, because this depends on the application that uses them.

qparis:
As you said, the Repository instance should be responsible for the retrieval AND the management of DTOs. Therefore ideally, it should also be able to notify if the data has changed. (E.g, the LocalRepository should notify if a folder was created). 

To push the concept even further, the Tee and Multiple repositories should trigger the notification when :
* A child has sent a notification 
* A child was added or removed

Then, you have a clean design without needing an extra class

qparis:
Concerning the filters, you could create a FilterRepository that implements the same interface and the same logic

madoar:
The question then is how to bind the `EventHandler` to the correct repository. Let's say you want to add a "AddRepositoryEventHandler" to the `MultipleRepository` you would need to have a direct handle on it.
The old structure doesn't give you a handle to the repository because they are treated as `Repository` instances. To do this the easiest way would be still to create a RepositoryManager that has handles to all the different tiers of repositories, e.g. `CachedRepository`, `FilterRepository`, `MultipleRepository` and `BackgroundRepository`. This makes it the easiest to set the correct properties of the right repository.

madoar:
My latest commits introduce a new ExpandedList to POL 5, that makes the mapping process between List<CategoryDTO> to List<ApplicationDTO> quite easy. This new list component is now used in `ViewApps`. 
In addition I've fixed some other bugs that appeared when playing around with the repository settings (or more detailed with the addition and removal of repositories).

Later today, I'll rewrite the `equals` and `hashCode` methods to use EqualsBuilder and HashcodeBuilder.

madoar:
Ok, I've just finished rewriting the `equals` and `hashCode` methods to use `EqualsBuilder` and `HashCodeBuilder`.

From my side, I would consider this PR as finished.

PR #693, opened 2017-04-04
New repositories are now added as the ones with the highest priority

madoar:
This PR changes the priority of new repositories. 
Previously a new repository were added as the repository with the lowest priority, now they get the highest priority.
In addition this PR fixes a bug, that the repository order wasn't saved to disk after the order has been changed.

The current implementation (including the introduced changes in this PR) of the `MultipleRepository` and the `TeeRepository` contradict the information of the Repository page in the wiki, because currently it seems like the most important repository is at the beginning of the list and not at the end.
This behavior already existed before PR #657, which modified both `MultipleRepository` and `TeeRepository`. The root cause for this is line 92 in `MergeableRepository` where the parameters in the `mergeApplications` call are swapped. This was already the case before PR #657.
Should we stay with the current behavior, that the first entry in the list has the highest priority and modify the wiki to go along with this, or should we change the implementation, that it fits to the wiki?

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/MultipleRepository.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/MultipleRepository.java
@@ -78,6 +78,17 @@ class MultipleRepository extends MergeableRepository {
 		this.repositories = newRepositories;
 	}
 
+	public void addRepository(int index, Repository repository) {
+		Repository[] newRepositories = new Repository[this.repositories.length + 1];
+
+		System.arraycopy(this.repositories, 0, newRepositories, 1, index);
+		System.arraycopy(this.repositories, index, newRepositories, 1 + index, this.repositories.length - index);
+
+		newRepositories[index] = repository;
+
+		this.repositories = newRepositories;
+	}
+
 	public void removeRepository(Repository repository) {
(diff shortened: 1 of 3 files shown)


plata:
I changed the [wiki](https://github.com/PlayOnLinux/POL-POM-5/wiki/Repository).

phoenicis-apps/src/main/java/org/phoenicis/apps/MultipleRepository.java
```diff
@@ -78,6 +78,17 @@ public void addRepository(Repository repository) {
...
 
+	public void addRepository(int index, Repository repository) {
+		Repository[] newRepositories = new Repository[this.repositories.length + 1];
+
+		System.arraycopy(this.repositories, 0, newRepositories, 1, index);
+		System.arraycopy(this.repositories, index, newRepositories, 1 + index, this.repositories.length - index);
+
+		newRepositories[index] = repository;
+
+		this.repositories = newRepositories;
```
plata: Why not just use a list?

madoar: You're right, a list would be a better (easier) choice here. 
I think we should change the array in `MultipleRepository` to a list in a new PR, after this PR has been merged/closed.

PR #694, opened 2017-04-04
Removed getName in ScriptDTO

madoar:
This PR removes the `getName` method in `ScriptDTO`.
This has been done, because `getName` is a duplicate method of `getScriptName`.

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/ClasspathRepository.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/ClasspathRepository.java
@@ -134,7 +134,7 @@ class ClasspathRepository implements Repository {
             }
         }
 
-        Collections.sort(scriptDTOs, Comparator.comparing(ScriptDTO::getName));
+        Collections.sort(scriptDTOs, Comparator.comparing(ScriptDTO::getScriptName));
 
         return scriptDTOs;
     }
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/ClasspathRepository.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/ClasspathRepository.java
@@ -134,7 +134,7 @@ class ClasspathRepository implements Repository {
             }
         }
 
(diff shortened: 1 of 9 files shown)


plata:
Do we want to remove `getName` or `getScriptName`? `ApplicationDTO` and `CategoryDTO` also have a `getName` so I think it might make sense to rather keep `getName`.

madoar:
If we want keep `getScriptName` we should also rename the variable from `scriptName` to `name` in addition we maybe should then rename `scriptSource` to `source` too.

plata:
Sounds reasonable. After all it should be clear that it's about a script in the `ScriptDTO` class.

madoar:
Do you know if it has any following effects in the scripts repository, if we change the variable names from `scriptName` to `name` and `scriptSource` to `source`?

plata:
I don't think so. You just have to be careful when you change beans. In any case searching for "scriptName" in the repository will give you the result.

madoar:
Ok, I just found out the hard way, that we shouldn't rename/remove `scriptName`.
`scriptName` is used in the `script.json` files inside the script folders. This file is deserialized by the object mapper. Therefore the ScriptDTO object (or at least its Builder) needs to contain a `scriptName` attribute.

plata:
Or you change it in the complete chain.

madoar:
I'm not sure this is worth it, at least in the `script.json` files I think `scriptName` is a better name than simply `name`. This leaves us either with only renaming half the occurrences of `scriptName` or losing a better fitting name at half the occurrences. 
Therefore, and because I don't want to rename ~50 `scriptName` occurrences inside the scripts, I would suggest to simply use `scriptName`.

plata:
Fine for me. So the current version is working?

madoar:
At least the tests run for me. I couldn't really test it, because of #697 and https://github.com/PlayOnLinux/Scripts/pull/194.

plata:
In this case I would like to wait until @qparis is ok with #697.

qparis:
You can still have getName and just change the builder annotation but I don't find this really clean. I suggest keeping scriptName :-)

PR #697, opened 2017-04-05
Remove the access to the cached repository from outside the RepositoryManager

madoar:
This PR removes the access to the `CacheRepository` in the `RepositoryManager` from outside.
This is needed, because we can't be sure that the repository structure of POL 5 will look like it currently does at a later time.

On another note:
What is the reason for the `MockedRepository` class extending a `TeeRepository`? Wouldn't it be better to directly return an instance of `MockRepository`?

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/RepositoryManager.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/RepositoryManager.java
@@ -3,142 +3,72 @@ package org.phoenicis.apps;
 import org.phoenicis.apps.dto.ApplicationDTO;
 import org.phoenicis.apps.dto.CategoryDTO;
 import org.phoenicis.apps.dto.ScriptDTO;
-import org.phoenicis.multithreading.ControlledThreadPoolExecutorService;
-import org.phoenicis.tools.ToolsConfiguration;
-import org.phoenicis.tools.files.FileUtilities;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.Arrays;
 import java.util.List;
-import java.util.concurrent.ExecutorService;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
 
 /**
- * Created by marc on 31.03.17.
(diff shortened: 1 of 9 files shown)


phoenicis-apps/src/main/java/org/phoenicis/apps/RepositoryManager.java
```diff
@@ -57,8 +59,12 @@ public void setOnError(Consumer<Exception> onError) {
         this.onError = onError;
     }
 
-    public Repository cachedRepository() {
-        return this.cachedRepository;
```
qparis: The name of the attribute should be repository instead of cachedRepository. Theoretically, the manager could work with any kind of repository, could'nt it?

madoar: The implementation of the RepositoryManager is adjusted to the underlying repository architecture. This architecture looks currently like this:

> MultipleRepository -> FilterRepository -> CachedRepository -> BackgroundRepository

The internal variables for the 4 different repositories (which are each saved in their own variable) are `multipleRepository`, `filterRepository`, `cachedRepository` and `backgroundRepository`.

From outside we currently have possibility to indirectly access the FilterRepository, CachedRepository and BackgroundRepository through specialized methods.
The important part is, that the access is only indirectly possible, because the variables are marked as private and there exist no getter and setter methods.

Therefore I would suggest, if we change the repository structure inside the RepositoryManager at a later time, we change the variable names too.

madoar:
@qparis I've just added a `RepositoryManager` Interface, which we can use as a contract, where we define how to access and use a RepositoryManager. 
Our RepositoryManager implementation is now called `DefaultRepositoryManager`.
In addition I've tried my best to write a bit of JavaDoc to document what should happen in an abstract way, when you call one of the `RepositoryManager` methods.
What do you think, does this fulfill your requirements?

qparis:
ok to test

qparis:
Why does these test fail?

madoar:
I think this is because of the now existing merge conflicts. When you triggered the tests, there were no merge conflicts.
I'll fix the conflicts today, then we'll see :)

madoar:
Just as a note:
The tests in this PR are failing, because of PR #699.

Issue #625, opened 2017-02-28
Change engine for existing container


plata:
Allow to change the engine for an existing container (i.e. write new values to `phoenicis.cfg`). 
Currently, the container view is not really prepared to do this as it only shows information from a DTO. Probably there should be a WinePrefixContainerController which can populate a ComboBox in the WinePrefixContainerPanel with the versions obtained from WineVersionManager.

feanor12:
As a workaround I  added a second installer to the game folder which just installs the wine version I want
feanor12/Scripts@b634719

madoar:
@plata I've just tried to work on this issue, but couldn't find the class `WineVersionManager`. Can you tell me where this class is located?
In addition is there an easy way implemented, to access (read and write) the file `phoenicis.cfg`?

plata:
The `Wine...` has been renamed to `Engine...` by now (Wine is a special Engine). You can find the related classes in `phoenicis-engines`. The `phoenicis.cfg` is handled by `Functions/Engines/Wine/script.js`. The general idea for the engines is that you define your engines (e.g. Wine) in JS and plug them into the GUI in a generic way.

madoar:
Ok, I've got a first working version, that can at least show me a list of all available engine versions in a `ComboCox`.

I have multiple "problems" with my approach and the current Engines architecture:

- On big problem is, that my solution looks really ugly and kind of glued together. This is founded in the existing method structure, where method a calls b, which calls c, which calls d and so on until finally method `drawInformationTab` in WinePrefixContainerPanel is called.
- Another problem is that the engines part is uncommented. I must confess, I don't really understand what the different classes in the project `phoenicis-engines` mean and contain and how they belong together. I know for example that, that an `EngineCategoryDTO` contains multiple `EngineSubCategoryDTO`s, which in turn contain multiple `EngineVersionDTO`s, but what is the meaning of each of these classes?
- Another problem is the filtering of the engine versions, which fit to the selected container. Currently in my simple version, I show all engine versions. But I think we should only show versions, that compatible with the selected container. Let's say for example we have a container with a Wine 2.0 engine version. In this case I think we should only show wine versions.
- One missing part in my implementation is the saving of the engine version after it has been changed in the ComboBox. Here I'm missing some code that enables us to easily write to the container configuration.

madoar:
I've just created a new PR with my current code. See #698.

plata:
Let me try to explain `phoenicis-engines` a bit:
[...]

PR #698, opened 2017-04-07
Make the engine version of a container changeable

madoar:
This PR adds a ComboBox to the information tab in the `WinePrefixContainerPanel`, that enables the user to change the engine version of an existing container.
After opening the container details the engine version ComboBox will automatically show the currently selected engine version.

grafik

**This PR is currently only a proof of concept and shouldn't be merged, because it is incomplete and has a few design problems in my eyes!**
For more information see #625.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/ControllerConfiguration.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/ControllerConfiguration.java
@@ -89,7 +89,8 @@ public class ControllerConfiguration {
                 viewsConfiguration.viewContainers(),
                 containersConfiguration.backgroundContainersManager(),
                 viewsConfiguration.winePrefixContainerPanelFactory(),
-                containersConfiguration.winePrefixContainerController()
+                containersConfiguration.winePrefixContainerController(),
+                enginesConfiguration.wineVersionsFetcher()
         );
     }
 
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
@@ -18,30 +18,38 @@
 
 package org.phoenicis.javafx.controller.containers;
 
+import com.sun.xml.internal.ws.api.pipe.Engine;
+import javafx.application.Platform;
 import org.phoenicis.containers.ContainersManager;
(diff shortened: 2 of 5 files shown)


plata:
It should also be possible to select upstream/staging. The problem is that the available versions depend on the staging/upstream selection. Maybe it would be better to have a small dialog to change the version (1. select upstream/staging, 2. select version).

madoar:
@plata this depends on the `EngineCategoryDTO` that belongs to the `WinePrefixContainerDTO`.
If the container is not a wine container it doesn't make much sense to add an upstream/staging selection I think.

plata:
You should always have a subcategory. Maybe the possible selection depends on the engine (e.g. you should not switch from 32bit to 64bit in a Wine container). So probably the change should be implemented in the engine JS directly to take care of this. This would require a common `changeVersion` interface for engines. I'm not really sure currently how to design the container-engine interaction.

madoar:
Yes, I think that would be the best approach, to add a `changeVersion` interface/method to the engines. This method should also have the job of changing and saving the needed configuration files, I think.

plata:
It could be implemented similarly to engine install/delete, i.e. 
- add `ChangeEngineVersionUi` (see `EngineProgressUi`)
- implement `changeVersion` in JS
- call `changeVersion` of correct engine depending on container from Java

madoar:
How/Where would you call `changeVersion` from java? As far as I see `EngineProgressUi` has been added to nashorn as a function callable from javascript.
Can you do a PR for this?

plata:
Have a look at `EngineController#installEngine`. It's pretty much copy+paste. You need to know the engine category and then you pass the container as parameter to `changeVersion`. Everything else is JS. In a first step, `SetupWizard` can be used for the UI. We can then replace that with a lighter UI, which is special for `changeEngine`, later on.

madoar:
Are you sure this is enough? I think it's possible that we need to do some engine specific deinstallation work for the old engine first? Or would you do this in `changeVersion`?

plata:
Especially if it is engine specific, it has to be done in JS. So yes: I think it should be done in `changeVersion`. Maybe we should ask the user if the engine should be removed completely if it is no longer used (like we do with the container if you uninstall an app).

qparis:
We had a thought about making engine versions generic, but I think we should keep things really simple for the moment.
You can go ahead with your approach and we might improve it later.

madoar:
@plata, do you want to continue working on this? I'm not really sure how the engines work and how this should best be done. Therefore I would appreciate if you could continue on this.

PR #704, opened 2017-04-14
define icons in theme CSS

plata:
With this PR, icons are changed on the fly. It is no more necessary to restart Phoenicis to apply a different theme.
The engine icons are handled in a similar way as the app category icons (see #681).


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/ThemeManager.java
```diff
@@ -67,4 +68,22 @@ public void setDefaultCategoryIconsCss(String defaultCategoryIconsCss) {
     public String getDefaultCategoryIconsCss() {
         return defaultCategoryIconsCss;
     }
+
+    /**
+     * sets the CSS file containing the default engine icons from the repository
+     *
+     * @param defaultEngineIconsCss default engine icons from the repository
+     */
+    public void setDefaultEngineIconsCss(String defaultEngineIconsCss) {
```
madoar: Is this a `File` or a string directly containing the css?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/ThemeManager.java
```diff
@@ -67,4 +68,22 @@ public void setDefaultCategoryIconsCss(String defaultCategoryIconsCss) {
...
+        this.defaultEngineIconsCss = defaultEngineIconsCss;
+    }
+
+    /**
+     * returns the CSS file containing the default engine icons from the repository
+     *
+     * @return default engine icons from the repository
+     */
+    public String getDefaultEngineIconsCss() {
```
madoar: Same question as for `setDefaultEngineIconsCss`

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/ViewEngines.java
```diff
@@ -68,6 +68,7 @@ public void changed(ObservableValue<? extends Boolean> observableValue, Boolean
     private final CombinedEnginesFilter currentFilter = new CombinedEnginesFilter();
     private Consumer<CombinedEnginesFilter> onApplyFilter = (filter) -> {};
     private Consumer<EngineCategoryDTO> onSelectCategory;
+    private Consumer<List<EngineCategoryDTO>> onSetDefaultEngineIcons;
```
madoar: When will this consumer be triggered? What does it do? I would wish for some JavaDoc for all used consumers.

madoar:
In addition we need to document the css additions in the wiki. 
I think it's ok if we do this at a later time.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/ThemeManager.java
plata: It's the URI to the file. Same approach as for #681.

madoar: Shouldn't we use an `URI` object in this case? 
At least this should be added to the javadoc.

qparis:
ok to test

qparis:
Can you resynchronize your repo with the master?

madoar:
One more thing I just found:
Can you also move the search clear icon from SearchBox to the theme css?

plata:
@madoar I've been thinking about this. Just didn't do it because I did not want to cause conflicts with your LESS PR.

madoar:
Don't worry about the LESS PR. 
It's just a proof of concept. 
If we decide to go with it, I'll make a new PR, based on the then current master branch.
For this it would be helpful if we already moved all the icon urls from the java code to the css files.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/SearchBox.java
```diff
@@ -15,8 +15,6 @@
     private TextField searchField;
     private Button clearButton;
 
-    private static final String cleanButtonIcon = "icons/mainwindow/general/edit-clear.png";
-
     public SearchBox(ThemeManager themeManager, Consumer<String> onSearch, Runnable onClear) {
```
madoar: I think you can/should also remove the `ThemeManager` here.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
```diff
@@ -113,4 +83,42 @@ public void setOnAppLoaded(Runnable onAppLoaded) {
...
+            setDefaultCategoryIcons(categoryDTOS);
+            this.view.populate(categoryDTOS);
+        });
+    }
+
+    private void setDefaultCategoryIcons(List<CategoryDTO> categoryDTOS) {
+        try {
+            StringBuilder cssBuilder = new StringBuilder();
+            for (CategoryDTO category : categoryDTOS) {
+                cssBuilder.append("#" + category.getName().toLowerCase() + "Button{\n");
```
qparis: You need to chain .append here?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
```diff
@@ -113,4 +83,42 @@ public void setOnAppLoaded(Runnable onAppLoaded) {
...
+                String categoryIcon = category.getIcon();
+                if (StringUtils.isEmpty(categoryIcon)) {
+                    cssBuilder.append("-fx-background-image: url('/org/phoenicis/javafx/views/common/phoenicis.png');\n");
+                } else {
+                    cssBuilder.append("-fx-background-image: url('" + categoryIcon + "');\n");
+                }
+                cssBuilder.append("}\n");
+            }
+            String css = cssBuilder.toString();
+            Path temp = Files.createTempFile("defaultCategoryIcons", ".css").toAbsolutePath();
```
qparis: Will the file be deleted on exit?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
```diff
@@ -113,4 +83,42 @@ public void setOnAppLoaded(Runnable onAppLoaded) {
...
+            themeManager.setDefaultCategoryIconsCss(defaultCategoryIconsCss);
+
+            // apply current theme again to fix hierarchy
+            final String shortName = themeManager.getCurrentTheme().getShortName();
+            final String url = String.format("/org/phoenicis/javafx/themes/%s/main.css", shortName);
+            final URL style = this.getClass().getResource(url);
+            this.view.getTabPane().getScene().getStylesheets().clear();
+            this.view.getTabPane().getScene().getStylesheets().addAll(defaultCategoryIconsCss,
+                    themeManager.getDefaultEngineIconsCss(), style.toExternalForm());
+        } catch (Exception e) {
```
qparis: In general, you should avoid catching generic Exceptions

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
madoar: I previously thought, that the file (if it is created afterwards) is automatically deleted at system shutdown, because at least on my OS it is created in the `tmp` directory, which should be emptied at shutdown.

But you're right the API documentation doesn't specify if the file is deleted.
I found http://stackoverflow.com/questions/28752006/alternative-to-file-deleteonexit-in-java-nio, which may help here.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
plata: Don't think so. The compiler should do that automatically (see [here](https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java)).

madoar:
@qparis, @plata can this be merged?
I'm currently doing a bit of refactoring, but to do it right it would be beneficial if this is merged first.

plata:
As @qparis approved it, I think it's ok if I merge it now.

Issue #707, opened 2017-04-14
Check why the test "testListPermutation" fails 


madoar:
We should investigate why the test `testListPermutation` fails, so that we don't need to ignore the test long term.

PR #708, opened 2017-04-14
fixed MappedListTest

VanCoding:
Hi guys!

The current version does not build (at least on Arch Linux) because of a missing import in MappedListTest.java.

I've fixed that and now it works.


qparis:
ok to test

plata:
I still wonder why this worked locally for @madoar and me.

madoar:
I think something may be wrong with the java version in jenkins.
That's one of the reason why I've opened #707, I think we should reopen #707.

PR #710, opened 2017-04-14
Added a "refresh Repositories" button to the repository settings

madoar:
This PR introduces a new "refresh Repositories" button to the repository settings.

grafik

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/ViewSettings.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/ViewSettings.java
@@ -25,9 +25,7 @@ import javafx.event.ActionEvent;
 import javafx.geometry.Insets;
 import javafx.geometry.VPos;
 import javafx.scene.control.*;
-import javafx.scene.layout.GridPane;
-import javafx.scene.layout.HBox;
-import javafx.scene.layout.VBox;
+import javafx.scene.layout.*;
 import javafx.scene.text.Text;
 import org.phoenicis.apps.RepositoryManager;
 import org.phoenicis.javafx.views.common.TextWithStyle;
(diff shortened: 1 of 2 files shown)


plata:
Can't we do that automatically? E.g. set a flag and update the repositories when the "Apps" tab is opened the next time.

madoar:
I'm not sure we want that, because the refresh may take a while until it's done.
Especially if you add additional repositories.

plata:
That's true. But on the other hand you will not add repositories if you do not want to use them. Therefore we can assume that any change in the repository settings shall trigger an update. To avoid updating several times because you add/delete multiple repositories, I suggested the flag.

madoar:
I think you've misunderstood for what I would use this button.
I would use it if I know or think to know, that the remote repository has a more up to date version of a script than my local copy.
Then I can use this button to refresh the data without the need to restart POL 5.

plata:
Ah ok. I guess we need some explanatory text for this then.

madoar:
What would you suggest?

plata:
Maybe something like: "Fetch updates for the repositories to retrieve the newest script versions"? It has to be clear but at the same time understandable for the average user.

madoar:
Should this be added as a legend text or as a tooltip?

plata:
I would suggest a two column layout.

madoar:
Ok, I just changed the code to use a two column layout:
grafik

plata:
Shouldn't the legend for the priority be above?

madoar:
I thought it's better to first show all the different settings the user can change, then at the bottom we show the legend for all these settings.
Currently we have only one point to show for the legend, therefore it only contains the priority information.

If needed I can change the order, that the priority information is shown directly below the repositories.
What do you think?

plata:
I think the legend belongs to the repository list.

madoar:
Like this?
grafik

plata:
Yes, that's better.

madoar:
@plata, @qparis is there anything left to do for this PR?
Otherwise it would be nice to see this merged, because I want to refactor the settings tab soon :)

PR #715, opened 2017-04-21
Refactoring the SideBar in the Library Tab (ViewLibrary)

madoar:
This PR moves the code responsible for populating and maintaining the sidebar in the Library Tab to a new class `ApplicationSideBar`.
This is better maintainable than having all parts making up the Library tab in one single class.

In addition I think that we should do this (moving the sidebar to its own class) for all tabs available in POL 5.


plata:
Good idea. I would prefer to call it `LibrarySideBar` so it gets more clear where it belongs to.

Do you think it makes sense to add a `SideBar` class with common functionality (just an idea, did not really think about it)?

madoar:
Renaming `ApplicationSideBar` to `LibrarySideBar` is no problem.

I've thought about introducing a `SideBar` class as an abstract parent class. 
The reason why I didn't add it yet is because there already exists a `LeftSideBar` class.
As far as I understand the current code the sidebar classes will be added to a `LeftSideBar` object.
In addition this would require other changes to the `MainWindowView` class too. 

Another point I'm not really sure about is, what common functionality do the different sidebars have?

plata:
As I said, just an idea. I was thinking about some object which contains a search box and is structured in categories with buttons. So that you have the same layout/structure everywhere. Not sure it makes sense.

madoar:
I'm not sure if every sidebar needs a search field.
For the settings as an example it wouldn't be really useful I think.

I agree we should add a class that signifies that a component class is to be used as a sidebar.
But the only think that would be needed for this is a class extending VBox, I think...

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/LibrarySideBar.java
```diff
@@ -0,0 +1,145 @@
...
+    // consumers called when a shortcut should be run, stopped or uninstalled
+    private Consumer<ShortcutDTO> onShortcutRun;
+    private Consumer<ShortcutDTO> onShortcutStop;
+    private Consumer<ShortcutDTO> onShortcutUninstall;
+
+    // consumers called when a script should be run or a console be opened
+    private Consumer<File> onScriptRun;
+    private Runnable onOpenConsole;
+
+    public LibrarySideBar(String applicationName) {
```
plata: Could you add JavaDoc to the methods?

madoar: Which methods do you mean? 
The callbacks or the methods inside the class?
About the callbacks: I myself don't know the use of all the callbacks :/

plata: The methods in the class (at least the public ones).

PR #718, opened 2017-04-22
wrap side bar title

plata:
Before:
before

Now:
wrap


qparis:
Can you send the whole screenshot please?

plata:
Before:
before

Now:
wrap

qparis:
What if it takes 10 lines to show the name of the game?

plata:
Then it will take 10 lines. To be hones I think that this is really a rare case which will look strange anyways. Many games need two lines so at least those should look ok.

PR #720, opened 2017-04-22
Refactoring and improving the engine side bar

madoar:
This PR moves the sidebar code from the `EnginesView` to its own class `EngineSideBar`.
In addition it introduces the `LeftToggleGroup` class in the engines sidebar to show the different engine categories.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EngineSideBar.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EngineSideBar.java
@@ -4,10 +4,8 @@ import javafx.beans.binding.Bindings;
 import javafx.beans.property.BooleanProperty;
 import javafx.collections.ObservableList;
 import javafx.scene.control.CheckBox;
-import javafx.scene.control.Toggle;
 import javafx.scene.control.ToggleButton;
 import javafx.scene.layout.VBox;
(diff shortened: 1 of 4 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EngineSideBar.java
```diff
@@ -0,0 +1,122 @@
...
+    private CheckBox notInstalledCheck;
+
+    private CombinedEnginesFilter currentFilter;
+
+    private Consumer<CombinedEnginesFilter> onApplyFilter = (filter) -> {};
+
+    private Consumer<EngineCategoryDTO> onCategorySelection;
+    private Runnable onAllCategorySelection;
+
+    public EngineSideBar() {
```
plata: some Javadoc please ;)

plata:
Does the `LeftToggleGroup` change the look?

madoar:
It changes it in some small aspects. The old (current) implementation uses `LeftButton`s for the engine selection in the left sidebar. The new implementation, which uses the `LeftToggleGroup`, uses `LeftToggleButton`s. These `LeftToggleButton`s change their background color after they are selected, see the "Apps" tab to see what I mean.
grafik

plata:
Ok. Yes makes sense to use toggle buttons.

Issue #719, opened 2017-04-22
Error when deselecting "Installed" or "Not installed" in the Engines Tab 


madoar:
When deselecting "Installed" or "Not installed" in the sidebar of the "Engines" tab I get the following exception:
```
Exception in thread "JavaFX Application Thread" java.lang.UnsupportedOperationException
	at java.util.Collections$UnmodifiableCollection.removeAll(Collections.java:1068)
	at org.phoenicis.engines.CombinedEnginesFilter.apply(CombinedEnginesFilter.java:67)
	at org.phoenicis.javafx.controller.engines.EnginesController.lambda$null$0(EnginesController.java:65)
	at com.sun.javafx.application.PlatformImpl.lambda$null$173(PlatformImpl.java:295)
	at java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$174(PlatformImpl.java:294)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)
	at com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
	at com.sun.glass.ui.gtk.GtkApplication.lambda$null$49(GtkApplication.java:139)
	at java.lang.Thread.run(Thread.java:745)
```

plata:
When did this break? It used to be working. Can't even debug: As soon as I place a breakpoint before that line, the error disappears. Maybe some threading issue?

madoar:
I don't know when it broke. I just know that it's not connected to #720.
I guess it's connected to the used list. If you use a list generated by `Arrays.asList()` you can't add items to the list. If you do you get the previous exception.

plata:
Could it be related to #684?

plata:
At least ea27e30d0171eded8625ae6cbe9eb5691339fbc5 works.

plata:
The `unmodifiableList` is in `EngineCategoryDTO`:
```Java
this.subCategories = Collections.unmodifiableList(builder.subCategories);
```

Probably it would make sense to use an `ObservableList` like for the apps for the filter (see #685). What do you think?

madoar:
I would prefer using `ObservableList`s only in the javafx projects, because they shipped with javafx.

plata:
Yes. Would you like to try it (you did it for the apps already so you know what has to be done)?

madoar:
[...]

PR #723, opened 2017-04-23
Refactoring the "Engines" Tab, moving the engines sub category tab to a new class and adding a javafx engine filter implementation

madoar:
This PR fundamentally refactors the code responsible for the "Engines" tab.
See the title for more details.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/ControllerConfiguration.java
```diff
@@ -143,4 +143,8 @@ public SettingsController settingsController() {
                 settingsConfiguration.settingsManager()
         );
     }
+
+    public String getWineEnginesPath() {
+        return this.wineEnginesPath;
```
plata: I think we should not consider special engine types on this level. I would prefer to build the specific engine path based on `application.user.engines` later on.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EngineSubCategoryTab.java
```diff
@@ -0,0 +1,137 @@
...
+        this.setContent(this.engineVersionsView);
+    }
+
+    /**
+     * This method populates the engines version miniature list widget and binds the filtered engine versions list to it.
+     */
+    private void populate() {
+        this.engineVersionsView = MiniatureListWidget.create(
+                engineVersionDTO -> MiniatureListWidget.Element.create(engineVersionDTO,
+                        Files.exists(Paths.get(wineEnginesPath, engineSubCategory.getName(), engineVersionDTO.getVersion()))),
```
plata: Regarding the comment above, this should be:
```Java
Paths.get(enginesPath, engineCategory.getName(), engineSubCategory.getName()...)
```

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginesFilter.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * The search term entered into the search field
+     */
+    private String searchTerm = "";
+
+    /**
+     * Are only installed engines searched
+     */
+    private boolean onlyInstalled = true;
```
plata: The name is a bit misleading: it suggests that you can filter either installed or uninstalled (because of the "only"). If I understand this correctly, you can also search both. Maybe something like `showInstalled` would be better.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginesFilter.java
```diff
@@ -0,0 +1,88 @@
...
+     */
+    private boolean onlyUninstalled = true;
+
+    /**
+     * Constructor
+     *
+     * @param engineSubCategory The engine sub category used for this filter
+     * @param wineEnginesPath   The path to the installed wine engines
+     */
+    public EnginesFilter(EngineSubCategoryDTO engineSubCategory, String wineEnginesPath) {
```
plata: Same as above: the filter should be agnostic of the engine category.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/ViewEngines.java
```diff
@@ -159,8 +155,14 @@ private void showEngineDetails(EngineDTO engineDTO) {
         showRightView(currentEnginePanel);
     }
 
-    public void setOnSelectCategory(Consumer<EngineCategoryDTO> onSelectCategory) {
-        this.onSelectCategory = onSelectCategory;
+    private void processFilterText(String filterText) {
```
plata: Maybe we should move this functionality into the `SearchBox` class? (not in this PR)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftToggleGroup.java
```diff
@@ -94,4 +94,14 @@ private LeftToggleGroup(String name, Supplier<ToggleButton> allButtonSupplier, F
     public void selectAll() {
         this.allButton.ifPresent(button -> button.setSelected(true));
     }
+
+    public void select(E element) {
+        int elementIndex = this.elements.indexOf(element);
```
plata: Why don't you pass the index directly?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/ViewEngines.java
madoar: I'm not sure what you mean with functionality. This method, `processFilterText`, is passed to the `SearchBox` as a consumer to be called when an entry has been made.
We can't handle the filter text processing the same for every tab, because there are some slight differences.

plata: Wouldn't it be possible to pass only the method which shall be called when the search shall start such that the handling of the `pause` and `filterText.toLowerCase()` is inside `SearchBox`?

madoar: I don't know if we want to add the `pause` to the `SearchBox`. Currently we're only using `pause` this, because the lists we're filtering with the search term are quite long which makes it work intensive.

madoar:
Before I forget it:
What is still open is the question what to do with the old `CombinedEnginesFilter` and the other related classes in `phoenicis-engines`. 
Are they still needed/required?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/ControllerConfiguration.java
plata: That's not what I had in mind. I think you should replace `this.wineEnginesPath` with `this.enginesPath` which shall be `application.user.engines`. Then you can use `enginesPath + engineCategory.getName()` to determine the path for the special engine (like Wine).
I know that this might interfere with `application.user.engines.wine` but I think that's a different topic. In my opinion, we should just remove those engine specific properties. If we can't do that, we have to create a generic interface which allows you to query the installation path from JS.

plata:
If you do not use the classes, remove them (we have them in the git history in case someone needs them again later).

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/ControllerConfiguration.java
madoar: I'm not sure I understand how you would like to have it. Can you change it in another PR to have like you want to?

plata: Ok.

plata:
Could you find a title which is more concise? We cannot use this as commit message.

madoar:
Refactoring of "Engines" tab + fixing engine filter?

plata:
fixes #719

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/ControllerConfiguration.java
plata: See #724

PR #725, opened 2017-04-23
Refactoring the "Settings" tab

madoar:
This PR refactors the "Settings" tab. This includes moving some functionality to new classes.

In this PR I've also tried to use `-fx-background-image` for the refresh icon in the "Refresh Repositories" button. The problem is that it didn't want to work, which is the reason why I'm still using `-fx-graphic`.
It would be nice if someone else could change `-fx-graphic` with `-fx-background-image`, because it seems like I don't get it working with `-fx-background-image`.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/AboutPanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/AboutPanel.java
@@ -1,0 +1,125 @@
+package org.phoenicis.javafx.views.mainwindow.settings;
+
+import javafx.scene.control.Hyperlink;
+import javafx.scene.control.Label;
+import javafx.scene.layout.GridPane;
+import javafx.scene.layout.VBox;
+import javafx.scene.text.Text;
+import org.phoenicis.javafx.views.common.TextWithStyle;
+import org.phoenicis.tools.system.opener.Opener;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import static org.phoenicis.configuration.localisation.Localisation.translate;
+
+/**
+ * This class represents the "About" settings category
+ *
+ * @author marc
+ * @since 23.04.17
+ */
+public class AboutPanel extends VBox {
+    private ApplicationBuildInformation buildInformation;
+    private Opener opener;
+
+    private Text title;
+
+    private GridPane aboutGrid;
+
+    private Text nameDescription;
+    private Label nameLabel;
+
+    private Text versionDescription;
+    private Label versionLabel;
+
+    private Text gitRevisionDescription;
+    private Hyperlink gitRevisionHyperlink;
+
+    private Text buildTimestampDescription;
+    private Label buildTimestampLabel;
+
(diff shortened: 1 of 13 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/AboutPanel.java
```diff
@@ -0,0 +1,125 @@
...
+    private Text versionDescription;
+    private Label versionLabel;
+
+    private Text gitRevisionDescription;
+    private Hyperlink gitRevisionHyperlink;
+
+    private Text buildTimestampDescription;
+    private Label buildTimestampLabel;
+
+    public AboutPanel(ApplicationBuildInformation buildInformation, Opener opener) {
```
plata: Javadoc please

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/ViewSettings.java
```diff
@@ -18,341 +18,73 @@
...
-		themeManager.setCurrentTheme(theme);
-		final String shortName = theme.getShortName();
-		final String url = String.format("/org/phoenicis/javafx/themes/%s/main.css", shortName);
-		final URL style = this.getClass().getResource(url);
-		getContent().getScene().getStylesheets().clear();
-		getContent().getScene().getStylesheets().addAll(themeManager.getDefaultCategoryIconsCss(),
-			themeManager.getDefaultEngineIconsCss(), style.toExternalForm());
-	}
+    private void initSelectSettingsPane() {
+        this.selectSettingsPanel = new MessagePanel(translate("Please select a settings category"));
```
plata: I think we should just select the first category by default.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.css
```diff
@@ -644,6 +644,11 @@
 /************************ icons ************************/
 /*******************************************************/
 
+/************************ util *************************/
+.refreshIcon {
+    -fx-graphic: url('/org/phoenicis/javafx/themes/breezeDark/icons/mainwindow/refresh.png');
```
plata: This should be removed (it doesn't work currently and does not belong to the topic of this PR at all).

plata:
Please open an issue for the refresh icon.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.css
madoar: Why doesn't it belong in the scope of this PR? It's a part of the refactoring. The refresh icon was used before and it's used afterwards. It's just that I've moved the icon graphic definition to the css files.

plata: Didn't see that you use it for the button. My mistake.

PR #728, opened 2017-04-25
fixed refresh repositories button icon

### LINKED ISSUE #726: Change the refresh graphic to a background definition
Currently we're using a the graphic parameter to set the refresh icon in the "Settings" tab when the "Refresh Repositories" button has been pressed to indicate that the repositories/scripts are currently refreshing. 
This approach has the disadvantage that the icon can't be resized, therefore we need to replace the graphic/`-fx-graphics` solution with a `-fx-background-image` one, which can be better adjusted.

The wanted design is that the refresh icon is shown in the left side of the button while next to it, in the right side, the "Refresh Repositories" text is shown.

```
-------------------------------
| Icon   Refresh Repositories |
-------------------------------
```

plata:
fixes #726


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
```diff
@@ -169,6 +169,7 @@ private void populateRepositoryRefresh() {
         this.refreshRepositoriesLabel.setWrapText(true);
 
         this.refreshRepositoriesButton = new Button("Refresh Repositories");
+        refreshRepositoriesButton.getStyleClass().add("buttonWithIcon");
```
madoar: I would prefer, if this line starts with a `this.` because the other lines are written like this.

In addition I think we should define when we want to use the keyword `this` and when not, to do this uniformly.

madoar:
Can you please provide a screenshot?
In addition it seems like Jenkins is currently on strike (maybe we should pay it better?^^)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
plata: Agree that we need a rule.

plata:
refresh icon

I think @qparis is the only one who has the rights to fix Jenkins.

qparis:
I don't understand, it seems that Jenkins works well isn't it?

plata:
@qparis it did not run the after my push yesterday.

madoar:
@plata it seems like a conflict occurred for this branch.

madoar:
As far as I'm concerned we can merge this, what do you think @qparis?

PR #730, opened 2017-04-26
Refactored and improving the LeftToggleGroup

madoar:
This PR does the following:
- refactored the LeftToggleGroup
- added an automatic removal from the ToggleGroup after a ToggleButton has been removed from the LeftToggleGroup
- added an AdhocList that combines an ObservableList with a number of normal objects to a new ObservableList
- improved the LeftToggleButton class to only fire an event after a click if the button isn't selected!

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/AdhocList.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/AdhocList.java
@@ -7,11 +7,24 @@ import javafx.collections.transformation.TransformationList;
 import java.util.stream.IntStream;
 
 /**
- * Created by marc on 26.04.17.
+ * This class combines an {@link ObservableList} and a number of non exchangeable objects called <code>others</code> to a single {@link ObservableList}.
+ * These <code>others</code> are prepended to the given {@link ObservableList}.
+ *
+ * @author marc
+ * @since 26.04.17
  */
 public class AdhocList<E> extends TransformationList<E, E> {
+    /**
+     * An array containing a number of objects that are prepended to the {@link ObservableList} <code>source</code>
+     */
     private E[] others;
 
+    /**
+     * Constructor
+     *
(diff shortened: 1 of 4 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/AdhocList.java
```diff
@@ -0,0 +1,88 @@
...
+import javafx.collections.ListChangeListener;
+import javafx.collections.ObservableList;
+import javafx.collections.transformation.TransformationList;
+
+import java.util.stream.IntStream;
+
+/**
+ * Created by marc on 26.04.17.
+ */
+public class AdhocList<E> extends TransformationList<E, E> {
```
plata: This class needs some Javadoc describing it's purpose and how it shall be used.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftToggleButton.java
```diff
@@ -34,6 +40,17 @@ public LeftToggleButton(String name) {
         this.setPadding(new Insets(2));
     }
 
+    @Override
+    public void fire() {
```
plata: What's the purpose of this? (comment in code please)

plata:
@qparis this looked OK for me. I wanted to have your opinion about the new `AdHoc` list as well. Could you please review this?

madoar:
@qparis, @plata, if there are any questions about the `AdhocList` just ask me.
The `AdhocList` is needed to combine a number of non changing (you could say static/final elements) elements with an `ObservableList` into a new `ObservableList`.
In our case this is needed to combine the single "all" button with a possible changing `ObservableList` containing the categories in the left sidebar.

plata:
I'm fine with that. Let's see if @qparis wants to add some remarks otherwise I will merge.

Issue #733, opened 2017-04-27
"Back" button


madoar:
Currently in the "Apps" and "Engines" tabs there is no defined way, how to return to the `MiniatureListWidget` from a miniature page.
For example, if you select an engine in the engines tab, you only see two buttons: "Install" and "Delete", but no button to return to the engines selection list.

Do you have an idea how we could/should handle the "move to the last window"/"go back" functionality?

plata:
I think there are mainly two options:
* a "Back" button in a bar at the top
* something like in Plasma Discover where a new column is opened to show the detailed information (see <https://userbase.kde.org/Discover>)

madoar:
I think option one would be the easiest to implement, but option two would look better. 
For option two we could replace the SearchBar with a "Back" button whenever a back button is needed.
We can do this, because a back button is normally needed, when a SearchBar isn't.
What do you think?

plata:
I guess we have to try it to see how usable it is. I hope that I find some time to implement something this weekend.

plata:
Quick try of the second option:
![new column](https://cloud.githubusercontent.com/assets/3973260/25555345/4970739c-2ce4-11e7-9be6-2c02f40091d4.png)

@qparis @madoar What do you think?

madoar:
Looks really nice. But I'm not sure how usable it is if you have a really small monitor or resolution.
In addition it may be better make a vertical divide.

I've just pushed a PR, that simply adds a "Back" button to the left side. The interesting part of my PR is, that it adds a navigation chronicle to the `MainWindowView` class that enables many layered back navigation, which we will eventually have if we use the engines install wizard to change the engine version of an existing container, see #625.

plata:
Maybe we can have a combination of both. I don't think a horizontal divide is a better idea (everyone is using wide screens). The right part for apps could be a lot smaller.

qparis:
I link it. Is the column always shown?

plata:
It occurs when you click on a miniature. The miniature should be highlighted and probably there should also be a close button for the right column.

madoar:
What would happen if we need a larger hierarchy of subwindows at a later time?
For example when we want to change the engine of a container:
> Container -> Engines List -> Selected Engine

plata:
I don't think we should go from Containers to the Engines tab. We should rather use a small embedded wizard.

madoar:
[...]

PR #735, opened 2017-04-29
Added a "Back" button to the apps and engines tabs

madoar:
This PR addresses #733. 
It adds a navigation chronicle to the `MainWindowView`, which is then used by the sidebar and the windows themselves to implement the back functionality.

grafik

grafik

grafik

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
@@ -134,7 +134,7 @@ public class ContainersController {
                             });
                 });
 
-                Platform.runLater(() -> viewContainers.showRightView(panel));
+                Platform.runLater(() -> viewContainers.clearChronicleNavigateTo(panel));
             });
         });
     }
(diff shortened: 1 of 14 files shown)


plata:
I'm not sure about replacing the search box. Maybe this confuses some users. How about a small bar with the button (right above "Audiosurf" in your screenshot)?

madoar:
I replaced the search box, because I think it's much more confusing to show it when you can't use it, which is the case, when you show an engine or an app. 
Therefore I thought it's better to replace it by a back button.

I could show a "Back" button or a bar above "Audiosurf", but I'm not sure how good this would look.
In addition do we have such a bar class?

plata:
We don't have such a class yet. I'm also not sure if it's really better. I just think we should compare several approaches before selecting the best one.

madoar:
Then we need to answer the question what should happen if you use the search bar when you've selected something?

plata:
Show the search result and close whatever has been open?

madoar:
We can do this, but again then again the question, is this intuitive?

plata:
No idea. At least e.g. Plasma Discover behaves like this.

qparis:
Hum nice work.
I agree that this button is quite confusing. I think we should stay on the navigation tree as possible. Maybe another option is to add a subitem in the Wine menu, corresponding to the version we are looking at. Then, the user will know that he has to click on "wine" to go back

plata:
@qparis have you tried #737? It doesn't use any back button but still feels quite natural in my opinion.

qparis:
So this PR is no longer needed right?

PR #737, opened 2017-04-29
new details column

plata:
The details column is closed if you select a different category or search.

apps

engines


madoar:
I've just opened PR #738, which continues the ideas of this PR with a `SplitPane`.
In addition it adds a close button to tabs after the first tab.

plata:
I can't really tell which solution is better as they are conceptually different: #737 is made to show the details of the currently selected item whereas #738 allows for a new navigation tree inside the right column.

Let's see which solution is preferred by @qparis

qparis:
The current approach is fine for me. Aside from that, I think it will be easier not to allow the end user to resize the frame between the two columns

madoar:
With current approach you mean this PR (#737) right?

About the resizing:
I think it's better to allow the user to resize the different columns. As long as the monitor used is large enough it isn't important if we allow the user to resize the columns, but if the resolution is quite low, it is important that the user can increase the size of a panel to see the complete content of it.
In addition does it have an disadvantages to allow the user to resize the columns? If he changes the column widths, we could "fix" the divider position when the user opens a new panel (like another app or engine)

qparis:
In my opinion, it is clearly not the responsibility of the end user to make the text readable. In any modern user interface (Material, iOS, ...), you won't see such possibilities.

madoar:
I don't mean to make the text readable by manually adjusting the column width. 
It's simply about having the ability to increase or decrease it to a level the user is comfortable with.
We can't adjust the software for any monitor resolution/size out there and some users may want to have the ability to make a column smaller or larger than the left one, which is the reason why I would allow them to do so.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -116,6 +111,7 @@ public void setOnSelectScript(Consumer<ScriptDTO> onSelectScript) {
      * Show available apps panel
      */
     public void showAvailableApps() {
+        this.showDetailsView(null);
         showRightView(availableApps);
```
madoar: The name of the method `showRightView` is kind of misleading in this case, because I would expect to set the details view in this case, because it is the most "right" view shown to the user.
I would suggest to rename it to something else.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -116,6 +111,7 @@ public void setOnSelectScript(Consumer<ScriptDTO> onSelectScript) {
      * Show available apps panel
      */
     public void showAvailableApps() {
+        this.showDetailsView(null);
```
madoar: For this I would suggest adding a `clearDetailsView` method, because I think it's unintuitive to clear the view by setting it to `null`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
```diff
@@ -141,7 +137,8 @@ public void setOnRetryButtonClicked(EventHandler<? super MouseEvent> event) {
     private void showAppDetails(ApplicationDTO application, SettingsManager settingsManager) {
         final AppPanel appPanel = new AppPanel(application, themeManager, settingsManager);
         appPanel.setOnScriptInstall(this::installScript);
-        showRightView(appPanel);
+        appPanel.setMaxWidth(400);
```
madoar: If we always want to limit the size of the right view, it may be a good idea to move this to the `showDetailsView` method.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/widget/MiniatureListWidget.java
```diff
@@ -61,7 +61,11 @@ private MiniatureListWidget(Pane content, Function<E, Element> converter, BiCons
         this.mappedElements = new MappedList<Element<E>, E>(items, value -> {
             Element newElement = converter.apply(value);
 
-            newElement.setOnMouseClicked(event -> setOnMouseClicked.accept(newElement, event));
+            newElement.setOnMouseClicked(event -> {
+                unselectAll();
+                setOnMouseClicked.accept(newElement, event);
+                select(newElement);
```
madoar: Just as a note:
Is it possible that the select css class is the same as the hover one?
If this is true we should choose a different style for the selection css class, to make it more clear to the user I think.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
plata: I just did it here because the app details were expanding too much. E.g. for the engines it was not necessary. It's not the smartest solution as it won't scale if you scale the UI but I think we can take care of that later on.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/widget/MiniatureListWidget.java
plata: It does the same (e.g. in Breeze Dark) but you could implement a different behavior if you wanted. Do you have an idea how it should behave such that it is more clear?

madoar: I think it would be enough if we use a different color to mark the hovered miniature.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
madoar: Have you tried this with the "Select, if you want to view the source repository of the scripts" setting in the User Interface Settings?
I'm not sure if everything is fully visible then (see my screenshots in #738)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindowView.java
```diff
@@ -74,16 +74,20 @@ protected void setSideBar(SideBar sideBar) {
...
         this.content.setCenter(nodeToShow);
     }
 
+    public void showDetailsView(Node nodeToShow) { this.content.setRight(nodeToShow); }
+
+    public void closeDetailsView() { this.content.setRight(null); }
+
     public void showWait() {
-        showRightView(waitPanel);
+        setCenter(waitPanel);
```
madoar: I think we should call `closeDetailsView` here too.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindowView.java
```diff
@@ -74,16 +74,20 @@ protected void setSideBar(SideBar sideBar) {
...
     }
 
     public void showFailure() {
-        showRightView(failurePanel);
+        setCenter(failurePanel);
```
madoar: I think we should call `closeDetailsView` here too.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindowView.java
```diff
@@ -74,16 +74,20 @@ protected void setSideBar(SideBar sideBar) {
         this.leftContent.setContent(sideBar);
     }
 
-    public void showRightView(Node nodeToShow) {
+    public void setCenter(Node nodeToShow) {
```
madoar: I think it would be quite nice if we use this opportunity to add some JavaDoc to this class.
This JavaDoc should describe, among other topics, how the view is structured.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/widget/MiniatureListWidget.java
plata: colors

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ViewApps.java
plata: Doesn't work indeed. I'm now using a tooltip for this.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindowView.java
plata: Agree in principle but I really don't want to do this in this PR (adds to many changes which are not actually related to this PR).

madoar: Ok, then let's do this in another PR at a later time

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindowView.java
```diff
@@ -74,16 +74,22 @@ protected void setSideBar(SideBar sideBar) {
...
         this.content.setCenter(nodeToShow);
     }
 
+    public void showDetailsView(Node nodeToShow) { this.content.setRight(nodeToShow); }
```
madoar: Something else we should think about at this point is how to show a close button for the details view.
In PR #738 I've used a `StackPane` to add a close button. Alternatively we can use a `BorderPane` for example, this would add the close button at the top and not overlapping to the details view.

I'm not sure if we should address this in this PR but I think we can already lay down the foundation for this by adding a BorderPane or a StackPane.

plata: I'm not even sure that we need a close button anymore. You have to try it out yourself but I didn't miss it.

madoar: It may be nice to enlarge the miniature list widget again, after you've clicked on a category.
On the current master this is only possible through a category change, because only this clears the details view, doesn't it?

plata: That's correct (or by searching etc.). We should keep it in mind.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindowView.java
madoar: What we should do here (in this method) too, is to call `closeDetailsView`. I think, that whenever this method is called the details view is automatically outdated.

madoar:
I think we can/should use this. 
Open topics for later PRs are:

- Adding JavaDoc/Documentation about out GUI makeup and the `MainWindowView` class in general
- Adding an close button to the details look, that enables the user to see the center view in its full size again

plata:
Could you create issues for the open topics? Otherwise we will forget them.

madoar:
I've just opened #758 and #759 for the two open topics.

PR #738, opened 2017-04-29
SplitPane column approach

madoar:
This PR uses a SplitPane and an ObservableList called navigationChronicle to show subwindows:
grafik
grafik
grafik
grafik


madoar:
I've just added close-Buttons to all tabs except the first:
grafik
grafik

madoar:
Here a screenshot, showing how it looks currently:
grafik

PR #751, opened 2017-05-01
scale the logo

plata:


qparis:
Can you send some screenshot please?

plata:
small:
small
big:
big

madoar:
It seems like it only resizes the text next to the icon?
I think the icon should be resized too, otherwise the proportions don't seem to fit.

qparis:
I agree with @madouar

plata:
I agree. However, I don't know why the logo does not scale.. The CSS says:
```CSS
-fx-background-size: 1.8em;
```

madoar:
I've just taken a look at google and found [this](http://news.kynosarges.org/2013/08/09/javafx-dpi-scaling/). In the given link the author explains some problems that occur with dpi scaling in javafx and how to work around them.
Maybe this helps us?

plata:
I've also seen that article before. I guess you're referring to

> Em sizes do appear to scale to the current default font size, but Oracle makes no guarantees to this effect.

What I don't understand is why it works for the left buttons. Shouldn't behave differently, should it?

madoar:
Are you sure it works for the left buttons? I've thought so too previously but when I look at them again I think the text is the only thing that scales, and the images only "look" larger but in reality have the same size.
Can you check, to be absolutely sure?

plata:
Just checked again. I think you're right.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.css
```diff
@@ -218,11 +218,15 @@
...
-#logoText {
+#logo {
+    -fx-background-image: url('/org/phoenicis/javafx/views/common/phoenicis.png');
+    -fx-background-size: 1.8em;
+    -fx-background-repeat: no-repeat;
+    -fx-background-position: left center;
+    -fx-label-padding: 0 0 0 2.2em;
+    -fx-vpos: center;
     -fx-font-family: "Maven Pro";
     -fx-font-size: 1.8em;
```
madoar: I think this line is the reason, the image is not scaling correctly.
If I understand #746 correctly, you use the `-fx-font-size` at the root level to define the required scaling.
This means that you're overriding the used scaling for `#logo`, which in my understanding leads to the logo being always the same size.

plata: That's not correct. The font size for the logo is defined relative to the root font size. (Nevertheless I've tried it. Didn't work.)

madoar: Ok, then I'm out of ideas.
Maybe it's because the image is shown inside a tab header?
It would be nice to see if the same java code and css would work, if the icon is added to another element via `setContent` or `getChildren().add`.

plata: It's strange. With `-fx-background-size: contain;` it scales. But then it's somehow impossible to get the label position right.

madoar:
Why not split the image and the label in two elements?
Then you can add `-fx-background-size: contain;` to the image css.

plata:
You're right. Why didn't I think about that myself...

plata:
scale logo

It seems to be impossible to specify the size of the `Region` in em in the CSS. As you can see in the screenshot, the scaling works. However, it behaves it bit strange if you change the scaling i.e. you have to hover over the logo until so that the padding is adjusted. Shouldn't matter normally.

madoar:
Can you force this by readding the logo component to the `TabPane`?

plata:
I don't know. Anyways I think it's not so important because it does not really inhibit the usage and you don't rescale all the time.

madoar:
How exactly does the padding need to be readjusted? Does it scale with the logo and its text?
Can we center it vertically in the space that's given?
In addition it seems to me like the css files of most themes are missing a ` #logoText` group.

plata:
It's hard to describe. You have to see it yourself. As I said, it's a bit strange but not really a problem.

madoar:
@plata can you fix the merge conflicts? Afterwards I think we can merge this.

Issue #686, opened 2017-03-30
Implement equals and hashCode methods


madoar:
To do a lot of things, we need to add `hashCode` and `equals` methods to all model- and DTO-classes.

plata:
@madoar I think this can be closed, right?

Issue #659, opened 2017-03-13
Log Repository information


madoar:
Sometimes, especially when writing logging code or debugging output, it's really useful to have a meaningful name for an `ApplicationSource`. Therefore I suggest that we add a getter-method declaration to `ApplicationsSource` called `getName()` that has the task to return the name of the application source.

plata:
What about putting this in `toString`?

madoar:
Hmm, under a `toString` method I imagine more than the simple printout of the name of the application source (which for a git application source is its url). In addition to the name, I would also print information such as the number of categories and scripts inside it, in the `toString` method.

plata:
You're right. Seems a bit strange to me to have a function which returns the name of the class for debugging, but if you think it's useful why not.

madoar:
Oh `getName` shouldn't (only) return the name of the class, because this wouldn't be enough to differentiate between two application sources of the same type, like two `GitApplicationsSource`s.
For git application sources I would return the repository url, and for local application sources I would return the path to the folder, with `getName`.

plata:
This sounds almost like `toString` then^^

plata:
@madoar what shall we do about this? Is it still valid?

madoar:
I'm not sure, I think we need this for debugging purposes, but because we've always had more important issues, I couldn't address this yet.

plata:
Ok. How about renaming this issue to: "Log used repositories" or something like that?

plata:
@madoar we have a log output for the used repositories by now. Apart from this, it seems like nobody needed more functionality here in the last months. Therefore I would suggest to close this.

madoar:
It seems like you're right.
The point of this issue was to do some fundamental work in regards to making the merge process of multiple repositories better debugable, i.e. which repository was merged in which order.
Every time I look at the merging code, I think it contains some kind of error...

plata:
Ok. I'll close it for now. If we ever encounter a problem there, we can still improve the logs.

PR #756, opened 2017-05-05
Cleanup of the phoenicis-apps project

madoar:
This PR partially addresses #686 and #659.
In addition it contains the following:
- moved the repository classes to a separate `repository` package inside the phoenicis-apps project
- added `toString` methods for the most often used repository types
- added equals and `hashCode` methods for the NullRepository
- removed the filter classes from the phoenicis-apps project. In my opinion detailed filters should be implemented in the projects, that make use of the repositories. The only exception to this is the operation system filter, which is still managed by the `FilterRepository`.


plata:
Why is the `MultipleRepository` deleted and added? Didn't you use `git mv` for this class (all others seem all right)?

madoar:
I didn't use `git mv`, I moved them in `IntelliJ` and added the changes afterwards with `git add .`, not sure how git perceives this.

plata:
I think it doesn't realize it. Just wondering why it worked for the other classes.

madoar:
Ah, I know why it was interpreted as deleted. Take a look at the indention. It's either done by tabs or using 8 spaces, while the other files are indented with 4 spaces.

plata:
We should have a guideline about the required indention somewhere. I also noticed that my IntelliJ started complaining about 4/8 spaces some time ago.

plata:
Seems like 8 is correct:
```xml
<setting id="org.eclipse.jdt.core.formatter.indentation.size" value="8"/>
```

madoar:
I know that the "official" code format is 8, but I think 8 is too much.
As far as I know we currently find both an indention of 8 and of 4 in the code, do you know what is more often used?
I myself would suggest to go with an indention of 4.

plata:
I don't know which is used more often. I'm just surprised why this is not changed automatically by the `maven-java-formatter-plugin`.

madoar:
In addition it seems like the maven formatter plugin ignores the indention too, see #765 for the corresponding issue.

plata:
I'm not so sure about the `repository` folder. E.g. why is `DefaultRepositoryManager.java` in `apps` and not in `repository`, why are the tests not moved to a `repository` folder?
Moreover, I think it would be better to rename `apps` to `repository` directly. Actually, the categories contain not only apps but also functions/verbs. Have a look at #622 for more background.

madoar:
I've moved the `Repository` implementations in a separate package, to make them easier to see and find.
The `RepositoryManager` isn't a repository itself (as can be seen by the fact that it doesn't implement `Repository`) and is therefore not contained in the new package.

plata:
I see. Nevertheless, I think that the `apps` folder is a bit misleading and should be named `repository` instead. Then we would have `repository/repository`.

plata:
Why didn't the formatter apply automatically before? It's supposed to run whenever you build.

madoar:
I tested it in `IntelliJ` and only run `mvn clean test`. So I guess that intellij doesn't automatically run the plugin?

plata:
I think the default build step for the plugin is validate. So it's not working as I would expect.

madoar:
I would suggest to solve the maven formatter plugin problems in another PR and view this one without the formatting problematics.

plata:
Yes.

plata:
Approved but we should clarify the naming of the module (app vs repository).

madoar:
You're right app(s) is maybe not the right term to use here.
I'm open for suggestions.

plata:
We can discuss that later in #770.

Issue #757, opened 2017-05-05
Problem with resizing images


madoar:
When opening POL 5, with the default scaling (the slider in the settings is in the middle), POL 5 will look like this:
![grafik](https://cloud.githubusercontent.com/assets/18488086/25745220/21d8d19e-319e-11e7-835c-5a64ac09e879.png)
We need to make sure that the images are resized too, and that the text below the images are fully visible.

If you change the slider to the left, the text below the images will disappear.

plata:
Agree to scaling the pictures. However, the broken text is not related to that. I will update the themes, that should fix it.

plata:
Regarding the images: Do you have an idea how to do this? I've tried that in #746 already but it's not so straight forward.

madoar:
Sorry I have no clue too. 
As far as I understand it the font size should apply to images too, but in some sporadic cases it doesn't work.
If we can't solve this we should think about using another approach to scaling/resizing the themes.

plata:
The problem here is that the image is created from a byte[] inside the code. It should work if we would set the path instead and define the size in the CSS.

madoar:
So we need to create a css file on the fly, for the images?
It would be nice if we could do this semi persistent (i.e. not in a temp folder), what do you think?

plata:
Don't think that's needed. We can directly set the CSS code for the image in the code.
```Java
xy.setStyle("-fx-background-image: url('...');");
```

madoar:
You're right this is more elegant and doesn't need us to create a temp file at all.

madoar:
Should we create a new issue for the extraction of the image data to the style properties, or should we do that in the scope of this Issue too?

plata:
I guess the question is obsolete after #773, right?

madoar:
Yes, it should be.

PR #760, opened 2017-05-05
scale default theme

plata:
fixes cut off text in #757

default theme


madoar:
Does this problem only occur in the default theme or in other themes too? (I didn't check)

plata:
This special problem is only in the default theme (it set the max. size of the miniatures...). I will nevertheless adjust all themes (different pull requests).

PR #764, opened 2017-05-05
removed obsolete HiDPI theme

plata:
it was still shown as option in the settings


madoar:
What about the css files belonging to the HIDPI theme?
Should we remove them at this point too?

plata:
The are removed already. I've just forgotten the settings when I did that.

madoar:
Then I think we can merge this, but I'm not sure if I should decide this (in general)?

plata:
No idea. I always try to do it this way: If the change does not have a big impact, I decide. If it affects the structure of the software, wait for @qparis OK.

madoar:
Then let's go with this approach. I think this has a small impact^^

madoar:
Ok not sure how the merging works, so that you're marked as the author....

plata:
You should squash and merge and fill the title and description properly.

PR #767, opened 2017-05-06
use formatter-maven-plugin

### LINKED ISSUE #765: Maven formatter plugin seems to ignore indention
After executing `mvn java-formatter:format`, it seems like the indention is ignored.

plata:
fixes #765


madoar:
I think before we merge this we should decide, if we want to change anything in our formatter configuration.

madoar:
In addition, are you sure that it works without the `file://` before the `configFile` string? 
When I previously tried locally I sometimes go URLMalformedExceptions, because of a missing protocol.

plata:
At least I get 183 files changed after compiling, so I think yes. I don't think it's necessary to change the config right now. The first think we have to do after merging this is creating a PR with the formatter changes.

madoar:
I'm not really happy with the 8 indentions :/
When we disregard that, this PR looks good to me.

plata:
I agree that 8 is too much but we can decide on that later. Let's merge this now so we can also merge the formatted files.

madoar:
This seems to be a small change to me. If you want you can merge this.

PR #773, opened 2017-05-06
Replaced byte[] with icon and miniature data in ApplicationDTO and ShortcutDTO with URI objects

madoar:
This PR replaces the byte arrays, that contain the image data corresponding to shortcuts and applications with `URI` objects.
The result now looks like this:
grafik
grafik
grafik
grafik

I know that the proportions are a bit off now. It would be nice if someone can correct them at a later time.

A follow up PR should concentrate on the problem with the ShortcutDTO content.
When a `ShortcutDTO` is created, it receives the `URI` objects from the corresponding `ApplicationDTO` object. Only after a restart at a later time the URI will be corrected to link to a copy of the icon/miniature that belongs to the shortcut.

### DIFF
--- a/phoenicis-apps/src/main/java/org/phoenicis/apps/dto/ApplicationDTO.java
+++ b/phoenicis-apps/src/main/java/org/phoenicis/apps/dto/ApplicationDTO.java
@@ -21,6 +21,7 @@ package org.phoenicis.apps.dto;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 
+import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -33,8 +34,8 @@ import java.util.List;
 public class ApplicationDTO {
     private final String name;
     private final String description;
-    private final byte[] icon;
-    private final List<byte[]> miniatures;
+    private final URI icon;
+    private final List<URI> miniatures;
     private final List<ScriptDTO> scripts;
     private final List<ResourceDTO> resources;
 
@@ -51,11 +52,11 @@ public class ApplicationDTO {
         return resources;
     }
 
-    public byte[] getIcon() {
+    public URI getIcon() {
         return icon;
     }
 
-    public List<byte[]> getMiniatures() {
+    public List<URI> getMiniatures() {
         return miniatures;
     }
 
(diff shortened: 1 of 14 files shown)


madoar:
After replacing cover with contain in `-fx-background-size` POL 5 looks like this:
grafik
grafik
grafik
grafik

plata:
I like it. Why do we need `appMiniature` and `miniatureImage`?

madoar:
To set the sizes of the images. If you don't have them the images are either to large or too small.
If you can get it working without them that's also fine :)

plata:
My question rather was why we need both of them. Is there anything special about the miniatures for apps?

madoar:
`appMiniature` is for the miniature inside the `AppPanel`, which you see after you've clicked on an application in the Apps tab.
`miniatureImage` is for the miniature inside the `MiniatureListWidget`.

plata:
Ah ok. Don't forget to update the wiki with that information.

madoar:
Do you want to merge this now, or do you want to wait for @qparis?

plata:
I would like to wait for @qparis (I don't know if he had any special plans with the byte[]).

madoar:
Another needed follow up PR for this one is to clean up the css classes for the miniature list widget.
Some classes are missing some important definitions, like `.miniatureText` which is missing `-fx-pref-width` and `-fx-pref-height`. 
Other classes contradict each other, like `.miniatureListElement` and `.miniatureImage` with their different `-fx-background-radius`, `-fx-pref-width` and `-fx-pref-height` attributes.

qparis:
This has already been discussed before. 
A DTO should not contain an URI, because this is not properly serializable. The ObjectMapper won't work anymore and we won't be able to use them to create webservices for instance

plata:
The CategoryDTO stores the icon path in a String. How's that supposed to work then?

qparis:
A string is fine, a URI is not serializable (afaik)

plata:
So we can just replace the URI by a String?

plata:
According to the [docs](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html) URI implements Serializable. That's not sufficient?

qparis:
Not sure that jackson will accept it

madoar:
I guess it can (see https://github.com/FasterXML/jackson-databind/issues/908) but I'm not sure how well.

madoar:
Even if we don't use URIs in the DTOs, we'll still need to use them somewhere, because we need them to scale the images in the miniature list widget and the AppPanel. 
I'm not sure how to do this elegantly otherwise...

madoar:
@qparis, is there an easy way to test if jackson can do this?

plata:
@madoar if we store the URI as String, everything should work fine (just like for CategoryDTO).

madoar:
If possible I would like to not convert the URI to a String internally, because it needs an additionally possibly unnecessary step in between and because we're losing a part of the type safety when doing that.

But you're right, we should have this option in mind, if jackson really can't serialize URI objects.

qparis:
It seems that jackson accepts URI objects

Issue #774, opened 2017-05-06
Repositories Settings has small unresizable box


Ads20000:
Why is it like this? This should automatically expand out to the right at I expand a window and automatically contract as I make it smaller. Like Ubuntu's Software & Updates app, for example.

![screenshot from 2017-05-06 19-49-26](https://cloud.githubusercontent.com/assets/3178485/25775135/15b9a430-3296-11e7-8552-d7cca5fa7717.png)
![screenshot from 2017-05-06 19-55-11](https://cloud.githubusercontent.com/assets/3178485/25775136/1c55a3de-3296-11e7-9f8f-b4786d7e8b01.png)
![screenshot from 2017-05-06 19-53-52](https://cloud.githubusercontent.com/assets/3178485/25775138/23f3a67c-3296-11e7-87b7-62dd1beae3ce.png)
![screenshot from 2017-05-06 19-54-11](https://cloud.githubusercontent.com/assets/3178485/25775139/28f1cf1e-3296-11e7-8888-1aa86d3e4372.png)

plata:
fixed by #775

PR #775, opened 2017-05-06
Changed Repository List in the Settings to scale with the window

madoar:
This PR addresses #774.
In addition this PR makes a bug in the css for the refresh button visible. I guess it's because something with the padding is wrong.
grafik
grafik
How do we want to handle a height increase of the window? Should the height of the list increased too in this situation?

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
@@ -95,7 +95,6 @@ public class RepositoriesPanel extends VBox {
         this.repositoryLayout.setSpacing(5);
 
         this.repositoryListView = new ListView<>(repositories);
-        this.repositoryListView.setPrefSize(400, 100);
         this.repositoryListView.getSelectionModel().setSelectionMode(SelectionMode.MULTIPLE);
         this.repositoryListView.setEditable(true);
         this.repositoryListView.setCellFactory(param -> new DragableRepositoryListCell((repositoryUrl, toIndex) -> {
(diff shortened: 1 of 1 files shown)


plata:
Can we adjust the height such that it fits the content?

madoar:
grafik
grafik
@plata can you take a look at the refresh button css?

plata:
I think you got me wrong. My idea was to fit the height of the list to the list content (such that it's large with 20 entries but small with only 2).
We should merge #760 - #763 first. Maybe this fixes the button already.

madoar:
I'm not a friend of making it larger and/or smaller when an element is added or removed, I think this would disturb the user.

plata:
I see. But like this it's also strange. Consuming the whole page for nothing.

madoar:
We could limit the repository list to a percentage of its parent panel, but I think that wouldn't make much sense. 
I guess this will look more natural, when more settings options are added for repositories.

plata:
Maybe you're right. Do you want to change something or should we merge?

madoar:
I've just discovered, that this introduces a new bug.
Take a look at the screenshot of the smaller window of POL 5.
You'll discover that the priority hint is shortened with "...". The problem is, that this shouldn't be the case because we've used `this.priorityHint.setWrapText(true);`. 
I'm not sure how to make this work correctly when we've an element that should always fill all remaining space.

plata:
What happens if you set the `vgrow` for that label?

madoar:
It has no effect.

plata:
Does it work if you use `Priority.SOMETIMES` for both?

madoar:
Using `Priority.SOMETIMES` has an effect if you manually increase the height of the window.
In you don't change the height, like in the first screenshot, it has no effect.

madoar:
I've just tried to remove all `setVgrow` and `setHgrow` calls from `RepositoriesPanel` and swapped the positions of `priorityHint` and `repositoryGrid`. This had no effect too.

My guess is that either a css class used inside POL 5 prevents the `setWrapText(true)` call from having an effect after startup, or the `ListView`, or maybe `GridPane`, takes precedence over the Label components.

plata:
Does this article help? <http://news.kynosarges.org/2013/11/20/javafx-listview-sizing/>

madoar:
@plata thanks, the article helped:
grafik

The problem seems to be, that the default prefHeight is wrong (too large?). When I set it to 0 (see my last commit) it works.

madoar:
Just as a note, the button seems to be shown correctly now too, if you exclude that the border is missing.

PR #778, opened 2017-05-12
use latest release of dependencies

### LINKED ISSUE #768: Update maven dependencies
Currently we're using some outdated maven dependencies (see #767 for example).
Therefore it would be good to go over each of them and update them to a more current or up to date version.
To see for which dependency a newer version exists we can use `mvn versions:display-dependency-updates`.

plata:
via `mvn versions:use-latest-releases`

fixes #768

### DIFF
--- a/phoenicis-apps/pom.xml
+++ b/phoenicis-apps/pom.xml
@@ -77,7 +77,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-all</artifactId>
-            <version>1.10.19</version>
+            <version>2.0.2-beta</version>
             <scope>test</scope>
         </dependency>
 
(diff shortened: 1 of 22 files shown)


phoenicis-apps/pom.xml
```diff
@@ -77,7 +77,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-all</artifactId>
-            <version>1.10.19</version>
+            <version>2.0.2-beta</version>
```
madoar: We shouldn't use beta versions

phoenicis-configuration/pom.xml
```diff
@@ -51,7 +51,7 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>18.0</version>
+            <version>22.0-rc1-android</version>
```
madoar: Release Candidates are also bad. In addition this seems to be an android version?

plata:
So this maven versions plugin doesn't seem to work properly.

phoenicis-apps/pom.xml
```diff
@@ -102,13 +102,13 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.7.0</version>
+            <version>2.9.0.pr3</version>
```
qparis: This version has some conflict with some spring versions.
Honestly, I don't like the idea of not controlling which version of which dependency we are using. 
This could break the code for no reasons

PR #779, opened 2017-05-12
use latest release of dependencies

### LINKED ISSUE #768: Update maven dependencies

plata:
via `mvn versions:use-latest-releases`, now with a rules.xml to exclude beta, alpha etc.

fixes #768

### DIFF
--- a/phoenicis-apps/pom.xml
+++ b/phoenicis-apps/pom.xml
@@ -28,6 +28,10 @@
 
     <artifactId>phoenicis-apps</artifactId>
 
+    <properties>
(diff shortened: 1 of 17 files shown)


plata:
@qparis it's not like this changes anything automatically. You still have to run `mvn versions:use-latest-releases` to do the update. Then you can decide which changes you want and which not before you commit.

Issue #736, opened 2017-04-29
miniature list: select miniatures or simple list


plata:
Allow to switch between a miniature list and a simple list (only Strings without icons).

madoar:
How shall the user select, which view he wants to have?

plata:
Maybe as global option in settings → UI or buttons like in a file browser
![buttons](https://cloud.githubusercontent.com/assets/3973260/25563095/9c859f70-2d94-11e7-8ba4-70dac7c34927.png)

madoar:
I think the second one looks approach in your screenshot looks better.

plata:
That's only GIMP. What concerns me a bit is that the icons are not directly related to the miniature view.

madoar:
The problem when using a global option in the settings for this is, that I don't think a lot of users will find this option or understand what exactly it does by looking at the description.

plata:
That's probably true.

madoar:
@plata if you have nothing against it, I would like to have a go at this. 
What I would like to know before I start is, which buttons in the left bottom of the sidebar do we want to use? It seems like the Faenza icon set has 2 icons we could use, at least for the default design.
But I must say that I like the icons from your screenshot better. If you agree with them too can you share them here?

plata:
Feel free to assign yourself. The icons in my screenshot are from Breeze Dark (actually from a screenshot I made of Dolphin). I'm not sure we should use them in the default theme. The style should match Faenza, so I would try those icons first.

plata:
Should be:
* view-list-compact-symbolic.png
* view-list-details-symbolic.png
* view-list-icons-symbolic.png

qparis:
Honestly I find the miniature list cleaner

qparis:
I think you could take advantage of the list to display extra information

qparis:
However, we should let the user chose want view he wants

madoar:
@qparis what do you mean with miniature list? My "new" one or the old one?

This is only my first approach, where I'm testing how to set the width/height etc. properties.
Therefore you can't switch the view atm.
What additional information do you want to have shown?

qparis:
The name of the container, or the engine could be displayed for instance

madoar:
As far as I know these information are not available for the `ApplicationDTO` objects in the "Apps" tab?
I'll keep it in mind, when I add the new list to all tabs

qparis:
This is correct. The idea is just ton leave less empty spaces.

[...]

PR #786, opened 2017-05-13
structure containers in categories

### LINKED ISSUE #739: categories for container tab
Currently, all containers are shown in the side bar of the container tab. In my opinion, the sidebar should show the container category instead (e.g. Wine) for two reasons:
1. If you have many containers, showing them in the side bar (even within a scroll pane) is not really usable.
2. With categories, the container tab can be streamlined with the apps and engines tabs which makes the UI more understandable for the user.

Therefore I propose the following solution:
* Use the sidebar to show the container category: All, Wine, ScummVM etc. (To represent this structure, also the folders should be organized in this way, i.e. move `wineprefix` into a new `containers` folder.)
* Show all containers of the selected category in a miniature view in the center (exactly like for apps).
* Show the container information in the right column (see: #733)

plata:
fixes #739 

container categories


madoar:
I'm not sure if I like the amount of space this needs.
It looks like this may be quite challenging to view on smaller monitors.

phoenicis-containers/src/main/java/org/phoenicis/containers/dto/ContainerCategoryDTO.java
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Represents a category of container
+ */
+@JsonDeserialize(builder = ContainerCategoryDTO.Builder.class)
+public class ContainerCategoryDTO {
+    private final String name;
+    private final String description;
+    private final List<ContainerDTO> containers;
+    private String icon;
```
madoar: `URI`?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ViewContainers.java
```diff
@@ -18,62 +18,99 @@
...
 
         this.sideBar.setOnApplyFilter(this::applyFilter);
 
-        this.sideBar.bindContainers(this.containers);
+        this.sideBar.bindCategories(this.sortedCategories);
+        Bindings.bindContent(this.availableContainers.getItems(), this.sortedContainers);
 
-        this.initSelectContainerPane();
+        // set the category selection consumers
+        this.sideBar.setOnCategorySelection(category -> {
```
madoar: We don't really need the `{` and `}` here.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ViewContainers.java
```diff
@@ -18,62 +18,99 @@
...
+            showAvailableContainers();
+        });
+        this.sideBar.setOnAllCategorySelection(() -> {
```
madoar: We don't really need the `{` and `}` here.

phoenicis-containers/src/main/java/org/phoenicis/containers/dto/ContainerCategoryDTO.java
plata: Didn't merge master yet.

plata:
I agree that it takes some space. However, there are several reasons why I believe that it still makes sense:

- similar UI as for apps/engines → more consistent user experience
- enables many containers from different engines (imagine you have Wine, ScummVM, ResidualVM with each ~10 apps and all containers shown in the sidebar)
- with #736 the miniature list will consume a lot less space
- there's no need to have 4 Wine tool buttons in a row; with a floating approach this can take more space vertically

phoenicis-containers/src/main/java/org/phoenicis/containers/dto/ContainerCategoryDTO.java
```diff
@@ -0,0 +1,114 @@
...
+    public List<ContainerDTO> getContainers() {
+        return containers;
+    }
+
+    public URI getIcon() {
+        return icon;
+    }
+
+    public static Comparator<ContainerDTO> nameComparator() {
+        return (o1, o2) -> o1.getName().compareToIgnoreCase(o2.getName());
```
madoar: We could do this by using `Comparator.comparing(container -> container.getName().toLowerCase())`, right?
I'm not sure which approach is more elegant.

plata: Don't know. I just copy pasted and renamed `EngineCategoryDTO` ;) Didn't think about changing the code.

madoar:
Looks good to me (especially if we change the container panel to better fit inside the details view), but I think it's better to wait for @qparis to see if this fits to his concept.

plata:
Agree to wait for @qparis.

qparis:
Totally agree with the concept. Good job!

PR #787, opened 2017-05-13
use URI for EngineCategoryDTO icon

plata:


madoar:
Does this require some changes to the `Scripts` repository?
I'm kind of missing the code that creates an `EngineCategoryDTO` object.

plata:
I don't think so. It is created in `EnginesSource`: 
```Java
EngineCategoryDTO wine = new EngineCategoryDTO.Builder()
```

madoar:
Which means that we're currently setting no icon for the engines?

plata:
Yes. It's using the default `WINE_MINIATURE` in `StaticMiniature`.

plata:
see #792

PR #789, opened 2017-05-13
show shortcut details in library

plata:
library details

This could also be used to show hints (e.g. for Prince of Persia The Sands of Time: "do not enable fog in the graphic settings").
Maybe we could also change the shortcut from here (e.g. debug level).

PR #791, opened 2017-05-13
implemented shortcut categories

### LINKED ISSUE #790: shortcut categories
It would be helpful to categorize the shortcuts (same categories as for apps). Then we could group them by category in the "Library" tab. If you have many shortcuts in your library, this would make navigating it a lot easier. The correct category can be determined during the installation of an app (simply use app category also for the shortcut).

plata:
fixes #790 

requires PlayOnLinux/Scripts#235

all

games

default


madoar:
Can you please give us a screenshot with the `run`, `close` and `uninstall` buttons?

phoenicis-containers/src/main/java/org/phoenicis/containers/wine/WinePrefixContainerController.java
```diff
@@ -133,18 +134,20 @@ public void deletePrefix(WinePrefixContainerDTO winePrefix, Consumer<Exception>
...
+                        interactiveScriptSession.eval("include([\"Functions\", \"Shortcuts\", \"Reader\"]);",
+                                ignored -> interactiveScriptSession.eval("new ShortcutReader()", output -> {
+                                        final ScriptObjectMirror shortcutReader = (ScriptObjectMirror) output;
+                                        shortcutReader.callMember("of", shortcutDTO);
+                                        final String container = (String) shortcutReader.callMember("container");
+                                        if (container.equals(winePrefix.getName())) {
+                                                shortcutManager.deleteShortcut(shortcutDTO);
+                                        }
+                                }, errorCallback), errorCallback);
+                }
```
madoar: I believe the Streaming API would be much nicer here:
```java
categories.stream()
   .flatMap(category -> category.getShortcuts().stream())
   .forEach(shortcut -> { ... });
```

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
```diff
@@ -105,11 +109,16 @@ public void deleteShortcut(ShortcutDTO shortcutDTO) {
         final String baseName = shortcutDTO.getName();
         final File shortcutDirectory = new File(this.shortcutDirectory);
 
+        final File categoryFile = new File(shortcutDirectory, baseName + ".category");
```
madoar: I think we need a definition in the wiki about what these files (`.category`, `.shortcut`, `.icon`, `.miniature`, `.description`) contain.

madoar:
In addition I think this PR should also show some icons for the new shortcut categories (e.g. Games and default)

plata:
icons

@madoar I would like to remove the default category but couldn't figure out how to do it. `LeftToggleGroup` doesn't like if the `converter` does not return a `ToggleButton`. Can we somehow enhance this such that the `converter` returns `Optional<ToggleButton>`?

plata:
run, close and uninstall:
grafik

However, I would suggest to move those buttons to the details panel on the right (see #789).

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
plata: https://github.com/PlayOnLinux/POL-POM-5/wiki/Shortcuts

madoar:
About the location of the buttons:
You mean the Run, Close and Uninstall buttons right?
I guess we can move them to the right details view.

About `Optional<ToggleButton>`:
We can add this, but I'm not sure if we should add this in the same class, because not all tabs need this, right?

About the implementation details of the `Optional<ToggleButton>` approach:
First we would need to change the supplier function to produce a `Optional<ToggleButton>` object.
Second we would need to add a `FilteredList` that retains all elements from `mappedToggleButtons` which are not empty
Third we need a `MappedList` that maps all elements in the `FilteredList` to their content.
This `MappedList` should then be passed to the constructor of the `AdhocList`.

plata:
Button locations: Yes, I've been talking about the run, close and uninstall button.

Do you see another way to avoid the "default" category? I also tried to use "All" instead of "default" but then you get two "All" buttons...

madoar:
What I'm unsure of is where the default comes from?
The problem with the "All" category is, that it shows the content of all known categories and is not a category itself.
In comparison as far as I understand, the default category is a category itself, that is used as a basin for all not clearly assigned applications.
I guess if you don't want to show this category explicitly it's really the best choice to ignore it using an `Optional<ToggleButton>` supplier.

plata:
I'm currently setting default as a fallback if nothing else is specified. I understand that the "All" button is not really a category, the idea behind that approach was to somehow hide the "default" in "All".
Maybe we should leave it like it is for now. I don't really want to implement the change for `Optional<ToggleButton>` in this PR and the "default" category should usually not show up in real life.

madoar:
I believe we can go with this. 
@qparis what do you think?

qparis:
What about fallbacking in the 'Other' category? It already exists and has a suitable icon

plata:
other

plata:
@qparis also good for you like this?

qparis:
Yep :) Thanks

PR #795, opened 2017-05-15
List widget selection

madoar:
This PR targets #736. It allows the user to choose between three different list representations.
This PR only adds the ability to choose between the different list representations to the "Apps" tab, because to allow this for all tabs it would be a good idea to change the super class of `LeftSideBar` from `VBox` to `BorderPane`. This enables us to add the button bar, that enables the user to switch between the list representations easily to the bottom of each sidebar.
grafik
grafik
grafik
grafik
grafik
grafik
grafik
grafik
grafik


plata:
I don't think it makes sense to show the description in the list. In my opinion, you select the list if you want to save space (horizontally).
Also, the buttons somehow look out of place (e.g. why do they have this blue border?).

As for this PR in general, however, I like it a lot.

madoar:
I can remove the description again, that's no problem. I've just added it to show some other content to make the list feel not this empty.

The blue border signals, which button is currently active. I choose blue because it was the first thing that came to mind. If you have a better idea feel free to tell me.

plata:
For Breeze Dark I would like something like:
grafik

Maybe something similar is possible for default?

madoar:
I've tried to use a different background color instead of a border color for the selection state but this isn't as visible for the default theme as for the breeze dark theme. The cause for this is, that the default icons have a lot of background already set with a low transparency.
Here a screenshot showing how a blue **background** color looks like: 
grafik
The "border" you see exists because the icon has a fully transparent band around the main part of the icon, which is fully blue after setting the background color.

madoar:
This is how it looks like now (almost the same as before):
grafik
grafik
grafik
grafik
grafik
grafik
grafik
grafik
grafik

plata:
For the buttons: We could use a lighter background for not select and a darker one for selected.
buttons

What do you think?

madoar:
I would suggest to leave it like it is now and change it in another PR.
The reason: We need to change the icons and the focus color for the other themes too and this should be done in individual PRs too. What do you think?

plata:
That's fine for me. Do you think we should wait for @qparis ok?

madoar:
I'm not sure...

qparis:
It looks ok for me :-)

madoar:
I would like if this is merged fast, because it's a big change which may interfere with other PRs. So the faster we do this the less work we'll have with other PRs.

madoar:
@plata want to approve and merge it?

plata:
@qparis can you approve the review then?

PR #798, opened 2017-05-20
updated Breeze Dark theme for list type selection

plata:
breeze dark


madoar:
Do you want to use a background change, to make the selected list widget visible, for this theme too?

plata:
I tried but it didn't work (you can hardly see that the background color even changed). Therefore I went for the border.

PR #800, opened 2017-05-20
updated default theme for list type selection

plata:
default


madoar:
images pls? :)

plata:
@madoar nothing changed. I had renamed the icons so I had to fix that in the other themes as well (forgot it before).

PR #801, opened 2017-05-20
moved shortcut buttons to details panel

plata:
shortcut buttons


madoar:
Maybe we should integrate the buttons in the layout of the details view:
```
Run            Run button
Close         Close button
Uninstall     Uninstall button
```
What do you think?

plata:
Like this (I don't like it)?
layout

madoar:
No the other way around, the icons should be on the right side

plata:
Doesn't look better in my opinion:
icon right

madoar:
and if you do a right alignment for the labels?

plata:
align right

madoar:
@qparis which look do you think looks best?^^

madoar:
@qparis can you please take a look at this too?

qparis:
That last propositions do not look very natural for me. What about trying to make the icons bigger and to put the text under each one?

plata:
Like this?

big icons

plata:
I still prefer the first approach.

madoar:
@plata what about this:
7d40fcee-4479-11e7-984f-ccc52b8170a9

plata:
Better. But still a bit too much "in your face" in my opinion. Maybe we should just merge this until we have a better idea (the functionality is working).

qparis:
Just reduce the size of the icons a little bit

plata:
Like this?
smaller buttons

qparis:
Yes, except the fact that the texts are not aligned

plata:
That's just a quick GIMP. I'll move this to #852 to fix it there.

PR #805, opened 2017-05-20
fixed padding of side bar buttons in Unity theme

plata:
padding

### DIFF
--- a/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/unity/main.less
+++ b/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/unity/main.less
@@ -389,7 +389,7 @@
         -fx-label-padding: 0.2em 0 0.2em 2.2em;;
         -fx-background-size: 2em;
         -fx-background-repeat: no-repeat;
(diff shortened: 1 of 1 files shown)


madoar:
I'm not sure, but shouldn't the `SearchBox` have a padding too?

plata:
No. It's on purpose that it takes the complete width of the side bar.

madoar:
Ok. After you've fixed the conflict I'll approve this PR.

madoar:
Or @qparis does it :P

madoar:
One more idea: what about an orange border for the list widget selection?

plata:
I will try it.

plata:
Doesn't look so good. The problem is that the border should have some radius but then it doesn't surround the icon well (because the icon has some shadow).

madoar:
What about using `-fx-background-color` and `-fx-background-radius` as an alternative for `-fx-border-color` and `-fx-border-radius`?

plata:
Also tried that. But than the icon gets a strange color as it is opaque.

madoar:
Ok, then let's stay with the current approach, it was worth a try ;)

PR #811, opened 2017-05-21
fixed font size in description.less

plata:
The font size was either not scaled or too large (Breeze Dark).

font size


plata:
I think I did that already once. But feel free to check again.

PR #813, opened 2017-05-21
Added a new DetailsView class with close button

### LINKED ISSUE #759: Add a "close" button to the details view
Currently it's only possible to close the details view through a category change or a searchterm entry.
In addition to these methods we should also provide a "close" button for the details view, that enables the user to close it and maybe also do some cleanup work, like clearing the selection in the miniature list widget.

A basic close button css class for the default layout can be found in #738.
For more details see #737 and #738.

madoar:
Fixes #759 
grafik
grafik
grafik
grafik
grafik
grafik

You'll see that there are some quirks in the `Container` and `Engines` tab details views, which need to be fixed in later PRs. In addition I would **really** suggest to refactor the `DetailsView` subclasses.


plata:
With the title, the button looks really out of place because it does not seem to be aligned in any way.

madoar:
The title is aligned to the top left while the button is aligned to the top right.
I'm open for suggestions :)

plata:
Can you try to align to center (vertically)?

madoar:
In addition there are two tabs that set no title. These are the container and engines tabs.

plata:
I just think that aligning both button and text to top instead of center looks strange because they do not have the same height.

madoar:
I would suggest to merge this first as is and later do some refactoring in the details views. If during this we decide that it's needed to change the alignment we should do it then. What do you think?

plata:
I don't know. Why not fix it now? Should not be a big deal to change `Pos.TOP_LEFT` to `Pos.CENTER_LEFT`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindowView.java
```diff
@@ -71,7 +74,7 @@ public void setCenter(Node nodeToShow) {
         this.content.setCenter(nodeToShow);
     }
 
-    public void showDetailsView(Node nodeToShow) {
+    public void showDetailsView(DetailsView nodeToShow) {
```
plata: I don't think we should change this (or do you use any special functionality of `DetailsView` here?)

madoar:
grafik
grafik

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/dark/main.less
```diff
@@ -210,6 +210,18 @@
     -fx-background-position: left;
 }
 
+.detailsPane {
+    .closeIcon {
+        -fx-background-image: url('/org/phoenicis/javafx/themes/default/icons/mainwindow/window-close.png');
+        -fx-background-size: 2em;
```
plata: Can we use `contain` here?

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/default/main.less
```diff
@@ -354,42 +366,42 @@
 }
 
 .listChooser {
-    -fx-spacing: 0.2em;
```
plata: Where are these changes coming from?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindowView.java
madoar: This change forces us to provide details views with a close button. I'm not sure if we want to allow details views that can't be closed by the user.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/default/main.less
madoar: IntelliJ formatting shortcut. It would be nice if our maven formatter plugin could format css files too :/

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindowView.java
plata: I don't see a reason to restrict it in the software if it is not required. In #817 we had the same situation: `VBox` instead of `Node` for no reason, now I had to change it because I use a `StackPane`.

plata:
Could you make the button a bit smaller so it has roughly the same height as the font?

madoar:
Do you have a suggestion in `em`?

plata:
Not really. Would have to try. Somehow it still looks a bit strange.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/default/main.less
plata: Yes. Do you also still have the 4/8 spaces formatting message?

madoar: Sometimes

plata: Any idea why that happens?

madoar: No, not really :/

madoar:
How about this:
grafik
grafik

Issue #808, opened 2017-05-21
create localization roadmap


plata:
Create a roadmap for the localization of Phoenicis. This includes:
* this application
* scripts (see PlayOnLinux/Scripts#49)
* website? (see #557)

For the Phoenicis application:
* decide how to do localization with Java(FX) (e.g. this [tutorial](https://docs.oracle.com/javase/tutorial/i18n/index.html), or with [GNU gettext](https://www.gnu.org/software/gettext/manual/html_node/Java.html))
* decide where to store the localization files in a way that people can provide localization files in an easy way
  * [Transifex](https://docs.transifex.com/formats/java-properties) (nice but expensive)
  * [Weblate](https://weblate.org/en/) (free, integrates with git/GitHub)
  * [Pootle](http://pootle.translatehouse.org/)
  * Rosetta (as in #4; often slow and buggy)

plata:
I'm currently looking at [gettext commons](https://github.com/jgettext/gettext-commons) which looks promising. However, I could not quite get it running yet.

madoar:
What problems did you meet?
I would like a solution for this, that can be updates via the web.
This would enable us to update the language files more often that the software itself, which would make sense when we fix some typos or want to add new languages

plata:
With the explanations from the gettext commons wiki, it's quite easy to add the Maven dependencies. Running `mvn gettext:gettext` will then give you a `pot` file with the extracted keys. I don't currently understand how to generate `po` files per language (if that can be done with Maven as well or not etc.) and how the generated classes/properties should be accessed from Java.

plata:
Here are some other tutorials:
<http://blogs.igalia.com/dpino/2009/08/30/setting-up-gettext-commons-for-i18n-java-files/>
<http://blogs.igalia.com/dpino/2009/09/05/using-gettext-commons-from-java-files/>

plata:
Ok. Getting closer:
* use [msginit](https://www.gnu.org/software/gettext/manual/gettext.html#msginit-Invocation) to create a `po` (`msginit -i keys.pot -o de.po`)
* call `mvn gettext:merge`
* call `mvn gettext:dist` (this creates a `Messages_de.properties`)

plata:
see #822

plata:
Translations are working. Translation service to be handled in #898.

PR #822, opened 2017-05-23
localization: gettext-commons proof of concept

plata:
Steps to try it:
1. run `mvn gettext:merge`
2. create `po` per language: `msginit -i keys.pot -o de.po` (for German)
3. translate (e.g. with poedit)
4. generate properties: `mvn gettext:dist`

The interesting part is in `Localisation.java`.

Here's a screenshot to prove that it works (changed "Library" to "Bibliothek"):
translate


madoar:
What would be needed to add a language, when we use this library? Is it possible to update a language file without creating a new `jar` file?

madoar:
When doing this we need to add an option to manually change the currently selected/shown language of POL 5

plata:
You can choose if you want `properties` or `class`. I'm not sure but I would assume that it is possible to use a `properties` file which is not part of the software. Does it even make sense to do this in our case? I don't think we will split Phoenicis into several .deb's.
It should be possible to select the language in the settings somehow. I think we can do that later on, however.

madoar:
I think it makes sense to use an external `properties` file for the language information, because this enables us to fix language typos etc. outside our normal release cycle.

plata:
That's only true if we split the translations into a separate deb and maintain two .deb's and two release cycles. I don't think we want to do that.

madoar:
My idea goes more in the direction of adding a function to the new website where you can query the newest version of the language files. If this query is sent automatically by POL 5 every time your start it and it has an internet connection you would always have the newest version of the language files without the need to update the software itself.

plata:
We can think about that in the future. Is it something which is usually done like this?

madoar:
We can think about this later, I just want to be sure that we can do such a thing when using this solution.
I'm not sure how you do it in practice, but I guess it depends on your release cycle. If you do a single release every 2 - 3 years it's quite useful to be able to do some things outside of the release cycle.

plata:
Every translation tutorial I saw for Java was using class, properties or resource bundle. I don't think that updating translations will be much different to this respect.

qparis:
Personaly I don't think we should let choosing the language in the settings. I mean, hardly any application is doing that if you think about it

qparis:
I think this implementation is very close to the way we did for POLv4 and is correct. 
Few remarks: 

- What do we plan to do for script translations. 
- Do we want to send translation to Rosetta?

plata:
@qparis we should discuss those topics in #808.

plata:
Would be good if we could merge this. I'm getting merge conflicts all the time.

plata:
I will merge this as @qparis said it's correct in general and to avoid further conflicts. We can improve on it later on.

Issue #747, opened 2017-05-01
AppPanel style is not updated


plata:
The style of the `WebView` in the `AppPanel` is not updated when you change the theme.

madoar:
I'm not sure how to do this, do you have an idea?

plata:
Not really. Maybe we could bind the `userStyleSheetLocationProperty` of `WebEngine` to the `ThemeManager` somehow (don't know if that would even reload the CSS). However, this issue is not important at all (you will not notice unless you switch themes all the time). Therefore it should not be assigned to "Beta".

plata:
Do you see now why I said that milestones are probably better suited than the project (looking at the amount of manual work you had updating this)?

madoar:
Yes I do -_-

Issue #711, opened 2017-04-15
Adding a StyleManager/CSSManager


madoar:
I don't like our current way of interacting with the stylesheets.
Ideally I would like to integrate the handling/interacting in the `ThemeManager` or a new class called `StyleManager` or `CSSManager`.
This class should encapsulate the interaction with stylesheets via an `set()` method, which takes two arguments:

- The first argument is an enum, which tells the method what type of stylesheet you want to set. This value can `DEFAULT_THEME`, `USED_THEME`, `CATEGORY_ICON_THEME` and maybe some other values we'll need.
Depending on the given enum value the method then decides which value inside the stylesheet-list it needs to change.
- The second argument is a File object. This file object is the new stylesheet to be set.

For context details see #681.

What do you think?

plata:
Makes sense. It should be done in `ThemeManager`. 

We should first think about the themes we need. I see different levels for this:
- DPI: defines only sizes depending on the screen resolution
- look'n'feel: default, Breeze Dark, Unity etc.
- icons
    - app category: default/themed
    - engine category: default/themed

madoar:
How would you set the DPI?
Only by changing the icons, or is there more to be done?
Another question about the DPI that appears to me is, do we then need to have multiple icon sets for each look'n'feel theme? (One in high DPI, another one in low?)

plata:
Currently, there's an (unfinished) HiDpi theme which is basically the default theme with everything enlarged. Therefore you cannot use e.g. the Unity theme on a HiDpi screen. I think it would be better to split the style and the sizes. For example, the Full HD size of a button could be 60x20 and the HiDpi size 120x40.

We do not need different icon sets because they are resized. Unfortunately, that's not possible if you set the button graphic in CSS so we had to set the background image (so the SearchBox has to use the same approach).

madoar:
About the SearchBox:
I have nothing against changing `-fx-graphic` to `-fx-background-image`. The reason I didn't use `-fx-background-image` was because you need to use some other css tags too if you want the image to be centered and not repeated.
In addition we need to set good insets for the searchfield clear button. 

About the HiDPI themes:
Is the HiDPI theme only about changing the size of the icons/elements and is this done uniformly (all elements are shown with 300% their width, height etc.) or is there more to it?
If we simply need to change the sizes of the components, then we can create one HiDPI css file, which contains the size information for **all** HiDPI themes, the same we can do for **all** low DPI versions.
[...]

PR #823, opened 2017-05-23
Improving class ThemeManager

madoar:
This PR mainly focuses on solving #711. In addition it does the following:
- made class `ThemeManager` to the central management class for themes, which is also responsible for maintaining the stylesheets of the scene
- moved the setting of the button images from the container tab to the theme less files
- replaced the `Theme` enum with an abstract `Theme` class and its subclass `ClasspathTheme`
- added the default theme to the stylesheet by default. This allows us to not explicitly redefine icons in all themes as long as they are defined in the default theme (see the added container tab images for example)


madoar:
In addition to #711 this PR now also fixes #747.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
```diff
@@ -19,6 +19,7 @@
 package org.phoenicis.javafx.controller;
 
 import javafx.application.Platform;
+import javafx.collections.ObservableList;
```
plata: why?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/themes/ClasspathTheme.java
```diff
@@ -0,0 +1,52 @@
+package org.phoenicis.javafx.views.common.themes;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * A theme located inside the classpath of POL 5
```
plata: please use "Phoenicis" instead of "POL 5" (also in the other places)

plata:
Can you remove the (now unused) icons in Unity and dark theme (should use the ones from default)?

What about the `WebView` in the app installation wizard? Do we also want to apply theme changes there? It's really an edge use case (you would have to change the theme while you are installing an app).

madoar:
Do we have easy access to the `ThemeManager` in the app installation wizard?

plata:
I didn't check that.

plata:
I didn't check that.

PR #824, opened 2017-05-24
added close icon for Breeze Dark theme

plata:


madoar:
Screenshot please :)
In addition: What happens if you only specify the new image inside the `breezeDark` theme and remove the other lines, are the missing parts taken from the default theme?

plata:
I didn't remove any line.

plata:
screenshot

madoar:
I know, I would like to know what would happen if you remove them.

plata:
Ah ok. I will try.

plata:
Seems to work as well. If we want to use that, we should have a separate task to remove everything which is already defined in default from the other themes.

madoar:
We should think about it, but we also need to have in mind that we may need to do some changes in themes because of the default theme.
Let's say for example the default theme uses a border for something for which another theme uses a background color, then we need to "deactivate" the border for the other theme too.

PR #828, opened 2017-05-25
Fixed CLI

qparis:


madoar:
Can you add a test for the CLI to ensure that we won't accidentally break it again in the future?

qparis:
@madoar Can we merge this?

PR #829, opened 2017-05-25
Removed unnecessary dependency to json

qparis:


qparis:
(PR #828 should be merged first)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/ViewLibrary.java
```diff
@@ -58,20 +57,18 @@
...
 
-    public ViewLibrary(String applicationName, ThemeManager themeManager) {
+    public ViewLibrary(String applicationName, ThemeManager themeManager, ObjectMapper objectMapper) {
         super("Library", themeManager);
         this.getStyleClass().add("mainWindowScene");
+        this.objectMapper = objectMapper;
 
         availableShortcuts = new CombinedListWidget<>(ListWidgetEntry::create, (selectedItem, event) -> {
-            ShortcutDTO shortcutDTO = selectedItem;
+            final ShortcutDTO shortcutDTO = selectedItem;
```
madoar: This whole line is not needed, because we're simply doing a rename here.
I think it would be better to use `selectedItem` instead of `shortcutDTO`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/LibraryPanel.java
```diff
@@ -58,20 +62,25 @@ private void populateContent() {
         gridPane.getStyleClass().add("grid");
 
         try {
-            JSONObject jsonObject = new JSONObject(shortcut.getScript());
+            System.out.println(shortcut.getScript());
```
madoar: This looks like a forgotten debug output.
Here we should use `LOGGER` or remove it.

qparis: Sorry, forgot that

pom.xml
```diff
@@ -73,6 +73,12 @@
                 <version>${mockito.version}</version>
                 <scope>test</scope>
             </dependency>
+
```
madoar: Here is an empty line too much (I know it's not much)

pom.xml
```diff
@@ -73,6 +73,12 @@
...
+			<dependency>
+				<groupId>com.fasterxml.jackson.core</groupId>
+				<artifactId>jackson-databind</artifactId>
+				<version>2.7.3</version>
+			</dependency>
```
madoar: The formatting seems to be wrong here, right?

PR #832, opened 2017-05-25
Allow maven to run phoenicis

qparis:


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/ViewsConfigurationLibrary.java
```diff
@@ -39,7 +39,11 @@
 
     @Bean
     public ViewLibrary viewLibrary() {
-        return new ViewLibrary(applicationName, themeConfiguration.themeManager(), phoenicisGlobalConfiguration.objectMapper());
+        return new ViewLibrary(
```
qparis: The maven formatter plugin is messing this. Do you know how we could improve it?

madoar:
Just out of interest: How does the plugin allow maven to run POL 5?

qparis:
Just run maven java:exec in the javafx module

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/ViewsConfigurationLibrary.java
madoar: My best guess is, that the problem is inside the formatter configuration.
@plata and me, we updated the formatter plugin version in #767, because the old version didn't do its job with the indention. Maybe the new plugin version requires a newer version of the formatter configuration.

madoar: In addition how do you run the plugin?
If I run it, the plugin is working fine (the given parameters are split to two lines).
I did run the plugin via `mvn test`

madoar: Did the plugin work for you with `mvn test` or did it at least give you a different formatting?

PR #841, opened 2017-05-29
Unity theme: use white text for selected tabs and category buttons

### LINKED ISSUE #838: [Unity theme] The text on the orange background should be white color.
The text on the orange background should be white color.
zrzut ekranu z 2017-05-28 13-16-28
This is image with the white text edited by me in gimp (:P)
zrzut ekranu z 2017-05-28 13-16-28editybygimp

plata:
fixes #838 

tab

### DIFF
--- a/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/unity/main.less
+++ b/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/unity/main.less
@@ -127,8 +127,8 @@
     -fx-border-radius: 0;
 }
 
-.tab-header-area .tab:selected .tab-label {
-    -fx-text-fill: white;
+.tab-header-area .tab:selected .tab-label .text {
+    -fx-fill: white;
 }
 
 .list-view {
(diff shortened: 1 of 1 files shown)


madoar:
As far as I understand #838 all text should in the `TabPane` labels and the selected category should be white.

plata:
I changed the color for selected left buttons but kept the tabs as they are (I like how this emphasizes the selected button). 
Please have a look at the change I did because I do not quite understand why this was necessary. Could it be related to LESS somehow?

white button text

madoar:
@plata What exactly do you mean?

plata:
It was
```CSS
    -fx-text-fill: white;
```
now it is
```CSS
.text {
    -fx-fill: white;
}
```
I don't understand why the `-fx-text-fill` does not work.

madoar:
I don't think the reason is the less. Can you take a look at the created `css` file in the target folder, just to be sure?
As far as I understand the cssref `-fx-text-fill` is only defined for `Labeled` and `TextInputControl` components, I'm not sure the tab header and the list view belong in these categories?

plata:
At least the `tab-label` should be a `Label` according to the reference.

madoar:
After thinking a little more about this, I believe this may be related to the fact, that we are always loading the default theme.

plata:
Ok... Anything we should change there? Isn't this the same behavior as Java does it (always load the Java default theme)?

madoar:
Without taking a look at the theme css/less code:
It may be possible that `-fx-fill` takes precedence over `fx-text-fill`.
If this is the case, then the behavior can be expected, because the default theme uses `-fx-fill` to define the text color, while unity used `-fx-text-fill`.

madoar:
What we could do would be to change all occurrences of `-fx-fill` to `-fx-text-fill` and hope that this has no unwanted side effects...

PR #847, opened 2017-05-30
Add filter functionality for the unused checkboxes inside the "Apps" tab

### LINKED ISSUE #731: Implement functionality for the filter checkboxes in the "Apps" tab
Currently we have 3 checkboxes in the sidebar of the "Apps" tab, that do absolutely nothing.
We should add some functionality to them.

madoar:
Fixes #731.
grafik
grafik
grafik
grafik
grafik
grafik
grafik

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/AppPanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/AppPanel.java
@@ -18,40 +18,44 @@
 
 package org.phoenicis.javafx.views.mainwindow.apps;
 
-import javafx.scene.control.Button;
-import javafx.scene.control.Label;
-import javafx.scene.control.ScrollPane;
-import javafx.scene.control.Tooltip;
+import javafx.collections.FXCollections;
+import javafx.collections.transformation.FilteredList;
+import javafx.scene.control.*;
 import javafx.scene.layout.*;
 import javafx.scene.web.WebView;
-import org.phoenicis.repository.dto.ApplicationDTO;
-import org.phoenicis.repository.dto.ScriptDTO;
 import org.phoenicis.javafx.views.common.ErrorMessage;
 import org.phoenicis.javafx.views.common.ThemeManager;
 import org.phoenicis.javafx.views.common.widgets.lists.DetailsView;
+import org.phoenicis.repository.dto.ApplicationDTO;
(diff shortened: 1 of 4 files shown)


plata:
Does it make sense to have "Free" and "Commercial"? The only use case for this would be to find only commercial apps but why should you do that? In my opinion it would be sufficient to have only one of them.

madoar:
grafik
grafik
grafik
grafik

plata:
Good. What's the default setting for the checkboxes? I think "Commercial" should be checked by default and the other ones not. Are the filters saved and reloaded on the next startup?

madoar:
The default setting is like you described, with only "Commercial" checked by default.
The filters aren't saved and reloaded on the next startup.

plata:
Just tried it. The "Testing" filter does not seem to work.

madoar:
What do you mean with not work? Does it nothing or something unexpected?

plata:
It did not show an application which has the following settings in its json even though I selected "Testing":
```JSON
"compatibleOperatingSystems": [],
"testingOperatingSystems": ["LINUX"],
```

madoar:
Did you maybe use some other filters too, which influences your result? 
The "Testing" checkbox only has an effect on your result if it's not selected.
Otherwise it would be good to add some debug output to the filter method in `ApplicationFilter`.

plata:
I think I also had commercial. What do you mean with "only has an effect on your result if it's not selected."?

madoar:
This is how the "testing" checkbox is used during the `filter` call:
```
+        /*
+         * If "Testing" is not selected don't show games that are currently in a testing stage
+         */
+        if (!containTestingApplications.getValue()) {
+            result &= application.getScripts().stream()
+                    .anyMatch(script -> script.getTestingOperatingSystems().isEmpty());
+        }
```

plata:
Shouldn't it be `if (!containTestingApplications.getValue()) {` (without the `!`)? You only need the `!` for commercial because commercial = !free.

madoar:
No, because:

- if you don't select "Testing", you want to see all applications that aren't in a testing stage (which is the case if the testing operating systems array/list is empty)
- if you select "Testing", you want to see all applications

right?

plata:
Yes. I've been mistaken. Anyways, I just figured out that the following works:
```JSON
"compatibleOperatingSystems": ["LINUX"],
"testingOperatingSystems": ["LINUX"],
```

PR #854, opened 2017-05-30
Add a new method getMainMiniature() to the ApplicationDTO

### LINKED ISSUE #785: Add a method to ApplicationDTO to get the main miniature
Currently the class `ApplicationDTO` contains no method to return the main miniature belonging the application, which can be identified by its filename which is `main.png`. 
This main miniature is used for in the `MiniatureListWidget` and eventually during the shortcut creation.

madoar:
Fixes #785.

### DIFF
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
@@ -22,10 +22,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.List;
+import java.util.*;
 
 /**
  * Represents an application
@@ -70,6 +67,27 @@ public class ApplicationDTO {
 
     public List<ScriptDTO> getScripts() {
         return scripts;
+    }
+
+    /**
+     * Returns the main miniature belonging to this {@link ApplicationDTO}.
+     * The main miniature is the miniature with the file name <code>main.png</code>.
+     * If no such miniature exists the first miniature of this {@link ApplicationDTO} is returned.
+     * If this {@link ApplicationDTO} contains no miniatures {@link Optional#empty()} is returned.
+     *
+     * @return An optional with the found URI, or {@link Optional#empty()} if no miniature exists
+     */
+    public Optional<URI> getMainMiniature() {
+        Optional<URI> result = this.miniatures.stream().filter(uri -> uri.getPath().endsWith("main.png")).findFirst();
+
+        /*
+         * Fallback in case no main miniature has been selected but the list contains at least one other miniature
+         */
+        if (!result.isPresent()) {
+            result = this.miniatures.stream().findFirst();
+        }
+
+        return result;
     }
 
     public static Comparator<ApplicationDTO> nameComparator() {
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
@@ -18,6 +18,7 @@
 
 package org.phoenicis.repository.dto;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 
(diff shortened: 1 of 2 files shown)


phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
```diff
@@ -72,6 +69,27 @@ public String getDescription() {
...
+    /**
+     * Returns the main miniature belonging to this {@link ApplicationDTO}.
+     * The main miniature is the miniature with the file name <code>main.png</code>.
+     * If no such miniature exists the first miniature of this {@link ApplicationDTO} is returned.
+     * If this {@link ApplicationDTO} contains no miniatures {@link Optional#empty()} is returned.
+     *
+     * @return An optional with the found URI, or {@link Optional#empty()} if no miniature exists
+     */
+    public Optional<URI> getMainMiniature() {
+        Optional<URI> result = this.miniatures.stream().filter(uri -> uri.getPath().endsWith("main.png")).findFirst();
```
qparis: A DTO cannot return an Optional type, Jackson would not serialize it properly.

In principle, the responsibility of à DTO is *just* to hold data. Any logic should be put in another place

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
```diff
@@ -72,6 +69,27 @@ public String getDescription() {
...
+
+        /*
+         * Fallback in case no main miniature has been selected but the list contains at least one other miniature
+         */
+        if (!result.isPresent()) {
+            result = this.miniatures.stream().findFirst();
```
madoar: I'm not sure if we need this fallback, because if no `main.png` file is found we would use a random miniature that may or may not be useful for identifying the application. 
I guess in this case it may be more useful to show a generic fallback image, which we show if `Optional.empty()` is returned.
What do you think?

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
madoar: This shouldn't be serialized, which is why it has no builder method and no private field.
It's just for the benefit of accessing the main miniature.
The old way it was handled was by reordering the miniatures, so that the main miniature was in the first position, which is not intuitive.

qparis: The builder is useful to deserialize, not to serialize. 
By default, Jackson will serialize all getters.
I will probably add a test that checks all DTOs soon

madoar: Isn't there an annotation or other way to specify methods that should not be serialized?
I know from other libraries (JPA, JAX-B) that such libraries normally contain ways to ignore fields/methods, I guess this should be possible for Jackson too, right?

qparis: @JsonIgnore is what you are looking for

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
plata: I'm not sure but any image related to the application is probably more useful for identifying it than a generic fallback image.

qparis: The problem is that you could have screenshots starting to appear in the list of available apps which would look messy

madoar: Another question from me:
What does the `icon` of an application contain? Is this the same as the main miniature?

qparis: The icon was designed to let the scriptor put an icon so that it can be presented like in POLv4

madoar: This sounds like the main miniature to me?
Is this icon currently used somewhere in POL 5?

qparis: The main miniature is not an icon
It was used before (https://www.playonlinux.com/en/comments-1299.html)

madoar: Do we want to retain the icon or do we not need it anymore?

qparis: I think it could be a good idea to have in in the future

madoar: Then we're back to the first question:
Shall I remove the fallback, to take a random image if no `main.png` could be found?

plata: @madoar maybe you should just merge the current functionality and open a new issue for this.

PR #855, opened 2017-05-31
Remove the miniature duplicates during the merging of two applications

### LINKED ISSUE #848: Duplicate miniatures when using multiple repositories
When you have two, or more, repositories with the same script miniature, it is shown twice:
grafik
We should add a function, which checks for duplicates to prevent POL 5 from showing duplicate miniatures.

madoar:
Fixes #848.
grafik

### DIFF
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/MergeableRepository.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/MergeableRepository.java
@@ -1,28 +1,24 @@
 /**
- * 
+ *
  */
 package org.phoenicis.repository.repositoryTypes;
 
-import java.net.URI;
-import java.nio.ByteBuffer;
-import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.SortedMap;
-import java.util.TreeMap;
-import java.util.function.Function;
-
+import org.apache.commons.codec.digest.DigestUtils;
 import org.phoenicis.repository.dto.ApplicationDTO;
 import org.phoenicis.repository.dto.CategoryDTO;
 import org.phoenicis.repository.dto.ResourceDTO;
 import org.phoenicis.repository.dto.ScriptDTO;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.*;
+import java.util.function.Function;
 
 public abstract class MergeableRepository implements Repository {
+    private final static org.slf4j.Logger LOGGER = LoggerFactory.getLogger(MergeableRepository.class);
+
     /**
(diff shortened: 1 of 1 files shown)


phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/MergeableRepository.java
```diff
@@ -107,12 +101,46 @@ protected ApplicationDTO mergeApplications(ApplicationDTO leftApplication, Appli
...
+         */
+        List<URI> miniatures = new ArrayList<>(leftMiniatures);
+        miniatures.addAll(rightMiniatures);
+
+        /*
+         * Remove duplicates
+         */
+        for (URI miniatureUri : miniatures) {
+            try (InputStream inputStream = miniatureUri.toURL().openStream()) {
+                String checksum = DigestUtils.md5Hex(inputStream);
```
qparis: Why not directly comparing byte to byte? It would be more efficient wouldn't it?

madoar: You mean of the files?
I believe it's longer and it wasn't this intuitive to write^^

madoar: Another problem would be that we need to save a representation for each miniature, otherwise we would need O(n*n) comparisons.

qparis: In fact it is shorter as you need to read all the bytes to calculate the checksum.
Have you checked out Arrays.equal?

qparis: Why n*n? 
You just need n operations

madoar: Are you sure, because I need to compare all images with each other, right?

qparis: Yeah, my bad, that's correct.
Then the right way to do it would be to use Arras.hashCode.
However, it seems that you are going to download the same uri several time with this solution, correct?

madoar: These `URI`s should be local files. I'm only converting them to `URL` objects, because it's quite easy to create an `InputStream` from an `URL`. 
When we later add an `OnlineRepository`, that uses REST for example, I would also suggest to download the full repository, save it locally and update it at startup, like we do for our `GitRepository`.
Because we've local files behind the `URI` objects this should be quite fast, right?

qparis: Yep, so far so good.
However, concerning the downloading of the whole repository, it is not that easy because basically, it takes something like one line of code if you decide to create webservices holding the DTO structure. 

If you want to fetch the whole repository as you described, it will not be as straightforward.

We will handle this later anyway.

PR #863, opened 2017-06-03
code formatter changes

plata:


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/ViewsConfigurationLibrary.java
```diff
@@ -39,11 +39,8 @@
 
     @Bean
     public ViewLibrary viewLibrary() {
-        return new ViewLibrary(
-                applicationName,
-                themeConfiguration.themeManager(),
-                phoenicisGlobalConfiguration.objectMapper()
-        );
+        return new ViewLibrary(applicationName, themeConfiguration.themeManager(),
```
qparis: It would be great if the code formatter would keep the other way of presenting beans parameters. It is a lot more readable

plata: It's only using the settings xml. What I don't understand is why some files are obviously not formatted.

qparis: It is probably because I did not run mvn compile before comitting.

plata: I see. Can we somehow make sure that this always happens? Or at least Jenkins gives an error? What I did once is to run `git status` in the Jenkins build and return an error if something changed.

qparis: I don't see how it could change something

plata: It would at least prevent that unformatted code is merged into master.

PR #864, opened 2017-06-03
German translation

plata:
Make sure all relevant Strings are translated and add German translation.

### DIFF
--- a/i18n/de.po
+++ b/i18n/de.po
@@ -9,7 +9,7 @@ msgstr ""
 "Project-Id-Version: POL-POM 5\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2017-06-03 16:47+0200\n"
-"PO-Revision-Date: 2017-06-03 16:48+0200\n"
+"PO-Revision-Date: 2017-06-03 18:13+0200\n"
 "Last-Translator: Plata <plata@phoenicis.org>\n"
 "Language-Team: German\n"
 "Language: de\n"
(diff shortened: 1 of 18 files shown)


protux:
Hi,
the translation looks good, just some remarks:
- the translation for "theme" is Thema / Erscheinungsbild, regarding the code it seems to be the long name which is translated there, so maybe the theme names should get an extra "Thema" added. (the short names translation seems to be missing due to missing msgids)
- translating "Refresh Repositories" with "Repositories auffrischen" might be basically correct, but it sounds weird. I think it is more common to call it "Repositories aktualisieren".
- I guess you left some translations empty by intent to use the same in german as in english, but I would translate "Video memory size" with "Videospeichergröße" or "Größe des Videospeichers"
- same as previous: "Strict Draw Ordering" with "feste Darstellungsreihenfolge"
- same as previous: "Always Offscreen" with "Immer offscreen"
- I'm not quite sure about the context, but translating "Installers" only with "Scripts" might be confusing. If the only use for scripts is installing stuff that might be ok, otherwise I would call it "Installationsskripte".

I guess the target group are technical people (developers etc.) so it is ok not to translate "scripts" etc., but if a common user (the arverage youtube/ facebook/ gaming audiance) is also part of the target group I would reduce the "denglish" words.

plata:
- I left away the "theme" part of "default theme" on purpose because I think that "Breeze Dunkel Erscheinungsbild" is too long. Moreover, the checkbox has the title "Erscheinungsbild" already.
- That's true. I was looking for another translation but didn't think of "aktualisieren".
- Any other "denglish" word I should translate? I was thinking about "Engines" but couldn't come up with a useful translation.

protux:
I thought about Engine and Executable. But I have no idea for a good translation. Executable might be "Ausführbare Datei" but that might be a little long. A fitting translation for "Engine" might be "Treiber", but there I'm not sure.

plata:
Let's leave it like it is until we have a better idea.

madoar:
I think "Platform" is a better translation for "Engine" in our case.

plata:
I'm not sure we should do that. I must be possible to discuss in English about a problem if you are using a German GUI. If someone writes in the forum "I have issue xyz in the Platform tab" no-one will know what she's talking about.

madoar:
I agree. I just added "Platform" as a better alternative to "Treiber" in my eyes

PR #867, opened 2017-06-03
Automatically test all DTOs

qparis:


plata:
The PR has many changes which are not related to this PR. This makes it quite difficult to review it properly. We do not have this issue in other reviews so I don't quite understand why this happens for your last PRs.

qparis:
Don't understand either... 
I think I need to refork the whole project. The only files I have changed are those whose names contain DTO

plata:
So this fixes #686 as well? (Is it just me or is the GitHub search broken?)

qparis:
Yep it does fix this :)

plata:
Ok. Then I would like @madoar to do the review because he opened that issue.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/LibraryPanel.java
```diff
@@ -118,6 +122,10 @@ public void setShortcutDTO(ShortcutDTO shortcutDTO) {
         this.setCenter(vBox);
     }
 
+    private String unCamelize(String s) {
```
madoar: What exactly does this method? 
Some JavaDoc would be nice here

phoenicis-tests/src/test/java/org/phoenicis/tests/DTOContractTest.java
```diff
@@ -0,0 +1,280 @@
...
+import java.net.URISyntaxException;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.junit.Assert.*;
+
+@RunWith(Parameterized.class)
+public class DTOContractTest {
+    private final static int DEFAULT_MAX_DEPTH = 5;
```
madoar: For what is this variable used?
I guess for some kind of recursion?
Again some JavaDoc would be nice.

phoenicis-tests/src/test/java/org/phoenicis/tests/DTOContractTest.java
```diff
@@ -0,0 +1,280 @@
...
+
+        return null;
+    }
+
+    /**
+     * This aim of the class is to customize the content of
+     * the name of the test suit
+     *
+     * @param <C> The class the
+     */
```
madoar: Please check the JavaDoc of this class again.

phoenicis-tests/src/test/java/org/phoenicis/tests/DTOContractTest.java
```diff
@@ -0,0 +1,280 @@
...
+        assertTrue(json.endsWith("}"));
+    }
+
+    @Test
+    public void testDeserialize() throws ReflectiveOperationException, IOException, URISyntaxException {
+        final Object dto = newDTOInstance(DTOContainer.getClazz());
+        final String json = objectMapper.writeValueAsString(dto);
+
+        final ObjectMapper alternativeObjectMapper = new ObjectMapper();
+        alternativeObjectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
madoar: This sounds like we're allowing unknown properties in the json, when we want to deserialize it.
Do we also allow this in production?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/LibraryPanel.java
qparis: This is not related to this PR. 
See @plata comment

qparis:
Sure!
Can we merge this first to allow others PR to be merged? (My fork is messed up)
I will add extra comments in a separate pull request

phoenicis-tests/src/test/java/org/phoenicis/tests/DTOContractTest.java
```diff
@@ -0,0 +1,280 @@
...
+        final Object dto1 = newDTOInstance(DTOContainer.getClazz());
+        final Object dto2 = newDTOInstance(DTOContainer.getClazz());
+        assertEquals(dto1, dto2);
+        assertEquals(dto1.hashCode(), dto2.hashCode());
+    }
+
+    @Test
+    public void testEqualsTrivialCases() throws ReflectiveOperationException, URISyntaxException {
+        final Object dto = newDTOInstance(DTOContainer.getClazz());
+        assertTrue(dto.equals(dto));
```
madoar: Just asking:
Is there a difference between: `assertTrue(dto.equals(dto));` and `assertEquals(dto,  dto);`?

qparis: No fondamental differences but in this particular case, we are testing the equals method. So we really want to call it explicitely

PR #868, opened 2017-06-04
fixed missing title in details panel

### LINKED ISSUE #865: title missing in details panel
see #862

plata:
fixes #865 

title

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/LibraryPanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/LibraryPanel.java
@@ -47,6 +47,12 @@ final class LibraryPanel extends DetailsView {
 
     private final ObjectMapper objectMapper;
 
+    private Label description;
+    private GridPane gridPane;
+    private Button runButton;
+    private Button stopButton;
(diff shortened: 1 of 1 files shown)


qparis:
Can you add little margins so that the title is aligned with the rest?

plata:
I don't know (this is done in the base class so it might not work in all places).

madoar:
This looks like something is overriding the style of the base class.
Take a look at the other tabs, for example the "Apps" tab, there you'll discover that the margins are larger. 
This is the reason why I previously wrote we need to refactor the details tabs, because some of them are defining their of css classes, which makes them look not uniform.

plata:
In this case we should not do this within this PR.

plata:
What happened to Jenkins?

plata:
run tests

madoar:
Hmm, it seems like we get a timeout inside the `AppConfigurationTest`:
```
java.util.concurrent.TimeoutException
	at org.phoenicis.javafx.AppConfigurationInitialisationTest.testAppConfigurationInitialisation(AppConfigurationInitialisationTest.java:21)
```
@qparis is there anything you can do about this?

qparis:
There were some problems with Jenkins yesterday. Should be fixed

plata:
Should be ok to merge now.

PR #872, opened 2017-06-04
Scripts i18n proof-of-concept

plata:
`resources` should be filled with the merged content of the language `.json` files which have to be loaded from every used repository. To do this, we need a `RepositoryDTO` with a list of `CategoryDTO` and a `Map` for the translations (language, json String).

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/NashornScriptInterpreter.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/NashornScriptInterpreter.java
@@ -18,23 +18,52 @@
 
 package org.phoenicis.scripts.nashorn;
 
+import org.phoenicis.repository.dto.RepositoryDTO;
+import org.phoenicis.repository.dto.TranslationDTO;
 import org.phoenicis.scripts.interpreter.InteractiveScriptSession;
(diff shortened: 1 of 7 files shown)


phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/NashornScriptInterpreter.java
```diff
@@ -32,7 +33,30 @@ public NashornScriptInterpreter(NashornEngineFactory nashornEngineFactory) {
 
     @Override
     public void runScript(String scriptContent, Consumer<Exception> errorCallback) {
-        nashornEngineFactory.createEngine().eval(scriptContent, errorCallback);
+        final String language = Locale.getDefault().getLanguage();
+        final String i18nextInit = "i18next.init({                                                                     "
+                + String.format("  lng: '%s',                                                             ", language)
```
madoar: I don't like to define code as a String.
This looks quite ugly and hard to maintain.
Is there another way to set this?

plata: I don't think so (I have to call JS).

qparis:
I still don't understand why we are not using Java tr here

madoar:
@plata is this PR still required?

plata:
No.

PR #873, opened 2017-06-04
added RepositoryDTO

plata:
new hierarchy level above CategoryDTO

required by #872

### DIFF
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/RepositoryDTO.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/RepositoryDTO.java
@@ -25,7 +25,7 @@ import org.apache.commons.lang.builder.ToStringBuilder;
 import java.util.*;
 
 /**
- * Represents a category of application
+ * Represents a repository
  */
 @JsonDeserialize(builder = RepositoryDTO.Builder.class)
 public class RepositoryDTO {
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/RepositoryDTO.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/RepositoryDTO.java
@@ -33,15 +33,15 @@ import java.util.*;
 public class RepositoryDTO {
     private final String name;
     private final List<CategoryDTO> categories;
-    private final Map<String, String> translations;
+    private final Set<TranslationDTO> translations;
 
     private RepositoryDTO(Builder builder) {
         this.name = builder.name;
         this.categories = Collections.unmodifiableList(builder.categories);
-        this.translations = Collections.unmodifiableMap(builder.translations);
+        this.translations = Collections.unmodifiableSet(builder.translations);
     }
 
-    public Map<String, String> getTranslations() {
+    public Set<TranslationDTO> getTranslations() {
         return translations;
     }
 
@@ -82,7 +82,7 @@ public class RepositoryDTO {
     public static class Builder {
         private String name;
         private List<CategoryDTO> categories = new ArrayList<>();
-        private Map<String, String> translations = new HashMap<>();
+        private Set<TranslationDTO> translations = new HashSet<>();
 
         public Builder() {
             // Default constructor
@@ -107,7 +107,7 @@ public class RepositoryDTO {
             return new RepositoryDTO(this);
         }
 
-        public Builder withTranslations(Map<String, String> translations) {
+        public Builder withTranslations(Set<TranslationDTO> translations) {
             this.translations = translations;
             return this;
         }
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/RepositoryDTO.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/dto/RepositoryDTO.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) 2015-2017 PÂRIS Quentin
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.repository.dto;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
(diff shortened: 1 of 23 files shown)


qparis:
Please merge my pull request first so that you new DTOs are tested

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/RepositoryDTO.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.*;
+
+/**
+ * Represents a repository
+ */
+@JsonDeserialize(builder = RepositoryDTO.Builder.class)
+public class RepositoryDTO {
+    private final String name;
+    private final List<CategoryDTO> categories;
+    private final Map<String, String> translations;
```
qparis: In general, we avoid using Map<String, String> in DTOs at it does not really say something about the data held.

Could we use a Set of TranslationDTO instead

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/RepositoryDTO.java
```diff
@@ -0,0 +1,97 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import org.apache.commons.lang.builder.ToStringBuilder;
+
+import java.util.*;
+
+/**
+ * Represents a repository
+ */
+@JsonDeserialize(builder = RepositoryDTO.Builder.class)
+public class RepositoryDTO {
```
qparis: Please add equals and hashCode

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/ClasspathRepository.java
```diff
@@ -63,10 +64,13 @@ public ClasspathRepository(String packagePath, ResourcePatternResolver resourceR
...
             Collections.sort(categoryDTOs, Comparator.comparing(CategoryDTO::getName));
-            return categoryDTOs;
+
+            final RepositoryDTO.Builder repositoryDTOBuilder = new RepositoryDTO.Builder()
+                    .withName("classpath repository").withCategories(categoryDTOs);
+            return repositoryDTOBuilder.build();
         } catch (IOException e) {
             LOGGER.warn("Error while reading resource directory", e);
-            return Collections.emptyList();
+            return null;
```
qparis: You should return an empty RepositoryDTO here

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/GitRepository.java
```diff
@@ -71,11 +67,11 @@ public GitRepository(URI gitRepositoryUri, String cacheDirectoryPath,
                 LOGGER.error(String.format("Couldn't create folder for git repository '%s' at '%s'",
                         this.gitRepositoryUri, gitRepositoryLocation.getAbsolutePath()));
 
-                return Collections.emptyList();
+                return null;
```
qparis: Please return an empty RepositoryDTO here

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/LocalRepository.java
```diff
@@ -60,15 +57,34 @@ private LocalRepository(File repositoryDirectory, ObjectMapper objectMapper) {
...
     @Override
-    public List<CategoryDTO> fetchInstallableApplications() {
+    public RepositoryDTO fetchInstallableApplications() {
+
         final File[] categoryDirectories = repositoryDirectory.listFiles();
 
         if (categoryDirectories == null) {
-            return Collections.emptyList();
+            return null;
         }
```
qparis: Please return an empty RepositoryDTO here

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/NullRepository.java
```diff
@@ -19,15 +19,12 @@
...
-import java.util.Collections;
-import java.util.List;
+import org.phoenicis.repository.dto.RepositoryDTO;
 
 public class NullRepository implements Repository {
     @Override
-    public List<CategoryDTO> fetchInstallableApplications() {
-        return Collections.emptyList();
+    public RepositoryDTO fetchInstallableApplications() {
+        return null;
```
qparis: Same here

phoenicis-repository/src/test/java/org/phoenicis/repository/repositoryTypes/NullRepositoryTest.java
```diff
@@ -19,13 +19,12 @@
...
 import org.junit.Test;
-import org.phoenicis.repository.repositoryTypes.NullRepository;
 
 import static org.junit.Assert.assertEquals;
 
 public class NullRepositoryTest {
     @Test
     public void testNullRepositoryTest() {
-        assertEquals(0, new NullRepository().fetchInstallableApplications().size());
+        assertEquals(null, new NullRepository().fetchInstallableApplications());
```
qparis: You may want to use assertNull

phoenicis-tests/src/main/java/org/phoenicis/tests/MockedRepository.java
```diff
@@ -38,17 +38,21 @@
...
-                                                            OperatingSystem.MACOSX))
-                                                    .withScript(wineScript()).build()))
-                                            .build()))
-                                    .withName("Functions").build());
+        public RepositoryDTO fetchInstallableApplications() {
+            return new RepositoryDTO.Builder()
+                    .withCategories(
+                            Collections.singletonList(
+                                    new CategoryDTO.Builder()
+                                            .withApplications(Arrays.asList(new ApplicationDTO.Builder().withName(
```
qparis: The formating here is strange

qparis:
Please merge the master to this branch to run DTO tests

phoenicis-tests/src/main/java/org/phoenicis/tests/MockedRepository.java
plata: I cannot change it. The formatter does it this way.

qparis: Then the formatter needs to be fixed or we need to remove it. 
The point of having code conventions is to make it as more readable as possible

qparis:
With your solution, what if we have two scripts that have the same name in two different repositories?

phoenicis-tests/src/main/java/org/phoenicis/tests/MockedRepository.java
plata: I guess in this particular case, we should build the objects step by step and not in one big call. Anyways this is completely unrelated to the PR.

plata:
The merge process should not have changed, so I would say: same behavior as before.

phoenicis-tests/src/main/java/org/phoenicis/tests/MockedRepository.java
qparis: One advantage of the builder pattern is to avoid building the objects step by steps. 
Code formatting should help us. If it is not the case, we should get rid of it. 
This is not the only place where it causes problems (see Spring configuration)
Agree that it is not related to this PR, but we should track this issue though

PR #874, opened 2017-06-05
split creation of RepositoryDTO in MockRepository

plata:


qparis:
I don't understand the point of this PR
This won't exempt us from fixing the formatter issue in my opinion

plata:
Maybe not. But on the other hand the structure got quite complex and wasn't well understandable anymore. Now it should be clearer what's going on. As a side effect, the formatting does not look so strange anymore.

qparis:
To be honnest, I find this a lot more easier to read:

qparis:
Same problem for Spring configuration: 
capture d ecran 2017-06-05 a 15 57 11

vs. 

capture d ecran 2017-06-05 a 15 57 29

qparis:
Also equals() and hashCode() ...
capture d ecran 2017-06-05 a 15 58 28
vs.
capture d ecran 2017-06-05 a 15 58 43

qparis:
I think you got my point

plata:
I agree that the second formatting looks better. I just don't know enough about the settings to get this behavior. I just wouldn't like to remove the formatter completely because that will totally mess up the code on the long run.

plata:
Maybe this: <https://stackoverflow.com/questions/7166795/can-eclipse-formatter-wrap-where-necessary-every-element-on-a-new-line>

plata:
Can you try if you prefer:
```xml
<setting id="org.eclipse.jdt.core.formatter.join_wrapped_lines" value="false"/>
```
That should keep a line wrap.

qparis:
Sounds that it is what we need

madoar:
How is the current state of this PR?

PR #876, opened 2017-06-06
translate app categories

plata:


plata:
It's the same issue as in #872: If we use `tr(category.getName())`, we have to provide the translations as part of the Phoenicis software. As a result, it would not be possible for the users to add their own (translated) categories.

qparis:
For this reason, it is really clear for me that we really need to find a way to use the same translation framework for Java and Javascript.
It makes a lot more sense

madoar:
This PR has a conflict.

plata:
PR is outdated. I'm just leaving it open for now because I might need some snippets from it. Will close it later.

Issue #857, opened 2017-06-01
Select branch for GitRepository


madoar:
Currently when adding a `GitRepository` to POL 5, the master branch is cloned.
It would be useful to have the ability to select a branch of a git repository in which the POL 5 application repository is located.
I guess this should be implemented together with #776.

madoar:
I've thought about this a bit. 
I believe we will face two main challenges when we try to do this:

- we will need an interface that allows us to add the branch **in addition** to the repository url. I guess this could be solved best by #776.
- we will need to save the branch information next to the repository url. I'm not sure how to approach this. I don't know of a way to include the branch in the repository url, therefore we would need another way to save it.

plata:
We could add another separator e.g. something like "git+https://github.com/PlayOnLinux/Scripts#master".

plata:
Or like GitHub (`tree` as separator) "git+https://github.com/PlayOnLinux/Scripts/tree/master".

madoar:
I'm not sure if `tree` is a good separator, because this could be a valid part of the path to a git repository.
Can `#` be used inside a typical url to a git repository?

plata:
Just created two branches: `a#b` and `a/tree` so it's probably not a good idea...

madoar:
Is it maybe possible to persist the repositories in json or xml, instead of like we currently do?

plata:
Maybe but that would be a big effort I think.

plata:
Check this out: <https://stackoverflow.com/questions/3651860/which-characters-are-illegal-within-a-branch-name>.

According to this, we could use "git+https://github.com/PlayOnLinux/Scripts:master".

madoar:
`:` is bad too in my eyes, because you can use it to specify a port in your url. 
I'm not sure about the other characters.

plata:
I don't see the problem. If we split on the last `:`, everything should be fine. "git+https://github.com/PlayOnLinux/Scripts:master:8080" will not work anyways.

madoar:
No, but `http://github.com:8080/PlayOnLinux/Scripts would be valid`, if your could reach a server under this url, right?

plata:
It's maybe a valid URL but not a valid entry for our property anymore (would have to be "git+http://github.com:8080/PlayOnLinux/Scripts:master").

madoar:
I think it's real possibility, that such a git url will used. 
It would be quite useful if we find a character that is not found in an encoded git repository url

[...]

Issue #776, opened 2017-05-06
Use the file manager for picking a local repository for scripts


Ads20000:
This would be much more intuitive than knowing you have to put `file:///` before a directory and typing it in manually.

madoar:
The idea to use a file manager to pick a local repository is good, but the problem is how to identify which kind of repository the user wants to add?
If the user wants to add a git repository a file manager is useless to him, the same is true for the classpath repository. 
Therefore we would need to somehow identify what type of repository the user wants to add.

plata:
This requires a small wizard with two steps:
1. select repository type (combobox)
2. select repository path (file chooser or text box)

madoar:
@plata should we implement such a wizard? I'm not sure if we'll need additional work to style it correctly?

plata:
I'm also a bit concerned about the look and feel. But I don't see another possibility currently.

PR #877, opened 2017-06-06
Added DTO classes for the repository locations

madoar:
This PR targets the first part of #857 by replacing the repository location string through DTO classes (`RepositoryLocation`) which are then persisted as json.
This enables us to easily introduce new information fields for different repositories, like a branch name or security credentials for a git repository.
To better introduce this first part I've temporarily disabled the ability to add new repositories.

The repository list is now saved in a file called `repositories.json`, which contains for example:
```json
[{"type":"git","gitRepositoryUri":"https://github.com/PlayOnLinux/Scripts"},{"type":"git","gitRepositoryUri":"https://github.com/PlayOnLinux/Oldwares"},{"type":"classpath","packagePath":"/org/phoenicis/repository"}]
```

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
@@ -12,6 +12,8 @@ import javafx.scene.layout.*;
 import javafx.scene.text.Text;
 import org.phoenicis.repository.RepositoryManager;
 import org.phoenicis.javafx.views.common.TextWithStyle;
+import org.phoenicis.repository.dto.RepositoryLocation;
+import org.phoenicis.repository.repositoryTypes.Repository;
 import org.phoenicis.settings.SettingsManager;
 
 import java.util.Optional;
@@ -35,7 +37,7 @@ public class RepositoriesPanel extends StackPane {
 
     private Text repositoryText;
     private VBox repositoryLayout;
-    private ListView<String> repositoryListView;
+    private ListView<RepositoryLocation<? extends Repository>> repositoryListView;
     private HBox repositoryButtonLayout;
     private Button addButton;
     private Button removeButton;
@@ -49,7 +51,7 @@ public class RepositoriesPanel extends StackPane {
 
     private VBox overlay;
 
-    private ObservableList<String> repositories;
+    private ObservableList<RepositoryLocation<? extends Repository>> repositories;
 
     /**
      * Constructor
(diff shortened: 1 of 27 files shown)


madoar:
@qparis it seems like I've killed jenkins. If jenkins doesn't terminate the test itself, can you please do it?
My latest commit should fix the problem.

qparis:
ok to test

madoar:
@qparis the old build for efeb50d is still running and blocking the new build for 8fb0f58.
I believe that the old build won't terminate after a timeout.

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/LocalRepositoryLocation.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        LocalRepositoryLocation that = (LocalRepositoryLocation) o;
+
+        return new EqualsBuilder().append(repositoryLocation, that.repositoryLocation).isEquals();
+    }
+
+    @Override
+    public int hashCode() {
+        return new HashCodeBuilder().append(repositoryLocation).toHashCode();
```
qparis: You may want to  repositoryLocation.hashcode

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/LocalRepositoryLocation.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.io.File;
+
+/**
+ * Created by marc on 06.06.17.
+ */
+@JsonDeserialize(builder = LocalRepositoryLocation.Builder.class)
+public class LocalRepositoryLocation extends RepositoryLocation<LocalRepository> {
+    private final File repositoryLocation;
+
```
madoar: What would I need to change, so that the code works without the Builder class?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
```diff
@@ -130,19 +132,20 @@ private void populateRepositoryGrid() {
...
 
             Optional<String> result = dialog.showAndWait();
-            result.ifPresent(newRepository -> {
-                repositories.add(0, newRepository);
-
-                this.save();
-
-                repositoryManager.addRepositories(0, newRepository);
-            });
+            //            result.ifPresent(newRepository -> {
```
madoar: This is only commented out because I want to replace this code by a new GUI in one of my next PRs (#776)

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/LocalRepositoryLocation.java
qparis: Just use a simple constructor with @JsonCreator and @JsonProperty annotations

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
qparis: Then, maybe explain it with a comment :)

madoar:
Should I already add a `branch` field to the git repository in this PR?
This would allow us to prevent the need to delete the repository config file later, because of the later addition of the `branch` field.
Otherwise I would appreciate it, if we could merge this soon, because I want to start working on the new GUI for the addition of a new repository.

PR #881, opened 2017-06-09
enable script translation via "tr"

### LINKED ISSUE #875: merge TranslationDTOs
`TranslationDTO's` from multiple repositories must be merged. Per language: concatenate json content.

plata:
fixes #875

### DIFF
--- a/phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
+++ b/phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
@@ -39,6 +39,13 @@ public final class Localisation {
         return i18n;
     }
 
+    /**
+     * add a ResourceBundle containing additional translations for the localisation
+     * - typically used to add translations for the scripts
+     * - only one additional ResourceBundle can be set at a time (calling the method again will override the old ResourceBundle)
+     *
+     * @param resourceBundle
+     */
     public static void setAdditionalTranslations(ResourceBundle resourceBundle) {
         getI18n().setResources(resourceBundle);
     }
--- a/phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
+++ b/phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
@@ -21,15 +21,26 @@ package org.phoenicis.configuration.localisation;
 import org.xnap.commons.i18n.I18n;
 import org.xnap.commons.i18n.I18nFactory;
 
+import java.util.ResourceBundle;
+
 public final class Localisation {
+    private static I18n i18n = null;
+
     // This is a static class
     private Localisation() {
 
     }
 
     private static I18n getI18n() {
-        I18n getI18n = I18nFactory.getI18n(Localisation.class, "Messages");
-        return getI18n;
+        // init only once
+        if (i18n == null) {
+            i18n = I18nFactory.getI18n(Localisation.class, "Messages");
+        }
+        return i18n;
+    }
+
+    public static void setAdditionalTranslations(ResourceBundle resourceBundle) {
+        getI18n().setResources(resourceBundle);
     }
 
     public static String tr(String str) {
--- a/phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/PropertiesResourceBundle.java
+++ b/phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/PropertiesResourceBundle.java
@@ -1,0 +1,24 @@
+package org.phoenicis.configuration.localisation;
+
+import java.util.Enumeration;
+import java.util.Properties;
+import java.util.ResourceBundle;
+
+public class PropertiesResourceBundle extends ResourceBundle {
+    private Properties properties;
+
+    public PropertiesResourceBundle(Properties properties) {
+        this.properties = properties;
+    }
+
+    @Override
+    protected Object handleGetObject(String key) {
+        return this.properties != null ? properties.get(key) : this.parent.getObject(key);
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Enumeration<String> getKeys() {
+        return this.properties != null ? (Enumeration<String>) this.properties.propertyNames() : this.parent.getKeys();
+    }
+}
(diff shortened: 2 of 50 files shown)


phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/LocalRepository.java
```diff
@@ -71,20 +72,21 @@ public RepositoryDTO fetchInstallableApplications() {
...
+            for (File translationFile : translationFiles) {
                 try {
-                    final String language = translation.getName();
-                    final String content = new String(Files.readAllBytes(translation.toPath()));
-                    final TranslationDTO.Builder translationDTOBuilder = new TranslationDTO.Builder()
-                            .withLanguage(language).withJson(content);
-                    translations.add(translationDTOBuilder.build());
+                    Properties langProperties = new Properties();
+                    langProperties.load(new FileInputStream(translationFile));
+                    mergedProperties.putAll(langProperties);
```
madoar: How are overrides handled in this case?
Let's say a key is used in multiple property files, which one is then taken in the end?

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/NashornScriptInterpreter.java
```diff
@@ -32,7 +32,9 @@ public NashornScriptInterpreter(NashornEngineFactory nashornEngineFactory) {
 
     @Override
     public void runScript(String scriptContent, Consumer<Exception> errorCallback) {
-        nashornEngineFactory.createEngine().eval(scriptContent, errorCallback);
+        final String scriptContenti18n = "var tr = Packages.org.phoenicis.configuration.localisation.Localisation.tr;"
```
madoar: I think
```java
NashornEngine engine = nashornEngineFactory.createEngine();

engine.eval("var tr = Packages.org.phoenicis.configuration.localisation.Localisation.tr;", errorCallback);
engine.eval(scriptContent, errorCallback);
```
would be a better and more clear implementation.

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/PropertiesResourceBundle.java
```diff
@@ -0,0 +1,24 @@
+package org.phoenicis.configuration.localisation;
+
+import java.util.Enumeration;
+import java.util.Properties;
+import java.util.ResourceBundle;
+
+public class PropertiesResourceBundle extends ResourceBundle {
```
madoar: Just asking:
Why don't we define a `JsonResourceBundle`?

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
```diff
@@ -21,15 +21,26 @@
...
-        I18n getI18n = I18nFactory.getI18n(Localisation.class, "Messages");
-        return getI18n;
+        // init only once
+        if (i18n == null) {
+            i18n = I18nFactory.getI18n(Localisation.class, "Messages");
+        }
+        return i18n;
+    }
+
+    public static void setAdditionalTranslations(ResourceBundle resourceBundle) {
```
madoar: JavaDoc pls :)

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/LocalRepository.java
plata: > If the map previously contained a mapping for the key, the old value is replaced by the specified value.
According to: <https://stackoverflow.com/questions/17827573/map-putall-overrides-or-adds>

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/PropertiesResourceBundle.java
plata: Why should we?

madoar: This could be quite useful, when we want to provide translations inside the repositories right?

plata: I don't think so. Check PlayOnLinux/Scripts#258 for how it works.
To translate the json files, I am planning a script which extracts the relevant Strings.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/NashornScriptInterpreter.java
```diff
@@ -32,7 +32,9 @@ public NashornScriptInterpreter(NashornEngineFactory nashornEngineFactory) {
 
     @Override
     public void runScript(String scriptContent, Consumer<Exception> errorCallback) {
-        nashornEngineFactory.createEngine().eval(scriptContent, errorCallback);
+        final NashornEngine engine = nashornEngineFactory.createEngine();
```
qparis: This code should be executed in NashornEngineFactory

plata: Yep. Done. One question: why is the `NashornEngine` created for every call? Can't we just do this once?

qparis: To avoid global states that could lead to bugs

plata: Ok.

plata:
@qparis is this approach now the way we should go? If so, I would like to merge this and PlayOnLinux/Scripts#258 in order to continue with the other open issues (mainly json translation).

qparis:
It seems so. I need to test it more deeply but we can go ahead and push examples

madoar:
This seems good to me. I'll let @qparis review this, after he has finished the tests

qparis:
Let's try to merge it and see if something remains to be fixed

plata:
@qparis I need a positive review then.

PR #884, opened 2017-06-11
translate apps and categories

plata:


plata:
@madoar @qparis could one of you review this?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationSideBar.java
```diff
@@ -158,7 +158,7 @@ private ToggleButton createAllCategoriesToggleButton() {
     private ToggleButton createCategoryToggleButton(CategoryDTO category) {
         final LeftToggleButton categoryButton = new LeftToggleButton(category.getName());
 
-        categoryButton.setId(String.format("%sButton", category.getName().toLowerCase()));
+        categoryButton.setId(String.format("%sButton", category.getId().toLowerCase()));
```
madoar: What's the difference between the name and the id?

phoenicis-library/src/main/java/org/phoenicis/library/LibraryManager.java
```diff
@@ -61,19 +61,19 @@ public LibraryManager(String shortcutDirectory) {
...
                 ShortcutDTO shortcut = fetchShortcutDTO(shortcutDirectoryFile, file);
-                String categoryName = shortcut.getCategory();
-                if (!categoryMap.containsKey(categoryName)) {
-                    categoryMap.put(categoryName, new ArrayList<>());
+                String categoryId = shortcut.getCategory();
+                if (!categoryMap.containsKey(categoryId)) {
+                    categoryMap.put(categoryId, new ArrayList<>());
                 }
-                categoryMap.get(categoryName).add(shortcut);
+                categoryMap.get(categoryId).add(shortcut);
```
madoar: This looks like a simple grouping task, right?
I believe it would be better readable and shorter if you use the java streaming api for this. (see here: https://www.leveluplunch.com/java/examples/java-util-stream-groupingBy-example/)

phoenicis-repository/src/test/java/org/phoenicis/repository/repositoryTypes/CachedRepositoryTest.java
```diff
@@ -38,8 +38,8 @@ public void testFetchInstallableApplications() throws Exception {
                         .thenReturn(
                                 new RepositoryDTO.Builder()
                                         .withCategories(
-                                                Arrays.asList(new CategoryDTO.Builder().withName("Category 1").build(),
-                                                        new CategoryDTO.Builder().withName("Category 2").build()))
```
madoar: No names for the tests?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationSideBar.java
plata: `name` is the displayed name and might therefore be translated whereas `id` is a unique identifier (currently simply the English name).

phoenicis-library/src/main/java/org/phoenicis/library/LibraryManager.java
plata: Maybe. I didn't want to change the code at all, just fix it to work with translations.

phoenicis-repository/src/test/java/org/phoenicis/repository/repositoryTypes/CachedRepositoryTest.java
plata: It doesn't matter. If the `name` is not given, it will fall back to the `id`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationSideBar.java
madoar: If this is the case, I think you should watch out when you use ids with ` ` spaces in the tests. 
I think you should ensure here that the id contains no space characters or other "bad" symbols, right?

plata: It's really not important. The thing is that I cannot use the `name` anymore because it can be different depending on the language. Apart from that there are no restrictions.

phoenicis-repository/src/test/java/org/phoenicis/repository/repositoryTypes/CachedRepositoryTest.java
qparis: In fact we can still use a name and it is easier. It is just that it needs to be translated when it is used in a label

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationSideBar.java
madoar: The problem I see here is with category ids such as `arcade games`.
The created button id string for such category ids would look like this: `#arcade gamesButton`.
This id string would be invalid, because it contains a space character and is therefore interpreted as a tag with the id `arcade`, which contains another element `gamesButton`.

plata: That's true! Didn't think about this case. However, this is also the case currently so I would prefer to fix this separately.

phoenicis-repository/src/test/java/org/phoenicis/repository/repositoryTypes/CachedRepositoryTest.java
plata: That was also my first approach. However, it's not that simple. For example, you also have to consider every place where apps/categories are sorted and the search. You could not write code like the following any more:
```Java
Comparator.comparing(CategoryDTO::getName)
```
So you would have to be very careful when implementing something that you translate everything in the right places.
Therefore I decided that it's easier/cleaner to have the translated name in the DTO already.

In general I believe it's a good idea to identify apps/categories by an ID and not the name. That's also what Steam etc. does.

Additionally, we might have issues with names if they contain whitespace or special characters as @madoar pointed out above.

qparis: So we could have two kinds of DTOs (translated and untranslated) at the end.
I suggest we have a Translatable interface then

qparis: And add Translatable tr(Translatable) method to Localisatiob

plata: What's the point of having untranslated DTOs? I mean why would you want something NOT to be translated?

plata: Also, how should it be used? E.g. how would 
```Java
Comparator.comparing(CategoryDTO::getName)
```
look like with your approach?

qparis: When you get the data from the webservice, you will have untranslated DTOs
The idea here is just to rationalize the way we translate our DTOs and avoid confusion between what comes from the source and what is displayed to the end user

plata: I think I got your point. However, I'm still unsure how this should be implemented. A new class `TranslatableApplicationDTO` which is derived from `ApplicationDTO` and implements `Translatable`? And then during the fetch create first an `ApplicationDTO` and then a `TranslatableApplicationDTO` from it and call `tr(TranslatableApplicationDTO)`?

qparis: Yes for the process. However, the data structure of translated and untranslated is the same so I guess we can keep the same class

plata: Ok. I'll try it.

plata: Is it ok that `name` and `description` of the DTO are not `final` anymore?

plata:
There's still a problem with the shortcuts. As the translations are stored in the repository, shortcuts can only be translated if the repository is available. So e.g. if you install an app and then remove the repository, the shortcut cannot be translated anymore.

phoenicis-repository/src/test/java/org/phoenicis/repository/repositoryTypes/CachedRepositoryTest.java
qparis: No it's not. These fields should really be immutable

madoar:
Why don't we use the translated string to create the shortcut?
Afterwards we only save the translated strings and not the "original" ones.
I don't see a benefit for the shortcuts to also save the other translation file for other languages.

plata:
Then you can never change your language.

madoar:
No you can't change the language of a created shortcut.
In addition: Is it easily possible to change the label of a system shortcut?

plata:
It's done like this:
```
Name=LibreOffice Calc
GenericName=Spreadsheet
GenericName[de]=Tabellendokument
```

phoenicis-repository/src/test/java/org/phoenicis/repository/repositoryTypes/CachedRepositoryTest.java
plata: Then I cannot translate it later.

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
```diff
@@ -74,4 +74,8 @@ public static String tr(String text, Object[] objects) {
         return getI18n().tr(text, objects);
     }
 
+    public static void tr(Translatable translatable) {
```
madoar: JavaDoc please

phoenicis-library/src/main/java/org/phoenicis/library/dto/ShortcutCategoryDTO.java
```diff
@@ -79,19 +88,27 @@ public boolean equals(Object o) {
...
+        this.description = tr(this.description);
     }
 
     @JsonPOJOBuilder(buildMethodName = "build", withPrefix = "with")
     public static class Builder {
-        private String name;
-        private String description;
+        private String id = "";
+        private String name = "";
+        private String description = "";
```
madoar: I'm not sure how we want to handle this, but in my eyes I would do all initial assignments in the constructor.

phoenicis-repository/src/test/java/org/phoenicis/repository/repositoryTypes/CachedRepositoryTest.java
qparis: Yes, you just have to clone the DTO

qparis: This is exactly the purpose of this pattern: to force you to clone the DTO to translate it

phoenicis-library/src/main/java/org/phoenicis/library/dto/ShortcutCategoryDTO.java
plata: There are two constructors for the `Builder` so I thing it's safer to initialize here (it's also done like this for the `shortcuts` already).

phoenicis-repository/src/test/java/org/phoenicis/repository/repositoryTypes/CachedRepositoryTest.java
plata: Ok like this?

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
```diff
@@ -74,4 +74,13 @@ public static String tr(String text, Object[] objects) {
         return getI18n().tr(text, objects);
     }
 
+    /**
+     * translate Translatable object
+     * @param translatable Translatable object
+     * @return translated object
+     */
+    public static Translatable tr(Translatable translatable) {
```
madoar: I'm still missing in the javadoc, what "translating" means in this case.
I would suggest to add something like:

> Translates the given Translatable object, by creating a new instance of it with its corresponding translations

madoar:
I think this PR can be merged, but I'll let @qparis decide

phoenicis-library/src/main/java/org/phoenicis/library/dto/ShortcutCategoryDTO.java
qparis: I don't think DTO fields should be empty by default : DTOs are agnostics, there are no reason to disallow null values. (You could have a DTO without description at all)

plata: Yes, but then we might have to check when we're using it. Not sure that's better.

plata: E.g. I would have to check if it's `null` before calling `tr` to translate it.

qparis: Yes, this is required anyway because one could call the builder with null arguments.
You can also take this into account directly in tr(String) method

plata: You're right. I fixed it.

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
```diff
@@ -50,28 +50,37 @@ public static void setAdditionalTranslations(ResourceBundle resourceBundle) {
...
-        return getI18n().tr(text, objects);
+        return text == null ? "" : getI18n().tr(text, objects);
+    }
+
+    /**
+     * Translates the given Translatable object, by creating a new instance of it with its corresponding translations
+     * @param translatable Translatable object
+     * @return translated object
+     */
+    public static Translatable tr(Translatable translatable) {
```
madoar: I think it would be better to provide a template method and generic interface for `Translatable` here:
```java
public static <T> Translatable<T> tr(Translatable<T> translatable) {
   return translatable.translate();
}
```
This would remove the need to cast the resulting element later on.

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
```diff
@@ -50,28 +50,37 @@ public static void setAdditionalTranslations(ResourceBundle resourceBundle) {
...
     }
 
     public static String tr(String text, Object o1, Object o2) {
-        return getI18n().tr(text, o1, o2);
+        return text == null ? "" : getI18n().tr(text, o1, o2);
     }
 
     public static String tr(String text, Object o1, Object o2, Object o3) {
-        return getI18n().tr(text, o1, o2, o3);
+        return text == null ? "" : getI18n().tr(text, o1, o2, o3);
```
qparis: I think the correct translation for "null" is null, not a blank string

madoar: If you use "null" as a translation for the not existing value `null` in the gui you will have "null"s in your gui components. I'm not sure we want that.

madoar: As an example: Let's say the description of an `ApplicationDTO` is `null`, then the description in the AppsPanel would contain the text "null".

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
qparis: Better: <T extends Translatable> T tr(Translatable translatable, Class<T> clazz

madoar: Why?

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
qparis: No, the GUI needs to check if the description is null or empty, this is really different.

Another point is that tr() should not change any behavior (I.e, the app needs to have the same behavior if we remove tr)

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
qparis: Because the tr function will directly return the right DTO and not a Translatable, so we can access all getters

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
madoar: Oh, I think I misunderstood you. I thought you wanted to return the string text value "null" but you meant the null value null.

qparis: Yeah exactly :) Let's not change the data structure with tr

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
madoar: But we still wouldn't require the passing of a `Class<T>` instance, right?

plata: And then
```Java
 RepositoryDTO implements Translatable<RepositoryDTO>
```?
I don't know but somehow this looks a bit strange for me.

Comment (qparis) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: Exactly like Hibernate or Jackson do in fact

CategoryDTO translated = tr(notTranslated, CategoryDTO.class)

Comment (qparis) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: Translatable do not need to be generic

Comment (qparis) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: Yep, this is also good

Comment (madoar) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: But why not use generics?
They add type safety to the code, why is useful here in my eyes.

Comment (qparis) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: Because it is easier for the developer once the translation package is written isn't it?

Comment (madoar) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: I'm not sure I can follow you, what is easier once the translation package is written?

Comment (qparis) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: To translate a DTO

I mean, what we are proposing here is typesafe isn't it?

Comment (plata) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: @madoar if I do that, I get `type argument T is not within bounds of type-variable E` for the `T` in `Translatable<T> translatable`.

Comment (madoar) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: The problem with passing a Class object in my eyes is, that it doesn't really tell us how we can generate a new object of its type.
In addition you can do the functionality through the use of generics, in this use case.
This needs shorter code, therefore I would refrain from passing `Class<T>` instances around.
I normally only pass `Class<T>` instances around in a JPA environment, where you can use them to automatically provide select all queries for every DAO type.

Comment (madoar) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: For the method itself, the generic interface or for the calling code?

Comment (plata) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: For the parameter in the method itself. I guess it expects that `T` extends `Translatable` or something like that.

Comment (qparis) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: We don't need to create a new instance because  we already have Translatable.translate which does it for us. 
However, we could also take inspiration from DTOContractTest and automatically translate any DTO without any code.
Up to you

Comment (plata) [inline phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java]: Yes. I have to:
```Java
public static <T extends  Translatable> T tr(Translatable<T> translatable) {
```

qparis: So yeah, maybe the Class<T> is not needed at the end

qparis: Maybe <T> T tr(Translatable<T> translatable) is sufficient enough

qparis: @plata Why?

madoar: Ok, then please try:
```java
public static <T extends Translatable<?>> T tr(Translatable<T> translatable) {
   return translatable.translate();
}
```

plata: But of course now the whole point of doing this is gone because I have to cast again.

qparis: Why would you have to cast again?

madoar: Another way to solve his would be to remove the `extends Translatable` from the `Translatable` class.

madoar: recursive references are tricky....

plata: `incompatible types: org.phoenicis.configuration.localisation.Translatable cannot be converted to org.phoenicis.repository.dto.RepositoryDTO`

qparis: Why not using <T extends Translable<T>>? Would it work?

qparis: @plata can you show the line that shows this?

qparis: Shall we continue this discussion on slack?

plata: It's `tr(repositoryDTO)` in `DefaultRepositoryManager#triggerRepositoryChange`.

The Problem is that the `T` is not resolved properly. I can do: `Localisation.<RepositoryDTO>tr()` but do we want that?

madoar: Let me give this a short try..

plata: @qparis let's continue here this time so we can look up the discussion later if needed.

madoar: Got it...
I think this should work:

```java
    /**
     * Translates the given Translatable object, by creating a new instance of it with its corresponding translations
     * @param translatable Translatable object
     * @return translated object
     */
    public static <T extends Translatable<?>> T tr(Translatable<T> translatable) {
        return translatable.translate();
    }
```

```java
public interface Translatable<E extends Translatable<?>> {

    /**
     * translate the object
     */
    E translate();
}
```

```java
public class RepositoryDTO implements Translatable<RepositoryDTO> {
   ...
}
```

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Translatable.java
```diff
@@ -0,0 +1,9 @@
+package org.phoenicis.configuration.localisation;
+
+public interface Translatable {
```
madoar: JavaDoc please :)

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
plata: Yep. But it's a brain twister really.

madoar:
Yes , but I think the generics support from IntelliJ is not the best, because it often shows no errors when a generic type is required.
I think Eclipse shows the error earlier, because it tells you when a generic type is missing.

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/CategoryDTO.java
```diff
@@ -92,9 +102,19 @@ public String getName() {
         return (o1, o2) -> o1.getName().compareToIgnoreCase(o2.getName());
     }
 
+    @Override
+    public CategoryDTO translate() {
+        List<ApplicationDTO> applicationDTOS = new ArrayList();
```
madoar: I think the s should be lower case here.

plata: We have it like this in several places (IntelliJ IDEA creates it like this). Unless we introduce a rule and have it the same way everywhere, I don't see a reason to change it here.

plata:
@qparis now it's up to you ;)

qparis:
I think we can also automate the translation of any DTOs in fact

qparis:
i.e, avoid having the need override Translate()

madoar:
This sounds like something for which we would need reflection.
I'm not sure we want to do that.

qparis:
why not?

madoar:
When you write reflection code it's adjusted to fit to the classes you want to analyse.
In my opinion this works as long as you have a good and detailed definition of classes to be analysed, but once you change something you will get strange effects and errors, because your compiler can't give you warnings and errors at compile time anymore.
We already have this problem with the `ContainerPanelFactory` where you get no compile time errors and warnings when you change something at the constructor of the created ContainerPanel.

qparis:
I agree with you but we have DTOContractTest that really strictly frame the way DTOs are used so it's not really a concern for me

plata:
Shouldn't we merge this PR such that translation works at all and discuss these refinements later on?

qparis:
Yep

PR #885, opened 2017-06-11
Close a Git in all situations

gverm:
fixes: http://www.playonlinux.org:9000/issues#issues=AVw-tQLX9U_f07q6H1BH


qparis:
ok to test

PR #886, opened 2017-06-12
Add a gui dialog for the addition of repositories

### LINKED ISSUE #776: Use the file manager for picking a local repository for scripts

madoar:
This PR fixes #776.
My problem with the taken approach is the used library `controlsfx`.
This library provides a `Wizard` component, which I'm using to divide the repository addition process in two parts:

1. The selection of the repository type (git, classpath, local)
2. The input phase of repository type dependent information (git url, classpath, file path)

This wizard restricts us in some aspects. One such aspect is the used css styles. The only way we can bind a stylesheet to the wizard seems to be to bind it to all the wizard steps. The same is true for the definition of a size for the wizard dialog.
My last and greatest problem is the provided translation from `controlsfx`. This translation has some "minor" bugs, like a misspelling of `Nächster`. In addition the chosen translations may not be the best, like someone wrote in the issue system of `controlsfx`. 

grafik
grafik
grafik
grafik

As a short note:
The git branch is currently not used, I've just added it because I want to provide its behavior in my next PR.

### DIFF
--- a/phoenicis-javafx/pom.xml
+++ b/phoenicis-javafx/pom.xml
@@ -176,6 +176,11 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.controlsfx</groupId>
+            <artifactId>controlsfx</artifactId>
+            <version>8.40.12</version>
+        </dependency>
     </dependencies>
 
 </project>
--- a/phoenicis-javafx/pom.xml
+++ b/phoenicis-javafx/pom.xml
@@ -175,12 +175,6 @@
             <artifactId>phoenicis-settings</artifactId>
             <version>${project.version}</version>
         </dependency>
-
-        <dependency>
-            <groupId>org.controlsfx</groupId>
-            <artifactId>controlsfx</artifactId>
-            <version>8.40.12</version>
-        </dependency>
     </dependencies>
 
 </project>
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/AddRepositoryDialog.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/AddRepositoryDialog.java
@@ -7,8 +7,6 @@ import javafx.scene.layout.GridPane;
 import javafx.scene.layout.HBox;
 import javafx.scene.layout.Priority;
 import javafx.stage.DirectoryChooser;
-import org.controlsfx.dialog.Wizard;
-import org.controlsfx.dialog.WizardPane;
 import org.phoenicis.repository.location.ClasspathRepositoryLocation;
 import org.phoenicis.repository.location.GitRepositoryLocation;
 import org.phoenicis.repository.location.LocalRepositoryLocation;
@@ -20,6 +18,7 @@ import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.Optional;
+import java.util.function.Supplier;
 
 import static org.phoenicis.configuration.localisation.Localisation.tr;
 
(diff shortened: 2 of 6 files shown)


plata:
With the problems you mentioned, I'm not so sure I like `controlsfx`. I'm not even sure we need a multi step wizard here. Why not have just one pop-up which first contains only the type selection combobox and shows the other fields below depending on the selection?

madoar:
Isn't the behavior you've described a wizard?

I first tried to implement the functionality using a `Dialog`. The problem was, that I couldn't get the "Continue" button to work correctly.
I've initialized, the "Continue" button with `ButtonData.NEXT_FORWARD`, but whenever you've clicked on it, the wizard automatically closed itself. 
Therefore I've used a `Wizard`.

qparis:
The thing is that we are too much depending on the UI.
With this implementation we cannot really manage repositories in CLI mode, which is possible with a SetupWizard

madoar:
What do you mean with SetupWizard?

qparis:
Wizard *

madoar:
@qparis I still don't understand your last argument.
To support different arguments for different types of repositories, we need a two stages approach:

1. Selection of the repository type to be added
2. Input of specific information related to the previously chosen repository type (like the branch name for the git repository)

This kind of behavior can be visualized best by a wizard. 
For the CLI tool I believe we'll need to add 3 commands, one for each type of repositories we have,
which the user needs to use to add a repository. This should be independent from the javafx gui.

plata:
I guess @qparis is talking about the current way we handle wizards and implement them in JavaFX and CLI. As an example, you can look at the `SetupUiJavaFXImplementation` and `SetupUiCliImplementation` which implement the `SetupUi` interface. Probably @qparis would like a similar approach for the repository selection. In this case, we should have a method `addRepository` in `RepositoryManager` which somehow uses the different implementation depending on how the wizar is wired up (@qparis or do you want to do this from JS?).

qparis:
You are correct, I think that we are putting to much process (the workflows when you add or edit a repository) in the JavaFX stack.

I understand that a Wiaard is not adapted because it requires to open a new tab an etc...

However, I believe that we can extend Wizard to add some "popup" logic. 

Concerning the language, I think we can use Java without any concern

plata:
Actually the wizard does not have to open a new tab. We only do this for the setup. E.g. the engine installation works differently.

qparis:
You are correct.

madoar:
Just to summarize for my own understanding:
You suggest that we add a new interface for the management of the addition of a new repository.
Afterwards we implement two subclasses, one for the javafx application and another for the cli?

In my eyes this seems like shooting with cannons on sparrows. I believe it would be better to simply add a new command in the cli for the addition of a new repository. The reason for this is, because the core part of the repository addition are only four to five commands:
```java
LocalRepositoryLocation location = new LocalRepositoryLocation(new File(pathField.getText()));

List<RepositoryLocation<? extends Repository>> repositories = settingsManager.loadRepositoryLocations();
repositories.add(location);

settingsManager.saveRepositories(repositories);
repositoryManager.addRepositories(0, repositoryLocation);
```

This is not an amount of code, for which I think we need to write a common interface, which fits both the cli and javafx application.

qparis:
Fair enough

madoar:
Ok, here the current look:
grafik
grafik
grafik
grafik

plata:
Can you align "Repository type" vertically with the combobox? I think some spacing between the label and the control would look better.

madoar:
I didn't really think about the design of the dialog content in this PR.
I would suggest, to design the new dialog in another future PR.
Currently I think it's more important to add a new dialog to enable the users again to add new repositories to Phoenicis.

plata:
Ok. If it works, that's fine for me.

madoar:
For me it does, I suggest someone else tests it locally too.
I think it's enough to simply readd currently added repositories

plata:
Then let's just go with it. We can improve anything later.

PR #893, opened 2017-06-17
restrict DTO IDs to letters and numbers

plata:
To avoid problems with IDs such as `arcade games` which would create the invalid button ID string `#arcade gamesButton` (would be interpreted as a tag with the ID `arcade`, which contains another element `gamesButton`).

### DIFF
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/LocalRepository.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/LocalRepository.java
@@ -101,8 +101,15 @@ public class LocalRepository implements Repository {
 
                 if (categoryJson.exists()) {
                     final CategoryDTO.Builder categoryDTOBuilder = new CategoryDTO.Builder(
-                            unSerializeCategory(categoryJson)).withId(categoryDirectory.getName())
-                                    .withApplications(fetchApplications(categoryDirectory));
+                            unSerializeCategory(categoryJson)).withApplications(fetchApplications(categoryDirectory));
+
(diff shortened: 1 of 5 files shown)


phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/LocalRepository.java
```diff
@@ -101,8 +101,15 @@ public RepositoryDTO fetchInstallableApplications() {
 
                 if (categoryJson.exists()) {
                     final CategoryDTO.Builder categoryDTOBuilder = new CategoryDTO.Builder(
-                            unSerializeCategory(categoryJson)).withId(categoryDirectory.getName())
-                                    .withApplications(fetchApplications(categoryDirectory));
+                            unSerializeCategory(categoryJson)).withApplications(fetchApplications(categoryDirectory));
+
+                    if (StringUtils.isBlank(categoryDTOBuilder.getId())) {
```
madoar: What would happen if `!StringUtils.isBlank(categoryDTOBuilder.getId())` is true and `categoryDTOBuilder.getId()` contains a space character?

plata: That would mean that
1. the category.json contains an ID (which is not the case currently)
2. this ID is invalid

But yes, in that case we might get into trouble.

madoar: Should we supply code for this case?

plata: It's unlikely to happen but I secured it now.

PR #897, opened 2017-06-17
list chooser: prevent unselecting all buttons

### LINKED ISSUE #853: Prevent deselection of selected list widget
Currently it's possible to show the currently selected list widget as "deselected" by clicking on it when it's selected. We should prevent this.
grafik
I would suggest to use the same approach we used to prevent the user from deselecting the currently selected engine/application category.

plata:
fixes #853

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftListWidgetChooser.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftListWidgetChooser.java
@@ -1,5 +1,6 @@
 package org.phoenicis.javafx.views.mainwindow.ui;
 
+import javafx.application.Platform;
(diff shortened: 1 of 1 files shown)


madoar:
I don't like this solution.
Basically this does trigger a second click to restore a correct state.
I would prefer a solution like I used for the category buttons in the `LeftToggleGroup` (see class `LeftToggleButton` for details).
This solution prevents the click from occuring if the clicked button is currently selected.

plata:
So we would need a new class just for this? Actually, the behavior is the same as a radio button. We could use a radio button with the style of a toggle button.

madoar:
No, you can also set also use an event filter which you bind to each toggle button.
This filter consumes the event if the clicked button is currently selected.
I just thought that a custom class for this is better.

plata:
Like this?

madoar:
Yes, this looks good to me.

PR #899, opened 2017-06-17
Automatic translation of any object

qparis:


qparis:
This commit contains : 
- A replacement of Translatable interface. Instead, you just have to use @Translatable annotation in any class you want to translate. (See DTOs to get examples) 
- Some unit tests

plata:
This needs some JavaDoc to explain how everything works together.

qparis:
Done

plata:
Let's see what @madoar thinks about it. He was a bit skeptical earlier on.

qparis:
I think it makes Localisation class harder to read, but if you isolate and disregard this, it becomes really easy and clear to read translatable items: 

capture d ecran 2017-06-17 a 19 35 56

madoar:
I think we should do this by restricting the developer to use field access to set the translated values.
When doing this we have three requirements for the programmer of a `Translatable` class:

- annotate the class with a `@Translatable` annotation
- supply an empty default constructor
- annotate every field that should be translated with a `@Translate` annotation

When we do this we reduce the number of annotations we need from currently `@ParameterName`, `@Translatable`, `@TranslatableBuilder`, `@TranslatableCreator` and `@Translate` to just `@Translatable` and `@Translate`.

One possible problem that may appear with this solution is, that a developer uses a class, that can't be "translated" by the `public static <T> T tr(T translatable)` method. In this case we wouldn't get an error but possibly strange behavior.
This is what I've meant with a strict or good definition. The better we specify how a developer should define what should be translatable, the less problems can occur later.
Lastly the code needs more JavaDoc, for example in `LocalisationHelper` and `Localisation`.

qparis:
It is really hard to reduce the number of required annotations because the fields are often final private, so the framework has to use the constructor to set them.

Concerning the behavior, if and unexpected object goes through tr(), Localisation shall return the original one

madoar:
`final` fields seem to be no problem for reflection.
I've just googled a bit and found https://stackoverflow.com/questions/4516381/changing-private-final-fields-via-reflection and https://stackoverflow.com/questions/3301635/change-private-static-final-field-using-java-reflection.
Maybe we can use this approach?
The less annotations we have the less annotation dependent code we need to write.

qparis:
Hum so far I have used reflection but have never changed the accessibility of fields. Are you sure we want to do this?

madoar:
Why not, we just need to use the annotations in a single application and as long as it makes our work easier and works I think it's a good choice.
Like you've previously said it's a hidden part of the annotation processing code.
Another solution would be to don't use the final modifier. I'm not sure why we need it, yes it tells us more clearly that the code should be non modifiable, but we can also do this by not suppling setter methods, right?

qparis:
If we directly edit the fields, it means that the getters are no longer needed at all. We could have public final fields in all our objectif.

Suppose that you want to keep the same contract but change the way the data is stored internally, it will no longer be possible

madoar:
We aren't really directly editing the field.
Reflection should be used as an exception to the rule. 
Therefore we'll still require the getter methods to query the information inside the fields.

What I'm proposing, through the removal of the `final` modifier, is more or less the normal behavior, which libraries like hibernate, jersey and so on provide.

qparis:
Concerning Hibernate it makes a lot of sense because a database row cannot be immutable.
I really think that we do need final modifiers as it is really good defensive practice

qparis:
An my point was more about editing a private field anyway.

madoar:
I think editing `private` fields with reflection is more or less ok, I would even say standard, but with editing `final` fields, I would be a bit more careful.
Like I said a lot of libraries already support the editing of `private` fields, I think jackson also supports this, right?

qparis:
I don't know if it supports it because Jackson supports a lot of things, but in the way we are currently using it, we do not access private fields.

Another argument to keep private fields for example in ScriptDTO: imagine that you have a thread reading a ScriptDTO, checking the signature and executing the script. A developer could edit the script through RepositoryDTO in the same time, which is not acceptable.

madoar:
I think you misunderstood me. I'm not proposing to make the `private` field `public`.
I'm just proposing to circumvent this when creating a translated object.
This should be absolutely harmless, because the object can only be used after it is returned from the `tr` method, which only happens after the translation process has been completed and the state is "stable" again.

qparis:
I understand but I can't find a way to do it safely without removing final modifiers.
We could additionnany support translation of mutable objects or add the support of Jackson notations for already annotated class though

qparis:
The final field guarantee you that the objects you are manipulating have not be modified and prevents you from modifying them statically. 

If you just don't put setter and use a default constructor 
* nothing forces the user of the class to ask himself what is the value of each of our fields
* nothing prevents him from adding the setter he needs 

I really don't like the philosophy of weakening already existing modules to help the introduction of a new one

qparis:
To put in a nutshell, if a field is not final, the developer is allowed to change the state of the object

madoar:
Yes, that's true.
Can you please add some more javadoc to the remaining new methods in this PR?
This makes it easier to understand what each method should do

qparis:
Yeah I will improve this overall to make it clearer

qparis:
I really don't know why my pull request are polluted by other's code

plata:
It's because you're on your master. Don't ask me why but I also had this. If you create a branch, you will not get this behavior.

madoar:
Try a rebase as described in https://stackoverflow.com/questions/6102297/how-do-i-clean-up-my-github-fork-so-i-can-make-clean-pull-requests

qparis:
My fork is clean as I deleted and recreated it yesterday. 
I will try to create a new branch next time

madoar:
It's not about being clean, it has another commit history because you've just merged 2 PRs in your master branch.
You need to clear this up, so that github will show your master branch correct again.

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
```diff
@@ -76,11 +85,153 @@ public static String tr(String text, Object[] objects) {
...
+        for (Annotation annotation : clazz.getAnnotations()) {
+            if (annotation.annotationType() == Translatable.class) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    private static List<?> trList(List<?> translatable) {
```
madoar: Wouldn't it be better to do this as a generic method?

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
```diff
@@ -76,11 +85,153 @@ public static String tr(String text, Object[] objects) {
...
+        return translatable.stream().map(Localisation::tr).collect(Collectors.toList());
+    }
+
+    private static Set<?> trSet(Set<?> translatable) {
+        return translatable.stream().map(Localisation::tr).collect(Collectors.toSet());
+    }
+
+    private static <T> T trFromTranslateCreator(T translatable, Constructor<?> translateCreator)
```
madoar: I think it should be `Constructor<T>` right?

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/ParameterName.java
```diff
@@ -0,0 +1,12 @@
+package org.phoenicis.configuration.localisation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.PARAMETER)
+public @interface ParameterName {
```
madoar: JavaDoc?

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
qparis: It is, it is just that I'm a not using the parameter in the method

madoar: That's what I mean.
Isn't this the method you're providing:
```java
private static <T> List<T> trList(List<T> translatable) {
   ...
}
```

phoenicis-configuration/src/main/java/org/phoenicis/configuration/localisation/Localisation.java
qparis: It won't compile

madoar: This is really strange because:
```java
T result = (T) translateCreator.newInstance(parameters.toArray());
```
shows, that the given constructor should be of type `Constructor<T>` or `Constructor<? extends T>`, right?

qparis: Yep, but this is where the constructor is instanciated earlier that we have problems

madoar: Where is it instantiated?

qparis: final Class<?> clazz = translatable.getClass();
        final Optional<Constructor<?>> translateCreator = findTranslateCreator(clazz);
        final Optional<Class<?>> translateBuilder = findTranslateBuilder(clazz);

madoar: Yes you'll likely need to replace all `<?>` by fitting generic type names, then it should work.

qparis: It will stucks at Class<T> clazz = translatable.getClass

madoar: No, take a look at https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html.
It seems like the return type of `translatable.getClass()` is `Class<? extends T>`, right?

qparis: Yeah but I have tried and it does not work

madoar: When using `? extends T` you need to constantly use it everywhere, that builds onto it.
This for example means, that the second line should return an `Optional<Constructor<? extends T>>`.

qparis:
Can we merge it?

madoar:
I would still prefer to see more types, instead of `?`, in the generics, but I think we can merge this for now and add the generic types later on.

PR #903, opened 2017-06-18
create TabPane in ViewLibrary only once

plata:
Fixes issue that tab of the installation wizard is closed before the installation is finished (see PlayOnLinux/Scripts#260).

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/ViewLibrary.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/ViewLibrary.java
@@ -47,7 +47,7 @@ public class ViewLibrary extends MainWindowView<LibrarySideBar> {
 
     private LibrarySideBar sideBar;
     private LibraryPanel libraryPanel;
-    private final Tab installedApplication;
+    private final Tab installedApplicationsTab;
 
     private CombinedListWidget<ShortcutDTO> availableShortcuts;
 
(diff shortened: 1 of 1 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/ViewLibrary.java
```diff
@@ -47,6 +47,7 @@
 
     private LibrarySideBar sideBar;
     private LibraryPanel libraryPanel;
+    private final Tab installedApplication;
```
madoar: `installedApplicationsTab` right?

plata: yes

plata:
Maybe we could bind the content instead of setting it when something changes?

madoar:
You mean instead of `this.installedApplicationsTab.setContent(availableShortcuts);`?
I'm not sure why this line is necessary currently, because we're already using a binding to set the content of `availableShortcuts`.

plata:
I don't know. Maybe it isn't. Just fixed the error without really thinking about it.

madoar:
Can you try to remove the row and see if the behavior is still the same?

plata:
It's not working (nothing shown at all).

madoar:
Hm strange.
Maybe we should open an issue for this (showing the `availableShortcuts` through the use of a binding).

PR #906, opened 2017-06-19
Improve add repository dialog

madoar:
This PR improves the implementation of our dialog for the addition of a new repository.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/addrepository/RepositoryType.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    },
+
+    /**
+     * {@link org.phoenicis.repository.location.ClasspathRepositoryLocation} repository type
+     */
+    CLASSPATH {
+        @Override
+        public String toString() {
+            return tr("Classpath repository");
```
qparis: I think that toString should never be translated

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/addrepository/RepositoryType.java
```diff
@@ -0,0 +1,73 @@
...
+            return new ClasspathRepositoryDetailsPanel();
+        }
+    };
+
+    /**
+     * Returns the translated name of the repository type
+     *
+     * @return The translated name of the repository type
+     */
+    public abstract String toString();
```
qparis: I think you should choose another method name. It is confusing

madoar: The "problem" here is, that these enums are directly shown inside the `ComboBox`.
The ComboBox automatically uses the result of the `toString` method as the label for the object, therefore if we want to use another method name we would need to write a converter for it, do we want that?

qparis: Then, you could define a separate abstract method and use it in the definition of toString in the parent class. 
I'm a bit surprised that ComboBox does not offer any way of not using toString thought

madoar:
Is it better like this?

Issue #901, opened 2017-06-17
Playonlinux can't connect with repositories.


pyroxar:
Playonlinux can't connect with repositories.
When I click on apps or engines tab, I see load indicator for a long time but nothing happens.
![aa](https://user-images.githubusercontent.com/23406555/27255596-0f9d1c2c-53a1-11e7-838c-c10e4e38fdc1.png)

madoar:
Does your terminal/console contain any stacktrace/error?
If yes can you please post it for us? :)

plata:
This might be fixed by #893 (it fixes also a bug where the ID can be null).

plata:
@pyroxar can you try again?

pyroxar:
This problem resolved.
![zrzut ekranu z 2017-06-19 13-59-28](https://user-images.githubusercontent.com/23406555/27284112-ad668a0e-54f7-11e7-89ad-90ac86e25c6e.png)

pyroxar:
I was wrong. When I compiled, all work good at first glance ( only first run) but then Playonlinux not display games or engines (next starts).

plata:
How does your repositories.json look like (in ~/.Phoenicis)?

madoar:
Ok, this is interesting.
Currently I seem to have the same phenomena.
My current state is, that I don't have a `repositories.json` file in `~/.Phoenicis` and the "Apps" tab only shows the spinning loading circle.
My log contains the following content:

```
[...]

PR #907, opened 2017-06-20
Improve the IconsListWidget and IconsListElement

### LINKED ISSUE #810: Improve the IconsListWidget
Currently it seems to me that each `IconsListElement` inside a `IconsListWidget` looks kind squeezed.
Therefore I would suggest to make the `IconsListElement` a little larger and enable the title at the bottom of each `IconsListElement` to contain up to two lines of text. This makes it possible to also show slightly longer titles correctly (and fully).

madoar:
This PR fixes #810.
Because of issue #901 I currently can't test if a long title will be displayed in 2 lines.
Because I did a little cleanup in the css/less it would be nice if you could take an extra careful look on the changes in the main.less files of all themes.

The engines tab looks now like this:
grafik
grafik
grafik
grafik


plata:
Let's wait until you can really check it (2 lines).

qparis:
Can we see the list of games?

madoar:
Give me a moment.

madoar:
So, the application tab now looks like this:
grafik
grafik
grafik
grafik

madoar:
This is strange. If you set the scaling to more than ~50% the label won't be wrapped...
I'm not sure why this is the case, because every size I've used is in `em`, which should scale just fine.
A similar effect can be seen with the miniature images. If you set the scaling to less than 50% (use a really low one for a more clear effect), you'll discover, that the images don't fill the whole space that is available to them.

plata:
Is this behavior something new or did you only discover it during this PR? If last is true, I would take this to master and open an issue for it.

madoar:
I discovered this during this PR.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/widgets/lists/icons/IconsListElement.java
```diff
@@ -80,13 +81,14 @@ public IconsListElement(E item, URI miniatureUri, String title, boolean enabled)
             this.setClip(clip);
         });
 
-        final Label label = new Label(title);
-        label.getStyleClass().add("miniatureText");
+        final Region miniature = new Region();
```
plata: As you moved this out of `StaticMiniature` (why?), do we need the `StaticMiniature` class at all?

madoar: The class `StaticMiniature` was only a Region containing an image and a css class.
I moved this functionality to the `IconsListElement` class, because it was only 2 lines of code and because this is in lieu to the `CompactListElement` implementation.

The class `StaticMiniature` still exists, because it contains the default miniature uri definitions. I'm not sure where to move them, therefore the I left them where they were and simply moved the class up to the `lists` package

madoar:
Can we merge this, or is there still something left to do?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/widgets/lists/icons/IconsListElement.java
plata: Maybe you could put it into the theme?

madoar: What do you mean? The miniature specific settings were already moved to the theme in the context of this PR.

plata: The default miniature URI definitions.

madoar: No this is not possible, because the default image depends on the tab where you show the list widget.

plata: I don`t understand. It's just a `static URI`. You should be able to put that anywhere.

madoar: The problem is, that I don't have one such URI but three, each is a default image URI for a different tab, how should I add that to the theme?

plata: My idea was to just copy+paste the three URIs into the `Theme` or `ThemeManager` and delete the `StaticMiniature` because it's unused.

madoar: If we want to make the URIs theme dependent (e.g. the breeze dark theme can use different default URIs) we need to do a lot more work.
I'm not sure this should be part of this PR.

plata: I don't want to make the URIs theme dependent. I just want to get rid of a class which is only used to store three static URIs.

madoar: Hmm, the problem is I don't think these URIs belong inside the `Theme` nor `ThemeManager` classes, because they are specific to the list widget.
Maybe we can add them to the `ListWidgetEntry` class, but I'm not sure...

plata: Ok. Then we leave it like this for now.

PR #908, opened 2017-06-20
Remove border from unity close button

madoar:
Currently the "close" button in the details panel inside the unity theme contains a strange looking border.
This PR removes this border:
grafik


phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/unity/main.less
```diff
@@ -440,6 +440,13 @@
     -fx-background-position: left;
 }
 
+.detailsPane {
+    .closeIcon {
+        -fx-border-width: 0;
+        -fx-border-style: none;
```
plata: Isn't this redundant?

madoar: You mean the `0` and `none`?
I'm not sure, I found this in the main.less file for the unity theme at another location....

plata: Ok. Doesn't really matter.

madoar: I'll move the `no border` definition to the default theme...
This will reduce the need to override it in the other themes (except the breeze dark theme)

Issue #904, opened 2017-06-18
update library with bindings


plata:
see #903

plata:
Looks good.

PR #910, opened 2017-06-20
Refactoring of the MainWindowView and its implementing classes

madoar:
This PR solves #904.
In addition it does some similar refactoring and it renames all occurrences of `SideBar` to `Sidebar`.


madoar:
@plata The `this.installedApplicationsTab.setContent(availableShortcuts);` call inside the `populate` method is still needed, because otherwise we can't remove the waiting screen which was added to the center of the `BorderPane` before.

qparis:
screenshot?

madoar:
This PR introduces no visible changes.

PR #913, opened 2017-06-21
show failure panel if repository connection fails

### LINKED ISSUE #912: ViewApps failure panel not shown
If there is an Exception in `Repository.java` l.41 (`errorCallback.accept`) which should lead to `failurePanel` being shown, this does not happen. Also, there's no log output in this case (which definitely should be there).

see #901

plata:
Shows the failure panel and prints out the exception if connecting to the repository fails. Also updated German translations.

fixes #912


phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/Repository.java
```diff
@@ -38,6 +38,7 @@ default void fetchInstallableApplications(Consumer<RepositoryDTO> callback, Cons
         try {
             callback.accept(fetchInstallableApplications());
         } catch (Exception e) {
+            e.printStackTrace();
```
madoar: Please use a logger here.
In addition we shouldn't catch `Exception` here but a more specific subclass

i18n/keys.pot
```diff
@@ -398,13 +422,13 @@ msgstr ""
 msgid "Uninstall"
 msgstr ""
 
-#: src/main/java/org/phoenicis/javafx/views/mainwindow/FailurePanel.java:43
+#: src/main/java/org/phoenicis/javafx/views/mainwindow/FailurePanel.java:41
 msgid ""
-"Connecting to ${application.name} failed.\n"
-"Please check your connection and try again."
+"Connecting to the repository failed.\n"
+" Please check your connection and try again."
```
madoar: Is the space at the beginning needed?

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/Repository.java
plata: I'm not sure which specific subclass might be thrown.

madoar: I think it might be a good idea to add a custom `RepositoryException` class which can be thrown inside repositories and is used as a wrapper for other exceptions.
This will enable us to specify behavior that should throw an exception, which leads to an error screen, and other problems that should terminate the application (like out of memory or other bad things).

plata: We can think about this. But I don't want to do it in this PR (out of scope).

Issue #911, opened 2017-06-21
Prevent scripts from directly accessing Java classes


qparis:
Advantage: 
* Security, a script will no longer be able to damage filesystems
* Portability, one can rewrite a compatible client with existing scripts

madoar:
How do you want to provide the needed functionality, which is implemented in java code, to the scripts?

qparis:
Bean()

madoar:
But then you allow some kind of access to the Java classes, right?
What prevents the script developer then to access other classes?

qparis:
We could easily add a security mechanism, like a custom @Safe annotation and Bean() could check that this annotation exists

plata:
How about: <https://stackoverflow.com/questions/24466203/how-to-remove-java-apis-from-nashorn-engine>?

qparis:
This is complementary

madoar:
I think this might be a nice idea, the question is how to we do this with annotations?
I know of two ways to tell nashorn to ignore classes, the first are command line arguments (see the link from @plata for an example) and a `.policy` file. 
I'm not sure if there is a way to do this at runtime?

plata:
Do we even need the annotation if we make only a certain API available in JS as in the link I posted?

madoar:
I think we will need to have some java classes available to be used in javascript by the scripts.
For these we will need a policy that tells nashorn which classes can and can't be used from inside the scripts.

qparis:
@madoar  Exactly. It is possible (I've done it before with Nashorn). 
The idea is :
 * Bean() and SetupWizard() are allowed to be called because they are injected from java and not the contrary
 * Bean() always checks that the bean the script is trying to instanciate is annotated with @Safe

qparis:
final ScriptEngine scriptEngine = scriptEngineFactory.getScriptEngine(ClassFilter classFilter);

class CustomClassFilter implements ClassFilter {
    @Override
    public boolean exposeToScripts(String fullClassName) {
        return boolean 
    }
}

madoar:
This sounds interesting.
So you would use reflection inside `exposeToScripts` to check if the given class can be exposed?

qparis:
Yes

madoar:
Do you think we'll need to change anything inside our existing scripts when we do this, or would it be enough to provide such a `ClassFilter`?

qparis:
We will need to change some stuff in the wine classes

qparis:
[...]

PR #914, opened 2017-06-21
Make Nashorn engine safer

qparis:
This will be the first part to implement #911


phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/ClasspathRepository.java
```diff
@@ -76,6 +79,11 @@ public RepositoryDTO fetchInstallableApplications() {
         }
     }
 
+    @Override
+    public boolean isSafe() {
```
madoar: Why not use a `@Save` annotation instead of another overriden method here?

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/LocalisationInjector.java
```diff
@@ -0,0 +1,10 @@
+package org.phoenicis.scripts.nashorn.builtins;
+import org.phoenicis.scripts.nashorn.NashornEngine;
+
+public class LocalisationInjector implements EngineInjector {
```
madoar: JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/ProgressUiInjector.java
```diff
@@ -0,0 +1,24 @@
+package org.phoenicis.scripts.nashorn.builtins;
+
+import org.phoenicis.scripts.nashorn.NashornEngine;
+import org.phoenicis.scripts.wizard.UiProgressWizardFactory;
+import org.phoenicis.scripts.wizard.UiProgressWizardImplementation;
+
+import java.util.function.Function;
+
+public class ProgressUiInjector implements EngineInjector {
```
madoar: JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/ScriptUtilitiesInjector.java
```diff
@@ -0,0 +1,12 @@
+package org.phoenicis.scripts.nashorn.builtins;
+
+import org.phoenicis.scripts.nashorn.NashornEngine;
+
+import java.io.InputStreamReader;
+
+public class ScriptUtilitiesInjector implements EngineInjector {
```
madoar: JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/SetupWizardInjector.java
```diff
@@ -0,0 +1,24 @@
+package org.phoenicis.scripts.nashorn.builtins;
+
+import org.phoenicis.scripts.nashorn.NashornEngine;
+import org.phoenicis.scripts.wizard.UiSetupWizardFactory;
+import org.phoenicis.scripts.wizard.UiSetupWizardImplementation;
+
+import java.util.function.Function;
+
+public class SetupWizardInjector implements EngineInjector {
```
madoar: JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/IncludeInjector.java
```diff
@@ -0,0 +1,38 @@
...
+import org.phoenicis.scripts.interpreter.ScriptFetcher;
+import org.phoenicis.scripts.nashorn.NashornEngine;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Consumer;
+
+public class IncludeInjector implements EngineInjector {
```
madoar: JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/EngineInjector.java
```diff
@@ -0,0 +1,12 @@
+package org.phoenicis.scripts.nashorn.builtins;
+
+import org.phoenicis.scripts.nashorn.NashornEngine;
+
+public interface EngineInjector {
```
madoar: JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/BeanInjector.java
```diff
@@ -0,0 +1,35 @@
+package org.phoenicis.scripts.nashorn.builtins;
+
+import org.phoenicis.configuration.security.Safe;
+import org.phoenicis.scripts.nashorn.NashornEngine;
+import org.springframework.context.ApplicationContext;
+
+import java.lang.annotation.Annotation;
+import java.util.function.Function;
+
+public class BeanInjector implements EngineInjector {
```
madoar: JavaDoc

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/ClasspathRepository.java
qparis: Because you would not be able to make a difference between a safe repository and a safe bean that it is also a repository

madoar: Then can't we add a new annotation for this (e.g. `@SaveRepository`)?

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/Repository.java
```diff
@@ -34,6 +34,10 @@
 
     RepositoryDTO fetchInstallableApplications();
 
+    default boolean isSafe() {
```
madoar: JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/NashornEngine.java
```diff
@@ -24,23 +24,23 @@
...
 
     NashornEngine(ScriptEngine scriptEngine) {
         this.scriptEngine = scriptEngine;
     }
 
-    void eval(InputStreamReader inputStreamReader, Consumer<Exception> errorCallback) {
+    public void eval(InputStreamReader inputStreamReader, Consumer<Exception> errorCallback) {
         try {
             this.scriptEngine.eval(inputStreamReader);
         } catch (Exception e) {
```
madoar: Can we catch more specific exceptions here?

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/EngineInjector.java
```diff
@@ -0,0 +1,14 @@
+package org.phoenicis.scripts.nashorn.builtins;
+
+import org.phoenicis.scripts.nashorn.NashornEngine;
+
+/**
+ * Injects some code into a Script Engine
+ */
+public interface EngineInjector {
+    void inject(NashornEngine nashornEngine);
```
madoar: Maybe better name this method `injectInto(...)`?

madoar: JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/EngineInjector.java
```diff
@@ -0,0 +1,14 @@
...
+
+    default void throwException(Exception e) {
```
madoar: JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/NashornEngine.java
qparis: Hum, in this particular case, it is right to catch all exception, because we cannot predict what exception will be thrown in the script.

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/ClasspathRepository.java
```diff
@@ -195,18 +203,15 @@ public boolean equals(Object o) {
             return false;
         }
 
-        ClasspathRepository that = (ClasspathRepository) o;
-
-        EqualsBuilder builder = new EqualsBuilder();
-
+        final ClasspathRepository that = (ClasspathRepository) o;
```
plata: Could we please not mix this kind of changes into this PR? It's not that easy to understand anyways so I would like not to be disturbed by this.

plata:
Looks good overall. 

There's one thing I'm missing: We do not have a clear API definition for script developers. Is it possible to create JavaDoc for everything annotated with `@safe`? And even if so: we are not explicit about the exposed API. E.g. Someone might add a method to a class which is declared `@safe` without thinking that this might now be exposed to JS. Wouldn't it be cleaner to explicitly define the API as wrapper classes in a separate folder and only call the methods of the real classes from there?

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/ClasspathRepository.java
qparis: I think on the contrary we should encourage small code cleaning like this without dedicated PR as long as it does not change the logic of the program, specially when it concerns classes that are modified inside the pull request

qparis:
This would either: 
* prevent us from structuring our beans (there are so many beans used by scripts) 
* or force us to create a facade structure for each component we need in the framework

I think we have to be carreful about what is added in the safe scope, but overall, this solution is good. 
However, I agree that we need some kind of javadoc for the scripts

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/ClasspathRepository.java
qparis: Basically, all repositories instance of a given class have not necessarely the same safeness state. So annotations won't be powerful enough.
I'm not even sure that isSafe method is the right solution.

madoar: What alternatives do we have?

madoar: You should add a JavaDoc describing why the `ClasspathRepository` is "safe"

qparis: This is a stub method for now. This concept of safe repository will be handled in a second time

qparis: As an alternative, we could write in a json file if a repository is safe or not

madoar: I think this "safeness" information should be saved together with the repository location in `repositories.json`. So we'll should add it to the `RepositoryLocation` class.

qparis: Shall we disallow some repositories to be safe?

madoar: I don't know, but I know that the user should have the choice to decide, when he wants to add a repository, if the repository is "safe" or not, right?

madoar: For example, let's say I want to add a local repository, then I want to decide if the content of this repository should have access to my POL 5 installation or not.

qparis: I'm not sure that we want a remote repository to be markable safe. This will guarantee that all external code will be compatible with an possible unofficial client

plata: I'm not sure we need the concept of a safe repository at all. I see two possibilities:
1. You want to access Java code which is safe to use. In this case, that code should be provided as part of the API (safe Bean).
2. You want to access Java code which might be dangerous. We cannot allow that some user executes this script so you cannot be allowed to mark the repository as safe.

qparis: I agree that we do not need it but the idea was more about avoiding rewriting wine related stuff by puttin it in a safe Classpath repository

plata:
I was indeed thinking about the facade solution. I know it's a lot of work but I think the safe annotation is potentially dangerous. Can we really expect that every developer fully understands the consequences of a method she adds to a `@safe` annotated class (even if she's aware that this will be exposed to JS)?

E.g. what happens if I pass a `ShortcutDTO` with the name something like `/../../xyz` to `ShortcutManager#deleteShortcut`?
Or in `Extractor`, `Tar`, `Zip`, `DirectoryWatcherFiles`: I can probably retrieve a `File` object to anything I like, delete it, make it executable etc.

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/ClasspathRepository.java
plata: Which parts of the Wine script would have to be rewritten?

qparis:
Can we really expect that every developer fully understands the consequences of a method she adds to a @safe annotated class (even if she's aware that this will be exposed to JS)?
Yes because the code is reviewed.

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/ClasspathRepository.java
qparis: The part that is calling ProcessBuilder

plata:
Ok. How can we forbid this? By policy or is there a programmatic way to do so? How can we perform file operations without using `File`?

Any ideas on the documentation of the API?

phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/ClasspathRepository.java
plata: I see. In my opinion we should rather rewrite that part. I wouldn't like to have it in the classpath (e.g. you cannot fork that). Moreover, Wine serves as an example implementation for any other engine someone might want to implement.

qparis:
This PR is only about restricting Bean(). 
The restriction of Java API should be handled later as it will prevent wine from working. First, we need to remove **all** direct java calls from JS and use beans instead

madoar:
The `Files changed` tab contains unrelated changes now

qparis:
ok to test

qparis:
@madoar I don't understand why

madoar:
I think the reason is the way you've merged the upstream changes in your local branch.
When looking at the commit message `Merge remote-tracking branch 'origin/Issue-911' into Issue-911` I see that you've merged from origin to origin.
Whenever I merge from the upstream master, I first merge these changes to my local master repository and push them.
Afterwards I merge my local master to the branch in question.
An example for this can be found for #877, where you see the commit message `Merge branch 'master' into save-repository-list-as-json`.

qparis:
The thing is that IntelliJ does this automatically for me... It used to work before

madoar:
I have the habit of using a terminal when working with git. I kind of feel like I have more control over it when using a terminal.
Maybe it's a bug in the git client implementation from IntelliJ?

qparis:
ok to test

PR #915, opened 2017-06-23
remember window size

plata:
When closing Phoenicis, remember the window size and whether or not it is maximized. Restore this state on next startup.


madoar:
Is it possible to add an option to the settings where you can reset your settings to a default state?

plata:
We could do that but I don't see the point. I don't know any application which would handle it like this (you can just resize to whatever fits your need).

madoar:
It's not only about resetting the window size, I would like to have a reset function that can reset all user made changes to the factory settings.
This could be useful, if you accidentally remove the main script repository for example.

plata:
Ok. For the repository settings it makes sense to have this. Can you open a new issue for that?

PR #921, opened 2017-06-24
translate shortcut categories in library tab

plata:
Has to be done after the repositories have been fetched to have the translation files.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
@@ -58,7 +58,6 @@ public class MainController {
         appsController.setOnAppLoaded(() -> {
             enginesController.loadEngines();
             containersController.loadContainers();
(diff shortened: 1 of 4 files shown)


madoar:
ok to test

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
```diff
@@ -58,6 +58,7 @@ public MainController(String applicationName, LibraryController libraryControlle
         appsController.setOnAppLoaded(() -> {
             enginesController.loadEngines();
             containersController.loadContainers();
+            libraryController.updateLibrary();
```
madoar: Is this method called at another location too?
I'm just asking because you didn't remove another call to `updateLibrary`

plata: It' called when you search and create/delete a shortcut (`refresh`).

PR #924, opened 2017-06-24
fixed width of left sidebar

plata:
before

now

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftButton.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftButton.java
@@ -29,7 +29,6 @@ public class LeftButton extends Button {
         super(name);
         this.name = name;
         this.getStyleClass().add("leftButton");
-        this.setMaxWidth(Double.MAX_VALUE);
         this.setAlignment(Pos.CENTER_LEFT);
         this.setPadding(new Insets(2));
     }
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftButton.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftButton.java
@@ -29,7 +29,7 @@ public class LeftButton extends Button {
         super(name);
         this.name = name;
         this.getStyleClass().add("leftButton");
-        this.setPrefWidth(Double.MAX_VALUE);
+        this.setMaxWidth(Double.MAX_VALUE);
         this.setAlignment(Pos.CENTER_LEFT);
         this.setPadding(new Insets(2));
     }
--- a/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
+++ b/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
@@ -295,7 +295,6 @@
     .leftPaneScrollbar {
         -fx-border-width: 0;
         -fx-background-color: @main-color;
-        -fx-padding: 0.8em 0 0.8em 0;
     }
 
     .leftPaneScrollbar > .viewport {
--- a/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
+++ b/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
@@ -290,15 +290,12 @@
 /*******************************************************/
 .leftPane {
     -fx-background-color: @main-color;
-    -fx-min-width: 15em;
-    -fx-pref-width: 15em;
     -fx-padding: 1em;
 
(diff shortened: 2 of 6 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftButton.java
```diff
@@ -29,7 +29,7 @@ public LeftButton(String name) {
         super(name);
         this.name = name;
         this.getStyleClass().add("leftButton");
-        this.setPrefWidth(Double.MAX_VALUE);
+        this.setMaxWidth(Double.MAX_VALUE);
```
madoar: Isn't this the default max width?
In any case I would prefer to see these definitions in the css files.

plata: If it's not set, the buttons have different width. Is it possible to set it in CSS (how to say `Double.MAX_VALUE`)?

madoar: I believe you can set it to `Infinity` (see [here](https://stackoverflow.com/questions/26940341/how-to-set-fx-max-width-to-use-pref-size-in-javafx-css))

plata: I have it already.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/default/main.less
```diff
@@ -101,16 +101,12 @@
...
-  -fx-min-width: 15em;
-  -fx-pref-width: 15em;
   -fx-padding: 1em;
 
   .leftPaneScrollbar {
     -fx-border-width: 0;
     -fx-background-color: @background-color;
-    -fx-min-width: 15em;
-    -fx-padding: 0.8em 1em 0.8em 1em;
+    -fx-padding: 0.8em 0 0.8em 0;
```
madoar: If you add this to the default theme you shouldn't need it in the other themes too, right?

plata: Yes. But e.g. in the Unity theme it's already different. We have to do a complete cleanup sometime anyways.

madoar: Are you sure?
The unity theme doesn't define a `-fx-padding` for the `.leftPaneScrollbar`, so it should use the default one.

plata: You're right. What I had in mind was the padding of `leftPane`.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/unity/main.less
```diff
@@ -304,14 +304,11 @@
 /*******************************************************/
 .leftPane {
     -fx-background-color: @main-color;
-    -fx-min-width: 15em;
-    -fx-pref-width: 15em;
     -fx-padding: 0 0 0.83em 0;
 
     .leftPaneScrollbar {
         -fx-border-width: 0;
```
madoar: This is not needed, because the default theme already sets it

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/dark/main.less
```diff
@@ -95,15 +95,11 @@
 /*******************************************************/
 .leftPane {
     -fx-background-color: @background-color;
-    -fx-min-width: 15em;
-    -fx-pref-width: 15em;
     -fx-padding: 1em;
 
     .leftPaneScrollbar {
         -fx-border-width: 0;
```
madoar: This is not needed, because the default theme already sets it

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
```diff
@@ -357,6 +353,7 @@
         -fx-background-size: 2em;
         -fx-background-repeat: no-repeat;
         -fx-background-position: left;
+        -fx-max-width: Infinity;
```
madoar: Shouldn't this be part of the default theme too?

plata: Sure. Somehow put it into the wrong file...

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/unity/main.less
plata: Yes. But I don't want to change this now. As I said above, it's a general task to cleanup the CSS which does not belong to this PR.

PR #928, opened 2017-06-25
Css cleanup

madoar:
This PR cleans up our less/css theme files.


madoar:
I tried my best, but there were some things that looked wrong to me, which I corrected.
These correction should only occur inside the default theme. Please take a look at the addition (+) lines in the changes to see what I mean. 
If you find any discrepancies don't hesitate to write a comment.

madoar:
Just as a note:
The less files still contain some contradictions between the themes.
For example the `-fx-padding` in `#presentationBackground` is for all themes except the default theme `0.8em`. My guess is, that this should be changed in the default theme too.

plata:
If you looked at the themes and they still look ok, it's fine for me.

madoar:
I think they look ok

PR #937, opened 2017-07-01
fixed default window width

### LINKED ISSUE #932: Window of PlayonLinux should be slightly wider.
Actually in default display mode in tabs like Library, App, Containers or Engines fit only three columns. It is ugly. I think that it is necessary corrections.
zrzut ekranu z 2017-06-28 11-57-08
zrzut ekranu z 2017-06-28 11-57-42

plata:
fixes #932 

Although we will never be able to perfectly adjust the width of the application window (e.g. due to different width of the sidebar because of translation), this PR should improve the appearance with default settings in the majority of cases.

default width

### DIFF
--- a/phoenicis-configuration/src/main/resources/Linux.properties
+++ b/phoenicis-configuration/src/main/resources/Linux.properties
@@ -24,7 +24,7 @@ application.theme                           =           default
 application.scale                           =           12
 application.viewsource                      =           false
 application.windowHeight                    =           600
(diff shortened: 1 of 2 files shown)


madoar:
Does it look "good" for all themes?

madoar:
In addition did you test it with the default scaling settings?

plata:
It's the default scaling, yes. The scaling should not be different in the themes.

PR #938, opened 2017-07-07
fixed wizard button width

plata:
Before:
before

Now:
now


plata:
Yes that's probably right.

qparis:
I would prefer the buttons to have the same size

plata:
Yes. But if I have the choice between the text is readable and the buttons have the same size, I'll go for readable.

madoar:
@qparis I agree with @plata for these two buttons. I think it's another topic when we add images to the equation.

PR #940, opened 2017-07-07
fixed shortcut button width

plata:
Before:
before

Now:
now


madoar:
Can we show the three buttons in a grid? This should allow us to easily set an equal space between the components, in addition this should force the components to have the same width and height.
This would make it look better in my eyes.

plata:
Like this?
grid

plata:
It's in percentage, so what will happen in this case is that the button label will be cropped.

Issue #939, opened 2017-07-07
cannot start and uninstall app


plata:
```
[pool-4-thread-4] ERROR org.phoenicis.multithreading.ControlledThreadPoolExecutorService - java.lang.IllegalAccessError: You are not allowed to instanciate uiQuestionFactory (of class org.phoenicis.javafx.views.scriptui.JavaFxUiConfiguration$$Lambda$479/1908057566) from a script.
	at org.phoenicis.scripts.nashorn.builtins.BeanInjector.fetchBean(BeanInjector.java:35)
```

@qparis I think this has been introduced by #914

madoar:
Which script needs this?

plata:
WineShortcutReader

plata:
You cannot even start apps anymore.

PR #941, opened 2017-07-07
hotfix: cannot start/uninstall apps

plata:
hotfix to restore basic usability of Phoenicis until #939 is fixed

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/BeanInjector.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/BeanInjector.java
@@ -1,10 +1,8 @@
 package org.phoenicis.scripts.nashorn.builtins;
 
-import org.phoenicis.configuration.security.Safe;
 import org.phoenicis.scripts.nashorn.NashornEngine;
 import org.springframework.context.ApplicationContext;
 
-import java.lang.annotation.Annotation;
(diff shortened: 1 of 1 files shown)


plata:
I tried but it didn't work. Maybe I just didn't do it right but I don't know how to do it.

qparis:
What error do you have when you add @Safe annotation?

plata:
There was no error. It just didn't work. Also I'm not quite sure where I have to add it.

qparis:
On top of the class

plata:
It's an interface.

qparis:
Add it on the implementation

plata:
I cannot find any implementation (it's `UiQuestionFactory`).

qparis:
Implementations are anonymous class currently. They must be promoted to proper class to be annotated

plata:
Ok. There's a couple of place where this is the case and currently the software is just not working. In my opinion we should do the hotfix now and the test everything before we enable the security stuff. Also we should have tests which avoid something like this in the future.

qparis:
I don't agree that we should merge this hotfix, it will require more work than fixing the two beans that we have missed. 
Also, if we merge this, this will hide us any other problems that could exist
Currently, POL-POM-5 is not released so this is normal that we have such problems in the master.

madoar:
I agree with @qparis, I think we should prefer a longterm solution instead of a short term solution.

PR #946, opened 2017-07-15
 	show engine tools in TilePane

### LINKED ISSUE #933: Wine tool tabs are displayed incorrectly.
zrzut ekranu z 2017-06-28 11-58-28

plata:
fixes #933

tile pane


madoar:
How does the layout react, if you make the window smaller?

plata:
You cannot change the width of the right panel currently.

madoar:
You're right you can't explicitly change the width of the details panel, but you can implicitly change it by changing the width of the whole window.
In the current version, if you take a look at the not maximized window you'll discover that the details view looks kind of clipped, i.e. the close button isn't shown. I believe the reason for this may be solved by this PR.

plata:
If you make the whole window smaller than the right panel, it's still clipped. However, I have the impression that this is not caused by the TilePane but rather by the containing tab (it happens even with a FlowPane).

madoar:
One of the contained tabs in the details pane is too large.
This makes the whole details pane too large.

plata:
I'm not sure. Maybe it's just the preferred width?

madoar:
I did some experiments some weeks prior, and as far as I remember the reason was a too large tab.

plata:
Do you know which one? I can't see which tab is that big. However, also the close icon is not visible anymore, so this does not relate to this PR (we should fix it nevertheless).

PR #948, opened 2017-07-15
remove logo in menu

### LINKED ISSUE #900: While run playonlinux not saw logo.
While run playonlinux not show logo.
When I change tab to other, logo is display normal.
it happens always when I run playonlinux.
zrzut ekranu z 2017-06-17 20-56-20
zrzut ekranu z 2017-06-17 20-56-28

### LINKED ISSUE #936: Menu width slightly wider and set size of rigidly.
Menu width needs to be slightly wider (up to red line shown in the picture) and its size rigid.
In this issue menu has diferent sizes in each tabs.
https://github.com/PlayOnLinux/POL-POM-5/issues/934
I think (but i don't know) that it is caused by size of menu's elements.
That's why you should change it.
zrzut ekranu z 2017-06-28 12-25-32
Long story short, the menu should be the same with as playonlinux logo.

plata:
no logo

fixes #936 
fixes #900

I believe removing the logo from the menu completely is the best solution because:
- As #936 shows, it suggests that the sidebar should have the same size. However, this doesn't make a lot of sense. If the sidebar should have the same size, the logo should be in the sidebar. As the sidebar logically belongs to the tabs and not the whole application, it should also be in the tab visually.
- It consumes a lot of space.
- I cannot recall any other application showing the logo in this place.


madoar:
I don't like that the menu begins over the sidebar.
I would like it more if the menu begins next to the sidebar, in terms of its horizontal position.

plata:
That's exactly what I described in the first bullet point: The sidebar belongs to the tab and not to the application. This should be represented by the design of the UI.

madoar:
I'm not sure this is visible to the user, because every tab contains a sidebar. 
This in my eyes makes it look like the sidebar is a functionality every tab must provide.
Therefore I think we can reflect this by adding an offset to the first tab label (if this is possible).

plata:
That doesn't make any sense in my opinion. Either the sidebar belongs to the tabs or to the application. If it belongs to the tabs, it should be in the tab. If it belongs to the application, we should have one sidebar (one object in the code) and show the menu right of the sidebar (such that the sidebar covers the complete height of the window).

madoar:
Just to be sure we're talking about the same thing:
I'm in favor of this approach: image (see also #936 for more details)
@qparis what do you think about this topic?

qparis:
I don't see any problem if the menu begins over the sidebar ; we could decide to let a theme put it on the right later.
I'm not totally sure we should remove the logo but it looks fine overall

madoar:
@qparis another suggestion made in #936 was to move the logo to the sidebar, take a look there for a screenshot.
Do you think this would be better?

plata:
@madoar ah ok. Didn't realize you were talking about the centered tabs. Well... There's a small problem about it. I have no idea how to implement this.

madoar:
@qparis I thing you should decide if we should go with this.

qparis:
The original proposal is fine for me. Do you agree with that?

madoar:
I think this PR can be used, it's just not my favorite approach, because I'm not used to the menu being directly over an always visible sidebar.
Therefore I want you to decide if this should be merged :)

qparis:
So far we can. Maybe we will decide to redesign the app later on

madoar:
Kk, then I'll approve this.

PR #949, opened 2017-07-15
improve style of "add repository" dialogs

### LINKED ISSUE #887: improve style of "add repository" dialogs
see #886

plata:
fixes #887

type
local
git
classpath

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
@@ -127,6 +127,7 @@ public class RepositoriesPanel extends StackPane {
         this.addButton.setText(tr("Add"));
         this.addButton.setOnAction((ActionEvent event) -> {
             AddRepositoryDialog dialog = new AddRepositoryDialog();
+            dialog.initOwner(this.getParent().getScene().getWindow());
 
             Optional<RepositoryLocation<? extends Repository>> successResult = dialog.showAndWait();
 
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/addrepository/ChooseRepositoryTypePanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/addrepository/ChooseRepositoryTypePanel.java
@@ -2,6 +2,7 @@ package org.phoenicis.javafx.views.mainwindow.settings.addrepository;
 
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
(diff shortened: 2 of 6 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
```diff
@@ -127,6 +127,7 @@ private void populateRepositoryGrid() {
         this.addButton.setText(tr("Add"));
         this.addButton.setOnAction((ActionEvent event) -> {
             AddRepositoryDialog dialog = new AddRepositoryDialog();
+            dialog.initOwner(this.getParent().getScene().getWindow());
```
madoar: Just asking:
What does this line do?
Does it add the stylesheet from the parent to the dialog?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/addrepository/ChooseRepositoryTypePanel.java
```diff
@@ -74,6 +75,8 @@ public RepositoryType fromString(String string) {
         choiceBoxLabel.setLabelFor(choiceBox);
 
         HBox content = new HBox(choiceBoxLabel, choiceBox);
+        content.setAlignment(Pos.CENTER_LEFT);
```
madoar: I don't like this approach.
I think we should set these things via css.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/RepositoriesPanel.java
plata: Didn't check if the stylesheet is applied but if you do not have this (probably yes), the dialog does not use the correct window icon.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/addrepository/ChooseRepositoryTypePanel.java
plata: I'm not sure. Do we really want every little dialog to be themeable?

madoar: It's not only a question about themeability. 
The important benefit of moving **every** style relevant line of code to css is, that we have everything in a single place, in a single file.
This in turn makes it quite easy to change some/fix some wrong layout without the need to search the whole codebase first.

In addition:
I don't think it's bad if make the dialogs themeable. It doesn't give us any disadvantage I think, right?

PR #951, opened 2017-07-15
use white background for icon list (except Breeze Dark)

### LINKED ISSUE #925: Set background color of the center component inside a MainWindowView to white
Currently all themes except the breeze dark theme use a gray background color in the center component of their tabs.
This is especially visible for the icons list widget, which has a gray background for selected and hovered elements.
To make this better readable we need to change the background color to white for all tabs for the default, dark and unity themes.

plata:
fixes #925

white


madoar:
Have you checked, that this works with all themes, e.g. with the unity theme too?

plata:
dark
unity
breeze

PR #961, opened 2017-07-29
fetch available engines from repository

plata:


phoenicis-engines/src/main/java/org/phoenicis/engines/EnginesSource.java
```diff
@@ -45,17 +46,22 @@ public EnginesSource(ScriptInterpreter scriptInterpreter, ObjectMapper objectMap
...
-                }, this::throwError), this::throwError);
+        for (CategoryDTO categoryDTO : categoryDTOS) {
+            final String engineName = categoryDTO.getName();
+            interactiveScriptSession.eval("include([\"Engines\", \"" + engineName + "\", \"Engine\", \"Object\"]);",
+                    ignored -> interactiveScriptSession.eval("new " + engineName + "().getAvailableVersions()",
+                            output -> {
+                                EngineCategoryDTO wine = new EngineCategoryDTO.Builder().withName(engineName)
+                                        .withDescription(engineName).withSubCategories(unSerialize(output)).build();
+                                engines.add(wine);
+                                callback.accept(engines);
```
madoar: Just as a note: These two lines:
```java
engines.add(wine);
callback.accept(engines);
```
look really strange to me.
It looks like the list gets longer with every callback call, I'm not sure if this is the correct behavior.

plata: I think it's correct. It creates a new `EngineCategoryDTO` for every `CategoryDTO`.

madoar: What I think is strange, that the list, which is passed to the callback becomes longer with every call... Normally I would expect a call `callback.accept(wine)` inside the loop or a single call `callback.accept(engines)` after the loop has ended.

plata: Oh yes, you are right. It should be in the end.

plata: Mh. Just tried. It's not working for some reason. The `engines` list is empty at that point.

plata: Probably because it's running in a separate thread. For now it's working, we will have to look into it again when we get more engines.

madoar: Can you create a new issue for this?

plata: Yes. Anyways I'm not so happy with the current structure of `CategoryDTO` and `EngineCategoryDTO`. I would prefer to have the `EngineCategoryDTO` in the repository DTO structure directly. Maybe something like extending `ApplicationCategoryDTO` and `EngineCategoryDTO` from `CategoryDTO`.

PR #960, opened 2017-07-28
Add an option to show applications disregarding the used operating system

madoar:
Fixes #959.

What is still missing:
Through the addition of the checkbox the `FilterRepository` isn't used anymore by the `DefaultRepositoryManager`, therefore if you're on linux you see some empty categories.
These categories need to be removed by an additional `FilteredList`.

PR #964, opened 2017-07-30
application filter: handle incomplete ApplicationDTO

plata:
The ApplicationFilter can now handle ApplicationDTO containing null values. This can happen e.g. if the application.json could not be read or is incomplete. Before, Phoenicis crashed with NullPointerException.


plata:
@madoar you should consider this in #960 as well.

madoar:
I need to merge the master branch either way, so I think we should merge this first.
I guess this fixes #965?

madoar:
I guess we can make the null check a little bit more beautiful, please take a look at https://commons.apache.org/proper/commons-collections/javadocs/api-release/org/apache/commons/collections4/CollectionUtils.html#isEmpty(java.util.Collection)

plata:
Yes, it fixes #965.

PR #972, opened 2017-08-13
Fix category selection in application tab

### LINKED ISSUE #971: apps: category buttons not working
It's not possible to select any category apart from "All".

@madoar could this be related to your changes in #967?

madoar:
Fixes #971.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationFilter.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationFilter.java
@@ -43,7 +43,7 @@ public class ApplicationFilter {
      * @param filterTextMatcher The matcher function for the filter text
      */
     public ApplicationFilter(OperatingSystemFetcher operatingSystemFetcher,
-            BiPredicate<String, ApplicationDTO> filterTextMatcher) {
+                             BiPredicate<String, ApplicationDTO> filterTextMatcher) {
         this.operatingSystemFetcher = operatingSystemFetcher;
         this.filterTextMatcher = filterTextMatcher;
 
@@ -147,20 +147,36 @@ public class ApplicationFilter {
         /*
          * A category can be shown, if it contains at least one visible application
          */
-        return category.getApplications().stream().anyMatch(application -> filter(application));
+        return category.getApplications().stream().anyMatch(application -> filter(application, true));
     }
 
     /**
-     * Filter function for {@link ApplicationDTO} objects
+     * Filter function for {@link ApplicationDTO} objects.
+     * This method call {@link #filter(ApplicationDTO, boolean)} with <code>ignoreFilterCategoryTest = false</code>.
      *
      * @param application The application which should checked
      * @return True if the given <code>application</code> fulfills the filter conditions, false otherwise
      */
     public boolean filter(ApplicationDTO application) {
+        return filter(application, false);
+    }
+
+    /**
+     * Filter function for {@link ApplicationDTO} objects
+     *
+     * @param application              The application which should checked
+     * @param ignoreFilterCategoryTest True if an optional filter category should be ignored, false otherwise
+     * @return True if the given <code>application</code> fulfills the filter conditions, false otherwise
+     */
+    private boolean filter(ApplicationDTO application, boolean ignoreFilterCategoryTest) {
         /*
-         * An application can be shown, if it contains at least one visible script and its text matches the filter text
+         * An application can be shown, if:
+         * - it belongs to the filter category, if such a category is set
+         * - it contains at least one visible script
+         * - its text matches the filter text
          */
-        return application.getScripts().stream().anyMatch(script -> filter(script))
+        return (ignoreFilterCategoryTest || filterCategory.map(category -> category.getApplications().contains(application)).orElse(true))
+                && application.getScripts().stream().anyMatch(script -> filter(script))
                 && filterTextMatcher.test(filterText.getValue(), application);
     }
 
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationFilter.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationFilter.java
@@ -43,7 +43,7 @@ public class ApplicationFilter {
      * @param filterTextMatcher The matcher function for the filter text
      */
     public ApplicationFilter(OperatingSystemFetcher operatingSystemFetcher,
-                             BiPredicate<String, ApplicationDTO> filterTextMatcher) {
+            BiPredicate<String, ApplicationDTO> filterTextMatcher) {
         this.operatingSystemFetcher = operatingSystemFetcher;
         this.filterTextMatcher = filterTextMatcher;
 
@@ -175,7 +175,8 @@ public class ApplicationFilter {
          * - it contains at least one visible script
          * - its text matches the filter text
          */
-        return (ignoreFilterCategoryTest || filterCategory.map(category -> category.getApplications().contains(application)).orElse(true))
+        return (ignoreFilterCategoryTest
+                || filterCategory.map(category -> category.getApplications().contains(application)).orElse(true))
                 && application.getScripts().stream().anyMatch(script -> filter(script))
                 && filterTextMatcher.test(filterText.getValue(), application);
     }
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationSidebar.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationSidebar.java
@@ -45,7 +45,7 @@ public class ApplicationSidebar extends LeftSidebar {
     private LeftScrollPane centerContent;
 
     private ObservableList<CategoryDTO> categories;
-    private FilteredList<CategoryDTO> filteredCategories;
+
     // the toggleable categories
     private LeftToggleGroup<CategoryDTO> categoryView;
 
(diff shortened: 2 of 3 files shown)


plata:
It's better but still not ok. The category button should be selected if you click it.

madoar:
Hmm, I have no clue why the category button isn't shown as selected...
I've added a test output in `LeftToggleButton` line 63 to test if the button has been selected, which is true.
So it should be shown as selected...

madoar:
I've done some more experimentation. It seems like the `MappedList` in `LeftToggleGroup` triggers the converter function with every category selection.
This triggers the creation of a new category button, which is obviously not selected...
The question is now, why does the `MappedList` trigger the converter function when a category has been selected?
This should only happen if an add, remove, permutation or update event has been triggered, which shouldn't be the case during category selection.
@plata can you take a short look over the code? Maybe you see something I've missed.

plata:
Not quite sure yet why, but if you put a breakpoint in `createCategoryToggleButton` you see that `ApplicationFilter#fire` triggers `MappedList#sourceChanged` (`addRemove`).

plata:
It's in `ApplicationSidebar` `Bindings.bindContent(categoryView.getElements(), filteredCategories);`. The filter changes `filteredCategories` which then triggers the recreation of all category buttons. If you change it to `Bindings.bindContent(categoryView.getElements(), categories);`, the buttons are selected correctly (of course we cannot do it like this).

madoar:
Hmm.
I think this is founded in `FilteredList#predicateProperty`, where `invalidated()` is defined as:
```java
@Override
protected void invalidated() {
   refilter();
}
```
`refilter` seems to trigger an add-remove change, with I believe all elements in the list...

I'm not sure how to handle this. I can think of two ideas:

1. we change the order of the observable list, and first map all categories to toggle buttons. Afterwards we filter them, to only show the toggle buttons with a category containing at least one visible application
2. we set the category filter only once

plata:
I'm not even sure the concept is consistent: If the same filter is used for selecting a category button and typing a search text, it would be logical to behave in the same way. So selecting a category would delete all other category buttons which is clearly not what we want.
Maybe it would therefore be better not to use the same filter for those cases and rather handle it in the way it was done before.

madoar:
Previously we didn't filter the categories at all.
I added a category filter because it makes sense if you enter a search keyword.
For example if you enter "league" in the search field, you will discover that only the all category and the games category are visible.

plata:
I agree that it makes sense to do this. However, the selected category must be shown as selected as well. I'm just not sure if the clicking the category button can work properly if we use the same filter mechanism as for e.g. the search box.

madoar:
We would need to modify the `LeftToggleGroup` class to contain the filtered list.
In addition we would need to modify the order in which the operations are done:

1. map the categories to `ToggleButton`s
2. filter the previously mapped `ToggleButton`s according to the number of visible applications

To do the second step after the first, we would need to add an internal wrapper component inside the `LeftToggleGroup` that makes it possible to find out which category belongs to which `ToggleButton` in the second step.

plata:
I'm not sure. If `sourceChanged` would work properly, everything would be ok. Currently it reports items as changed which didn't change.

madoar:
You're right about that, the fact that the `FilteredList` will report items as changed will not change through my suggestion.
Another approach would be to write our own `FilteredList` class, which is more clever and only fires a `sourceChanged` event when a filter result has changed.
This could happen in two cases:

- an element has been added, removed, updated or permutated in the source list
- the filter has been changed **and** the new filter function produces a different result than the old one

What do you think?

plata:
I believe that we're using it in the wrong way maybe. Check out e.g. <https://stackoverflow.com/questions/30915007/javafx-filteredlist-filtering-based-on-property-of-items-in-the-list>.

madoar:
This could work
Do you want to give it a try?

plata:
I would. But to be honest, I don't know much about this whole stuff so I'm a bit lost regarding how exactly it has to be applied in our case.

madoar:
I've just taken a more detailed look at the stackoverflow thread and the java api documentation.
It seems like you can specify some observable attributes via `public static <E> ObservableList<E> observableArrayList(Callback<E,Observable[]> extractor)` that should be watched by the observable list.
Whenever one of the given attributes, given by the `extractor` callback, changes a `sourceChanged` call will be triggered.
If you have a `FilteredList` afterwards in the pipeline you'll refilter only the changed items (the items that triggered the `sourceChanged` call).

Our problem is, that we don't have an `Observable` for each category, which the `ObservableList` should track and I don't think we want to add one to the `Category` class.

plata:
What if we create a list of all possible buttons once and then only add them to the sidebar in `populate`?

madoar:
Without filtering?

plata:
With filtering but without recreating the buttons over and over again.

plata:
Similar to what you have suggested above: a map with category ID -> button. And then in `populate` fetch the buttons from that map instead of creating a new one.

madoar:
This can be achieved through switching the order of the observable lists.
So if we first map the categories to toggle buttons and then filter the created toggle buttons we would receive this result. Still we would trigger a complete refiltering of the filtered list, which is not really efficient.

plata:
That's true. I cannot believe that there is no other way to avoid this.

madoar:
It makes kind of sense to invalidate the whole list and refilter it, when the filter function changes.
The problem is that we use the change of the filter function as a trigger to check the list.
Ideally we would want to only invalidate the elements that are affected by the change of the filter function. The only "clean" way I see to do this is to implement our own `FilteredList` class, which fixes the `refilter` behavior.

plata:
I don't really like this because it feels somewhat hacky but if we cannot find a better solution, we should do it.

madoar:
You're right, this is like using cannons to shoot on sparrows...
Alternatively we could introduce a wrapper class around the category objects, that combine them with a filter `BooleanProperty`. This should allow us to use the approach described in https://stackoverflow.com/questions/30915007/javafx-filteredlist-filtering-based-on-property-of-items-in-the-list. 
The problem is, that I don't know how we should update these filter `BooleanProperty` objects, when the some filter settings are updated.
In addition I don't really know if this approach will still lead to an invalidation of the whole list...

plata:
What if we remove the filtering of categories for now and open a new issue to re-introduce it? Maybe @qparis has some idea how to do it properly.

madoar:
We can do that.
When thinking about the category filtering I just discovered another problem:
What should happen if you've selected for example the category "Games" and search for `notepad++`?
The current behavior would remove the category "Games", because it doesn't contain `notepad++`, but meanwhile another category is not selected.

plata:
Good point. Probably it should select "All".

madoar:
Ok, I've just done some additional experiments. It seems like the approach taken in https://stackoverflow.com/questions/30915007/javafx-filteredlist-filtering-based-on-property-of-items-in-the-list doesn't work.
I've tried the following:

```java
    private void populateCategories() {
        this.categories = FXCollections.observableArrayList(category -> new Observable[] {Bindings.createBooleanBinding(() -> filter.filter(category), filter.categoryFilterProperty())});
        this.filteredCategories = new FilteredList<CategoryDTO>(categories, category -> filter.filter(category));

        this.categoryView = LeftToggleGroup.create(tr("Categories"), this::createAllCategoriesToggleButton,
                this::createCategoryToggleButton);

        Bindings.bindContent(categoryView.getElements(), filteredCategories);
    }
```

But it seems like the list `categories` doesn't detect any changes in the observables.
I've first thought, that maybe the creation of the `BooleanBinding` doesn't work, but I've then tried to read the values before and after changes to detect if the value switches with the input of a search term, which worked.
Maybe it would work if we don't create the `BooleanBinding` inside a lambda as a local variable, but I'm not sure where else to create and store it.

plata:
Does it work if you have the property in the `CategoryDTO` directly? I don't think we want that in the end, just to check if the other stuff works as intended.

madoar:
I'm not sure how to add it to `CategoryDTO`, because the property itself depends on the filter, which I don't want to add to `CategoryDTO`. If you have a way to test it, feel free to do so.

plata:
Not really. Was just a quick idea.

madoar:
The important question is, where do we want to have the filtering mechanism?
We have two choices:
- inside the sidebar class
- inside the `LeftToggleGroup` class

Currently, the only tab that requires a filter for the `LeftToggleGroup` is, as far as I know, the "Apps" tab, but I think it's the cleanest place to add the filter mechanism.

plata:
The problem is that there are in fact two filters (logically). The first one filters the applications. The second one filters the `LeftToggleGroup` based on that filter.

madoar:
Yes

plata:
I believe that this should be reflected in the software (i.e. separate the filter for the category buttons from the application filter). Then, the filter for the category buttons should be in `LeftToggleGroup`.

plata:
That's good. So we have a functional version for alpha. I'll approve it like this. Could we filter the empty categories?

madoar:
Before doing my last commit this morning, I first tried to create a new class `LeftFilteredButtonGroup`, which adds filter functionality to the `LeftButtonGroup`. During this I tripped over multiple problems:
- when using `FXCollections#observableArrayList(Callback<E,Observable[]> extractor)` in combination with `Bindings#createBooleanBinding(Callable<Boolean> func, Observable... dependencies)` I found out that any change in one of the given `dependencies` will trigger an update in the `BooleanBinding`, which in theory would trigger an update for the element in the `ObservableList`. This means, that if you use a single `dependency` for all elements in the `ObservableList`, like we wanted, we would trigger an update of the whole list, which is something we don't want.
- the `Property` and `Binding` class structure in the JavaFX API is not really easy to understand. I first thought I could write some custom Observables that fit our needs, but I stopped this, because I couldn't completely find where each method was defined and how they call each other. Take for example the class `BooleanBinding`. First it extends another class, implements two interfaces, and is then often created through an abstract class... In addition I haven't yet found the location where the `set(...)` method is defined...

Therefore I'm not sure it's feasible to use the `FXCollections#observableArrayList(Callback<E,Observable[]> extractor)` approach.

About removing the empty categories:
We could remove them but I don't think we should. Because these categories are normally filled with applications that are filtered out via the checkboxes on the left side (often through the OS checkbox).

plata:
I meant the categories which do not contain any application at all like "Science" currently.

madoar:
You're right, categories that are empty by default should be filtered out.
What do you think, should I add a new filter method for that or should I change the existing one?

I would like to retain the current filter method for `CategoryDTO`, because I think we'll need to fix the filter behavior later on. It's just the case that I'm currently not sure how to do this best.

plata:
Maybe we could filter them in `fetchInstallableApplications` already?

madoar:
How about like this?

PR #976, opened 2017-08-15
use logback to write to logfile

### LINKED ISSUE #975: Error logs should be saved
I got an error log when trying to install Age of Empires II HD, but the log wasn't saved, so since I had closed the window with the log in it, I couldn't attach the log to [my bug report for that script](https://github.com/PlayOnLinux/Scripts/issues/324#issuecomment-322242473).

plata:
fixes #975

### DIFF
--- a/phoenicis-cli/pom.xml
+++ b/phoenicis-cli/pom.xml
@@ -78,7 +78,7 @@
 
         <dependency>
             <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-simple</artifactId>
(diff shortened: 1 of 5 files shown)


plata:
It would be nice to set the log level and log file locations from Phoenicis (i.e. store it in `config.properties`). I'm not sure if that's possible (maybe somehow with `PropertyConfigurator`?). I didn't manage to use other properties in the log4j.properties directly.

madoar:
I think it would be better to use `logback` (see here: https://logback.qos.ch/) instead of `log4j`.

PR #979, opened 2017-08-16
added isUpdateAvailable to Downloader

plata:
Allows to check whether a URL contains a newer version of a downloaded file.


madoar:
Ok to test

PR #980, opened 2017-08-18
moved wineprefix directory to containers directory

plata:
fixes PhoenicisOrg/Scripts#340

To use existing prefixes, just move the "wineprefix" directory to a new "containers" directory and change the .shortcut files in "shortcuts".


plata:
ok to test

madoar:
Just asking:
Shouldn't the wineprefix path be set somewhere inside the scripts repository (i.e. in javascript code)? Otherwise we need to be willing to add a variable for each engine type to the POL 5 code repository, which I believe we don't want to do, right?

plata:
Yes you're completely right. I thought it was already handled like this which is the reason why I added the issue to Scripts. I want to avoid that people have to move their prefixes manually in alpha. To achieve this, the current solution is sufficient. We can implement your suggestion later because it will not have a visible effect for the user.

madoar:
Ok, just to complete my idea:
I think we should add the base path, i.e. `${application.user.root}/containers/`, to the config file in POL 5, like it is currently done. I believe we should just set the final foldername in javascript, i.e. `wineprefix`.

plata:
Totally agree. Can you open a new issue to track this?

PR #982, opened 2017-08-18
check if application.json and script.json exist when loading repository

plata:
fixes error log messages on startup

### DIFF
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/LocalRepository.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/repositoryTypes/LocalRepository.java
@@ -177,9 +177,14 @@ public class LocalRepository implements Repository {
 
         for (File applicationDirectory : applicationDirectories) {
             if (applicationDirectory.isDirectory()) {
-                File applicationJson = new File(applicationDirectory, "application.json");
-                final ApplicationDTO.Builder applicationDTOBuilder = new ApplicationDTO.Builder(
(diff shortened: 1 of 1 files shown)


madoar:
I'm not sure it's a good approach to let POL 5 run if some files for scripts are missing, at least in the official script repositories. I think in this case we should ignore (i.e. not load) the script/application, whose file is missing and write a explaining message to the console.

plata:
I thought the same but it won't work. The problem is that everything in `Engines` and `Functions` is also stored in `CategoryDTO` and `ApplicationDTO`. So if you do what you suggest, you suddenly do not have Wine etc. anymore.
In my opinion, the repository management is a bit weird in that regard. We should either find better names for `CategoryDTO` and `ApplicationDTO` or structure the repository DTO tree differently.

madoar:
I'm not really sure why we implement the engines in javascript like we do.
I think it would be much better, with regards to the interoperability with our java application, to define a java interface or abstract class, called `Engine` or something similar with a set of required method stubs. This interface/abstract class can then be extended in javascript. Next we would be able to use this extended/implemented javascript class from java.
For more information take a look at http://docs.oracle.com/javase/8/docs/technotes/guides/scripting/prog_guide/javascript.html#CIHJCFAH

plata:
We can certainly discuss this. However, this is not related to this PR anymore. Please open a new issue or add it to the "generic engines" issue.

madoar:
Ok then let's do it like this.

PR #989, opened 2017-08-20
implemented JavaFxSettingsManager

### LINKED ISSUE #950: implement JavaFxSettingsManager
Currently, all settings (except repositories) are stored in one properties file and are managed by `SettingsManager`. However, some of the settings (like window size) only apply to JavaFX. Therefore, those should be extracted and handled separately. This is also required for #850 because the list type enum cannot be used in `SettingsManager` right now (circular dependency between phoenicis-javafx and phoenicis-settings).

plata:
fixes #950

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/settings/JavaFxSettingsManager.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/settings/JavaFxSettingsManager.java
@@ -26,6 +26,9 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 
+/**
+ * manages JavaFX settings
+ */
 public class JavaFxSettingsManager {
     @Value("${application.theme}")
     private String theme;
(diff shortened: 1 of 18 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/settings/JavaFxSettingsManager.java
```diff
@@ -0,0 +1,124 @@
...
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.util.DefaultPropertiesPersister;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+public class JavaFxSettingsManager {
```
madoar: Please add JavaDoc to this class.

plata:
I don't think it affects CLI.

madoar:
@qparis can you take a look at this? I think this is working but I can't estimate if this has any effects on the cli.
In addition these double build success messages from jenkins are quite annoying...

qparis:
I don't think this commit will have any impact on CLI. I don't know why jenkins is behaving like this but I will investigate on it

plata:
@qparis regarding Jenkins: it looks like it runs because it detects a PR but the Jenkins run is not set as required in GitHub so we get the comment instead of the notification via a green check mark. Also, it is currently possible to merge even if Jenkins did not run at all or failed.

PR #999, opened 2017-08-20
use same style in all details views

### LINKED ISSUE #1006: use same style in all details views

plata:
fixes #1006

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginePanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginePanel.java
@@ -49,8 +49,6 @@ final class EnginePanel extends DetailsView {
 
         this.engineDTO = engineDTO;
 
-        this.getStyleClass().add(CONFIGURATION_PANE_CSS_CLASS);
-
         this.populateContent();
     }
 
@@ -104,7 +102,9 @@ final class EnginePanel extends DetailsView {
         progressSpacer.getStyleClass().add("engineSpacer");
         VBox.setVgrow(progressSpacer, Priority.NEVER);
 
-        this.setCenter(new VBox(informationContentPane, spacer, buttonBox, progressSpacer));
+        final VBox content = new VBox(informationContentPane, spacer, buttonBox, progressSpacer);
+        content.getStyleClass().addAll(CONFIGURATION_PANE_CSS_CLASS);
+        this.setCenter(content);
     }
 
     public void showProgress(VBox progressUi) {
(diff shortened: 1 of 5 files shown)


madoar:
I would prefer a solution that works completely without setting a second css class for the details view. 
This would make the different details views look more similar than they look if they come with different styles.

plata:
I do not believe that this is possible. You cannot control the content of the details panel (at least without artificially restricting the possibilities). Therefore it makes sense to control the "frame" in the details view but leave the design of the content to the implementations.

madoar:
I'm not so sure about that.
Take a look at the class `appPresentation` for example, it only provides a spacing, background color and offset for the details view. These are normally things we can standardize in the whole application. I don't believe we want to have multiple spacing standards etc. defined for normal formular like GUIs. If we're talking about special widgets like our ListWidgets, I agree with you, these things should contain their own styles, but not normal forms.

plata:
Isn't this what's implemented currently? Ok, the CSS class is called "containerConfigurationPane" which doesn't quite fit but apart from that...

madoar:
Yes, but take a look at the class definition for `DetailsView`. There we're adding the css classes `detailsPane` and `rightPane`. I don't think we need to additionally add `containerConfigurationPane` or for the "Apps" tab `appPresentation`. I would like to reduce the number of css classes added to the `DetailsView`

plata:
So you're saying I should simply remove the additional setting of "containerConfigurationPane"?

madoar:
Exactly. I would at least try how it looks.

plata:
Ok. One minute.

plata:
Like this?

madoar:
Exactly. Does it look good?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginePanel.java
```diff
@@ -104,7 +101,8 @@ private void populateContent() {
         progressSpacer.getStyleClass().add("engineSpacer");
         VBox.setVgrow(progressSpacer, Priority.NEVER);
 
-        this.setCenter(new VBox(informationContentPane, spacer, buttonBox, progressSpacer));
+        final VBox content = new VBox(informationContentPane, spacer, buttonBox, progressSpacer);
+        this.setCenter(content);
```
madoar: I believe it would be ok to initialize the VBox on the fly in `this.setCenter()`.

plata:
I didn't realize any change so I think it's ok.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginePanel.java
plata: Sure. I'll revert this.

madoar:
What about the other tabs? I'm asking because you've added the padding and spacing to the `.detailsPane` css class.

plata:
I looked at all. Maybe we will find out that we need a little adjustment somewhere but I didn't notice anything.

PR #1014, opened 2017-08-21
added RepositoryException

### LINKED ISSUE #919: Add RepositoryException which is a wrapper for a catchable Exception
Currently all exceptions are catched, that are thrown inside the application fetching process of a Repository.
This approach has the problem that errors, which are repository independent, are catched too.
To make this more clear, which exceptions can or should be thrown and processed inside a repository I suggest to introduce a new `RepositoryException` class.

For more details see #913

plata:
fixes #919

### DIFF
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryException.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryException.java
@@ -23,7 +23,7 @@ public class RepositoryException extends RuntimeException {
         super(message);
     }
 
-    public RepositoryException(String message, Exception e) {
-        super(message, e);
+    public RepositoryException(String message, Throwable cause) {
+        super(message, cause);
     }
 }
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryException.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryException.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2015-2017 PÂRIS Quentin
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryException.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.phoenicis.repository;
+
+public class RepositoryException extends RuntimeException {
+    public RepositoryException(String message) {
+        super(message);
+    }
+
+    public RepositoryException(String message, Exception e) {
```
madoar: I believe `public RepositoryException(String message, Throwable throwable)` is better.

plata: I just copied `CabException`. I guess @qparis had a reason to implement it like this. If you still think `Throwable` is better, I can change it of course.

plata:
I didn't see any place where this could be used in the other repository types. If you think I should use it in a particular place, please tell me where.

madoar:
I think we should throw it in the catch blocks of the `GitRepository`:
```java
        } catch (RepositoryNotFoundException | GitAPIException e) {
            LOGGER.error(String.format("Folder '%s' is no git-repository", this.localFolder.getAbsolutePath()), e);
        } catch (IOException e) {
            LOGGER.error("An unknown error occurred", e);
        } finally {
            // close repository to free resources
            if (gitRepository != null) {
                gitRepository.close();
            }
        }
```

For the `LocalRepository` I agree, it seems to be more useful to simply ignore single scripts, applications or categories, instead of throwing an exception, which skips the whole repository.

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryException.java
madoar: Take a look at the extended class https://docs.oracle.com/javase/7/docs/api/java/lang/RuntimeException.html. The constructor you call with `super(message, e);` takes a `Throwable` as its second argument.

PR #1015, opened 2017-08-21
moved default miniature URIs to ListWidgetEntry

### LINKED ISSUE #923: Find a better location for the default miniature URIs
See #907 for more details

plata:
ListWidgetEntry is the only place where they are used.

fixes #923


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/widgets/lists/ListWidgetEntry.java
```diff
@@ -28,6 +29,20 @@
 public class ListWidgetEntry<E> {
     private static final Logger LOGGER = LoggerFactory.getLogger(ListWidgetEntry.class);
 
+    private static URI DEFAULT_MINIATURE;
+    private static URI WINE_MINIATURE;
+    private static URI CONTAINER_MINIATURE;
+
+    static {
```
madoar: I would like to find a way to initialize the values without a static block. Do you have an idea?

plata: No, not really.

PR #1016, opened 2017-08-21
added documentation for MainWindowView

### LINKED ISSUE #758: Add JavaDoc/Documentation for the window makeup and its management
Currently we have no documentation, that describes how the POL 5 GUI is made up.
Therefore we should add a documentation that describes the elements that make up the POL 5 GUI, and how they work together.
In addition we should add JavaDoc to the MainWindowView class, because it makes up the core of the POL 5 GUI.
For more details see #737.

plata:
fixes #758

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/ViewEngines.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/ViewEngines.java
@@ -90,7 +90,7 @@ public class ViewEngines extends MainWindowView<EngineSidebar> {
 
         Bindings.bindContent(availableEngines.getTabs(), mappedSubCategoryTabs);
 
-        this.initializeSidebar();
+        this.setSidebar(this.sidebar);
     }
 
     public void setOnInstallEngine(Consumer<EngineDTO> onInstallEngine) {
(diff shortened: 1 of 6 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/ViewEngines.java
```diff
@@ -90,7 +90,7 @@ public ViewEngines(ThemeManager themeManager, String enginesPath) {
 
         Bindings.bindContent(availableEngines.getTabs(), mappedSubCategoryTabs);
 
-        this.initializeSidebar();
+        this.setSidebar(this.sidebar);
```
madoar: I think that it may be better to do it the other way around. I.e. we remove the `setSidebar` method and only provide the `initializeSidebar` method. The reason for this is, that I think we don't want to make the sidebar swapable from outside. In addition a setter method is in my eyes normally not used to directly set complete graphic components. This is normally done by `populate` or `initialize` methods.

plata: I'm not sure. `initializeSidebar` requires that you have knowledge about the internal structure of `MainWindowView` when deriving from it (i.e. that it has a field `sidebar`).

madoar: That's true. This leads us to the question, if we should provide a protected sidebar field, which can be used from the extending classes.

plata: Yep. But different topic.

PR #1017, opened 2017-08-21
fixed naming of style classes for list widgets

### LINKED ISSUE #814: Unify and rename the different css classes for the list widgets
Currently a lot of the css classes for the list widgets are named wrong. 
An example for this would be the css class `.iconListCell` inside the css class `.compactListWidget`, which should be renamed to `.compactListCell`.
These kinds of css classes should be renamed to more fitting names.

In addition we should try to define properties, that are or should be used for all types of list widgets, only once. 
An example here would be the background color for a selected element, which is currently only defined for an `IconsListElement`.

plata:
fixes #814


madoar:
This only solves the first part of #814, right?

plata:
Maybe. That's why you should not mix different topics in one issue. So please move the second part into a new issue (if you still want it).

PR #1018, opened 2017-08-21
remember list widget type selection

### LINKED ISSUE #850: Add a way to change the default list widget to the settings tab
Currently the default list widget at startup is the `IconsListWidget`. It would be nice if the user can change this inside the "Settings" tab.

plata:
fixes #850

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/settings/JavaFxSetting.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/settings/JavaFxSetting.java
@@ -20,8 +20,10 @@ package org.phoenicis.javafx.settings;
 
 public enum JavaFxSetting {
     THEME("application.theme"), SCALE("application.scale"), VIEW_SOURCE("application.viewsource"), WINDOW_HEIGHT(
-            "application.windowHeight"), WINDOW_WIDTH(
-                    "application.windowWidth"), WINDOW_MAXIMIZED("application.windowMaximized");
+            "application.windowHeight"), WINDOW_WIDTH("application.windowWidth"), WINDOW_MAXIMIZED(
+                    "application.windowMaximized"), APPS_LIST_TYPE("application.appsListType"), CONTAINERS_LIST_TYPE(
+                            "application.containersListType"), ENGINES_LIST_TYPE(
+                                    "application.enginesListType"), LIBRARY_LIST_TYPE("application.libraryListType");
 
     private final String propertyName;
 
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftListWidgetChooser.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftListWidgetChooser.java
@@ -7,14 +7,19 @@ import javafx.scene.control.ToggleGroup;
 import javafx.scene.layout.HBox;
 import org.phoenicis.javafx.views.common.widgets.lists.CombinedListWidget;
 import org.phoenicis.javafx.views.common.widgets.lists.ListWidgetType;
+import org.slf4j.LoggerFactory;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
+import java.util.function.Consumer;
 
 /**
  * Created by marc on 15.05.17.
  */
 public class LeftListWidgetChooser<E> extends HBox {
+    private final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(LeftListWidgetChooser.class);
+
     private List<CombinedListWidget<E>> listWidgets;
 
     private ToggleGroup toggleGroup;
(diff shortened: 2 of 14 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/settings/JavaFxSetting.java
```diff
@@ -20,8 +20,10 @@
 
 public enum JavaFxSetting {
     THEME("application.theme"), SCALE("application.scale"), VIEW_SOURCE("application.viewsource"), WINDOW_HEIGHT(
-            "application.windowHeight"), WINDOW_WIDTH(
-                    "application.windowWidth"), WINDOW_MAXIMIZED("application.windowMaximized");
+            "application.windowHeight"), WINDOW_WIDTH("application.windowWidth"), WINDOW_MAXIMIZED(
```
madoar: Can we please fix the format of this enum definition?
I believe an enum to be better readable, if one value is defined in each row.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftListWidgetChooser.java
```diff
@@ -7,14 +7,19 @@
...
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
+import java.util.function.Consumer;
 
 /**
  * Created by marc on 15.05.17.
  */
 public class LeftListWidgetChooser<E> extends HBox {
+    private final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(LeftListWidgetChooser.class);
```
madoar: Please use only `Logger` instead of `org.slf4j.Logger`

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftListWidgetChooser.java
```diff
@@ -23,6 +28,8 @@
     private ToggleButton compactListButton;
     private ToggleButton detailsListButton;
 
+    private Optional<Consumer<ListWidgetType>> onChoose;
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftListWidgetChooser.java
```diff
@@ -48,27 +57,54 @@ public LeftListWidgetChooser(List<CombinedListWidget<E>> listWidgets) {
...
         this.iconsListButton.addEventFilter(ActionEvent.ANY, filter);
 
         this.compactListButton = new ToggleButton();
         this.compactListButton.setToggleGroup(toggleGroup);
         this.compactListButton.getStyleClass().addAll("listIcon", "compactList");
         this.compactListButton
-                .setOnAction(event -> listWidgets.forEach(widget -> widget.showList(ListWidgetType.COMPACT_LIST)));
+                .setOnAction(event -> listWidgets.forEach(widget -> {
+                    widget.showList(ListWidgetType.COMPACT_LIST);
+                    this.onChoose.ifPresent(consumer -> consumer.accept(ListWidgetType.COMPACT_LIST));
```
madoar: Watch out, you're adding the `onChoose` call to the forEach loop. This means you will call it three times!

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftListWidgetChooser.java
```diff
@@ -48,27 +57,54 @@ public LeftListWidgetChooser(List<CombinedListWidget<E>> listWidgets) {
...
         this.compactListButton.addEventFilter(ActionEvent.ANY, filter);
 
         this.detailsListButton = new ToggleButton();
         this.detailsListButton.setToggleGroup(toggleGroup);
         this.detailsListButton.getStyleClass().addAll("listIcon", "detailsList");
         this.detailsListButton
-                .setOnAction(event -> listWidgets.forEach(widget -> widget.showList(ListWidgetType.DETAILS_LIST)));
+                .setOnAction(event -> listWidgets.forEach(widget -> {
+                    widget.showList(ListWidgetType.DETAILS_LIST);
+                    this.onChoose.ifPresent(consumer -> consumer.accept(ListWidgetType.DETAILS_LIST));
```
madoar: Watch out, you're adding the onChoose call to the forEach loop. This means you will call it three times!

phoenicis-javafx/src/main/java/org/phoenicis/javafx/settings/JavaFxSetting.java
plata: I have tried that several times but the formatter does not respect manual line breaks. If you know which setting we have to change so that this works, let me know (it respects it for functions).

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/LeftListWidgetChooser.java
plata: Oops. Copy paste error.

PR #1022, opened 2017-08-22
add "Installations" tab

### LINKED ISSUE #1020: add "Installations" tab
I propose to structure it similar to Steam, i.e. add an "Installations" tab which collects all running installations. We could then also show the progress for engine installations there. If you currently install an engine and ten show the details of another one, you completely loose the install progress/if anything went wrong.
If we do it like this, we can remove the "My apps" tab completely which will avoid the whole confusion in the first place.

plata:
fixes #1020 

open issues:
- search does not work (seems like `onSearch` is not set correctly)
- icon list/details list should use the correct miniature

I would like to handle the open issues separately later on (this PR is quite big already).

installations

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/installations/InstallationsController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/installations/InstallationsController.java
@@ -27,8 +27,8 @@ public class InstallationsController {
         this.viewInstallations = viewInstallations;
     }
 
-    public void setOnTabOpened(Runnable onTabOpened) {
-        this.viewInstallations.setOnTabOpened(onTabOpened);
+    public void setOnInstallationAdded(Runnable onInstallationAdded) {
+        this.viewInstallations.setOnInstallationAdded(onInstallationAdded);
     }
 
     public ViewInstallations getView() {
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/installations/InstallationsController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/installations/InstallationsController.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2015-2017 PÂRIS Quentin
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.javafx.controller.installations;
+
+import org.phoenicis.javafx.views.mainwindow.installations.ViewInstallations;
+
+public class InstallationsController {
+    private final ViewInstallations viewInstallations;
+
+    public InstallationsController(ViewInstallations viewInstallations) {
+        this.viewInstallations = viewInstallations;
+    }
+
+    public void setOnTabOpened(Runnable onTabOpened) {
+        this.viewInstallations.setOnTabOpened(onTabOpened);
+    }
+
+    public ViewInstallations getView() {
+        return viewInstallations;
+    }
+}
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsFilter.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsFilter.java
@@ -1,0 +1,69 @@
+package org.phoenicis.javafx.views.mainwindow.installations;
+
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.property.StringProperty;
+import javafx.collections.FXCollections;
+import javafx.collections.ListChangeListener;
+import javafx.collections.ObservableList;
+import javafx.collections.transformation.FilteredList;
+
+import java.util.function.BiPredicate;
+import java.util.function.Predicate;
+
+/**
+ * Created by marc on 29.03.17.
+ */
+public class InstallationsFilter<E> {
+    private StringProperty filterText;
+
+    private BiPredicate<String, E> filterTextMatcher;
+
+    private ObservableList<Predicate<E>> filters;
+
+    public InstallationsFilter(FilteredList<E> filteredList, BiPredicate<String, E> filterTextMatcher) {
+        this.filterTextMatcher = filterTextMatcher;
+
+        this.filterText = new SimpleStringProperty("");
+        this.filters = FXCollections.observableArrayList();
+
+        this.filterText.addListener((observableValue, oldValue, newValue) -> filteredList.setPredicate(this::filter));
+        this.filters.addListener(
+                (ListChangeListener<? super Predicate<E>>) change -> filteredList.setPredicate(this::filter));
+    }
+
+    public void setFilterText(String filterText) {
+        this.filterText.setValue(filterText);
+    }
+
+    public void addFilter(Predicate<E> filter) {
+        this.filters.add(filter);
+    }
+
+    public void setFilters(Predicate<E>... filters) {
+        this.filters.setAll(filters);
+    }
+
+    public void clearAll() {
+        this.filterText.setValue("");
+        this.filters.clear();
+    }
+
+    public void clearFilters() {
+        this.filters.clear();
+    }
+
+    public boolean filter(E value) {
+        boolean result = false;
+
+        if (filters.isEmpty()) {
+            result = true;
+        }
+
+        for (Predicate<E> filter : filters) {
+            result |= filter.test(value);
+        }
+
+        return result && filterTextMatcher.test(filterText.getValue(), value);
+    }
+
+}
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsPanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsPanel.java
@@ -19,6 +19,7 @@
 package org.phoenicis.javafx.views.mainwindow.installations;
 
 import org.phoenicis.javafx.views.common.widgets.lists.DetailsView;
+import org.phoenicis.javafx.views.mainwindow.installations.dto.InstallationDTO;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -29,41 +30,9 @@ final class InstallationsPanel extends DetailsView {
         super();
     }
 
-    /*public void setShortcutDTO(ShortcutDTO shortcutDTO) {
-        this.setTitle(shortcutDTO.getName());
-    
-        this.description.setText(shortcutDTO.getDescription());
-    
-        this.gridPane.getChildren().clear();
-    
-        try {
-            LOGGER.info("Reading shortcut: {}", shortcutDTO.getScript());
-    
-            final Map<String, Object> shortcutProperties = objectMapper.readValue(shortcutDTO.getScript(),
-                    new TypeReference<Map<String, Object>>() {
-                    });
-    
-            int i = 0;
-            for (Map.Entry<String, Object> entry : shortcutProperties.entrySet()) {
-                final Label keyLabel = new Label(tr(unCamelize(entry.getKey())) + ":");
-                keyLabel.getStyleClass().add(CAPTION_TITLE_CSS_CLASS);
-                GridPane.setValignment(keyLabel, VPos.TOP);
-                this.gridPane.add(keyLabel, 0, i);
-    
-                final Label valueLabel = new Label(entry.getValue().toString());
-                valueLabel.setWrapText(true);
-                this.gridPane.add(valueLabel, 1, i);
-    
-                i++;
-            }
-    
-        } catch (IOException e) {
-            LOGGER.warn("Could not parse shortcut script JSON", e);
-        }
-    
-        this.runButton.setOnMouseClicked(event -> onShortcutRun.accept(shortcutDTO));
-        this.stopButton.setOnMouseClicked(event -> onShortcutStop.accept(shortcutDTO));
-        this.uninstallButton.setOnMouseClicked(event -> onShortcutUninstall.accept(shortcutDTO));
-    }*/
+    public void setInstallationDTO(InstallationDTO installationDTO) {
+        this.setTitle(installationDTO.getName());
+        this.setCenter(installationDTO.getNode());
+    }
 
 }
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsPanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsPanel.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2015-2017 PÂRIS Quentin
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.javafx.views.mainwindow.installations;
+
+import org.phoenicis.javafx.views.common.widgets.lists.DetailsView;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class InstallationsPanel extends DetailsView {
+    private final Logger LOGGER = LoggerFactory.getLogger(InstallationsPanel.class);
+
+    public InstallationsPanel() {
+        super();
+    }
+
+    /*public void setShortcutDTO(ShortcutDTO shortcutDTO) {
+        this.setTitle(shortcutDTO.getName());
+    
+        this.description.setText(shortcutDTO.getDescription());
+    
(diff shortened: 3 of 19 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsFilter.java
```diff
@@ -0,0 +1,69 @@
...
+import javafx.collections.ObservableList;
+import javafx.collections.transformation.FilteredList;
+
+import java.util.function.BiPredicate;
+import java.util.function.Predicate;
+
+/**
+ * Created by marc on 29.03.17.
+ */
+public class InstallationsFilter<E> {
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsPanel.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package org.phoenicis.javafx.views.mainwindow.installations;
+
+import org.phoenicis.javafx.views.common.widgets.lists.DetailsView;
+import org.phoenicis.javafx.views.mainwindow.installations.dto.InstallationDTO;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class InstallationsPanel extends DetailsView {
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsUtils.java
```diff
@@ -0,0 +1,131 @@
...
+        return categories;
+    }
+
+    /**
+     * removes as installation from an existing list of installations
+     * @param list existing list of installations
+     * @param toRemove installation which shall be removed
+     * @return new list of installations containing the existing installations without the installation which shall be removed
+     */
+    public List<InstallationCategoryDTO> removeInstallationFromList(List<InstallationCategoryDTO> list,
```
madoar: I think this cries for a streaming API solution

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
```diff
@@ -0,0 +1,169 @@
...
+import org.phoenicis.javafx.views.mainwindow.installations.dto.InstallationDTO;
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+import static org.phoenicis.configuration.localisation.Localisation.tr;
+
+public class ViewInstallations extends MainWindowView<InstallationsSidebar> {
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private Consumer<InstallationDTO> onInstallationSelected = installation -> {
+    };
+
+    public ViewInstallations(ThemeManager themeManager,
+            JavaFxSettingsManager javaFxSettingsManager) {
+        super(tr("Installations"), themeManager);
+        this.getStyleClass().add("mainWindowScene");
+
+        activeInstallations = new CombinedListWidget<>(ListWidgetEntry::create, (selectedItem, event) -> {
```
madoar: To fit the style, please add a `this` at the beginning of the row

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationCategoryDTO.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+/**
+ * Represents a category of application
+ */
+@JsonDeserialize(builder = InstallationCategoryDTO.Builder.class)
+@Translatable
+public class InstallationCategoryDTO {
```
madoar: How does an `InstallationCategoryDTO` differ from a normal `CategoryDTO`?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationDTO.java
```diff
@@ -0,0 +1,202 @@
...
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.URI;
+import java.util.Comparator;
+
+@JsonDeserialize(builder = InstallationDTO.Builder.class)
+public class InstallationDTO {
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationDTO.java
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        name = builder.name;
+        category = builder.category;
+        description = builder.description;
+        miniature = builder.miniature;
+        node = builder.node;
+    }
+
+    public enum InstallationType {
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsUtils.java
plata: Can you tell me how to do it with streaming API? (I'm not that familiar with it).

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsFilter.java
plata: Does it even make sense to have this class? It's just a copy of `ShortcutFilter`. I could also rename `ShortcutFilter` to `SimpleFilter` and reuse it.

madoar: You're right. It's better to implement it only one time. When renaming the class, please don't forget to add some JavaDoc :)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
```diff
@@ -0,0 +1,169 @@
...
+        this.installations = new ExpandedList<>(this.sortedCategories, InstallationCategoryDTO::getInstallations);
+        this.filteredInstallations = new FilteredList<>(this.installations);
+        this.sortedInstallations = this.filteredInstallations.sorted(Comparator.comparing(InstallationDTO::getName));
+
+        this.filter = new InstallationsFilter<>(filteredInstallations,
+                (filterText, installation) -> installation.getName().toLowerCase().contains(filterText));
+
+        this.activeInstallations.setOnMouseClicked(event -> {
+            this.activeInstallations.deselectAll();
+            this.onInstallationSelected.accept(null);
```
madoar: Are you sure we want to pass `null` to a function?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
```diff
@@ -0,0 +1,169 @@
...
+        this.installationsPanel.setMaxWidth(600);
+        this.showDetailsView(this.installationsPanel);
+    }
+
+    /**
+     * adds new installation
+     * @param installationDTO new installation
+     */
+    public void addInstallation(InstallationDTO installationDTO) {
+        populate(new InstallationsUtils().addInstallationToList(this.categories, installationDTO));
```
madoar: Is it really required to repopulate the whole list widget when adding a new installation?
I think it would be better to only add the installation to the category and trigger an update.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationCategoryDTO.java
plata: It contains `InstallationDTOs` instead of `ApplicationDTOs`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
```diff
@@ -0,0 +1,169 @@
...
+        });
+        this.onInstallationAdded.run();
+    }
+
+    /**
+     * removes installation (if it exists)
+     * @param installationDTO installation to be removed
+     */
+    public void removeInstallation(InstallationDTO installationDTO) {
+        populate(new InstallationsUtils().removeInstallationFromList(this.categories, installationDTO));
```
madoar: Same as for `addInstallation`

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationDTO.java
```diff
@@ -61,16 +65,19 @@ private InstallationDTO(Builder builder) {
...
+    /**
+     * type of the installation
+     */
     public enum InstallationType {
-        APPS, WINE;
+        APPS, ENGINES;
 
         @Override
         public String toString() {
             switch (this) {
```
madoar: As an alternative you could add the display name of the installation types as a parameter to the enum, then you would simply need to return this parameter.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
plata: This is just a copy from `ViewLibrary`. I didn't really think aobut it.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
plata: I don't know how this could be implemented.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
plata: I don't know how this could be implemented.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationDTO.java
plata: Yes but I don't see a big advantage currently.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
madoar: It should be enough to update modified element in the `ObservableList` `categories`. 
If you modify the content of this list, you should automatically trigger an update for the modified element.

plata: But I cannot modify the `InstallationCategoryDTOs` in that list. The list with the `InstallationDTOs` is final.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationDTO.java
madoar: It's shorter and easier extendable. Let's say you want to add a new enum type later on, then it's possible you forget to add the new case to the `toString` method. If you modify the enum to take a display string with a constructor this can't happen.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
madoar: final as in immutable? 
If the list is not immutable you should be able to call `add`, `set` and `remove` right?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationDTO.java
plata: What will happen if someone puts a typo in the display string? Than it might suddenly not be translated anymore etc.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
madoar: In addition I'm talking about the observable list. It's ok to build a new `InstallationCategoryDTOs` object and update it in the `ObservableList`. I just want to prevent us from deleting **all** `InstallationCategoryDTOs` and adding them again.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationDTO.java
madoar: Yes, the same can happen with the current approach, right?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
plata: No, I cannot `add` or `remove` (that was also my first approach). We will usually have 1, max. 2 `InstallationCategoryDTOs`. So there's no real difference if I delete it or just change it.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/dto/InstallationDTO.java
plata: No because you can only decide which enum you use. So this really restricts the space for errors.

madoar: My approach is the same.

```java
public enum InstallationType {
   APPS("Apps"), ENGINES("Engines");

   private String displayName;   

   InstallationType(String displayName) {
      this.displayName = displayName;
   }

   public String toString() {
      return this.displayName;
   }
}
```

plata: Ah ok. Now I understand. I thought you wanted to set it from outside...

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/ViewInstallations.java
madoar: I think we may be talking about different lists.
Because the `ObservableList` categories must be modifiable, otherwise we wouldn't be able to set its values.

plata: Yes, I can add/remove in that list. But that doesn't help. I have to change the list of `InstallationDTO` inside the `InstallationCategoryDTO`. So I could create a new `InstallationCategoryDTO` and add it to the `ObservableList` but like I explained above, this doesn't make much sense as the `ObservableList` will usually contain only 1 item.

madoar:
I believe we can go with this, but I would like to get the input of @qparis to see if this fits in POL/POM 5.

plata:
Ok. In the meantime: Could you checkout the branch and have a look if you get the search working? I think you know that code pretty well.

madoar:
In a future PR, I think we should modify the list to contain the status of the installation. I.e. is it still in the input phase, where you need to press a lot of continues or is it already in the main installation phase, with a progress bar etc.

madoar:
What exactly should be searched?

plata:
It should work exactly like for shortcuts: If you have e.g. an installation for Notepad++ and one for 7-zip and then search for "note", it should show only the Notepad++ installation.

plata:
To enable miniatures, change the `setupWizard` creation in JS to:
```Javascript
var appsManager = Bean("repositoryManager");
var application = appsManager.getApplication([this._type, this._category, this._name]);
var setupWizard = SetupWizardWithMiniature(this._name, application.getMainMiniature().get());
```

miniature

madoar:
I'm not sure I see the difference between miniature and without?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/installations/InstallationsController.java
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.javafx.controller.installations;
+
+import org.phoenicis.javafx.views.mainwindow.installations.ViewInstallations;
+
+public class InstallationsController {
```
madoar: This class looks like a wrapper class for the `ViewInstallations` instance. I'm not sure this class is useful in this case, because it simply forwards all calls to the `ViewInstallations` instance. I would suggest to use the `ViewInstallations` instance directly and skip the controller.

madoar:
This should have the same reason as #1026 (#1025)

plata:
There's no screenshot without miniature anymore (I have updated the PR to show the current state). Without the miniature, you didn't have the Notepad++ miniature but just the default one.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/installations/InstallationsController.java
plata: I wanted to have so that the same structure is used everywhere. Maybe we decide to have a module `phoenicis-installations` with the dtos and an `InstallationsManager` or something like that. Then it will be good to have the controller.

phoenicis-cli/src/main/java/org/phoenicis/cli/scriptui/SetupUiFactoryCLI.java
```diff
@@ -30,7 +32,7 @@ public SetupUiFactoryCLI() {
     }
 
     @Override
-    public SetupUi createSetupWindow(String title) {
+    public SetupUi createSetupWindow(String title, URI miniature) {
```
madoar: JavaDoc

PR #1029, opened 2017-08-23
Add a new class DelayedFilterConsumer

madoar:
And move `onSearch` and `onClear` callbacks to the `ApplicationSidebar` class


madoar:
This approach doesn't force us to have a delay in the `SearchBar`, I like this more than adding `pause` to the `SearchBar` directly.

PR #1031, opened 2017-08-23
fixed handling of FailurePanel

### LINKED ISSUE #740: Show error message when something during the repository fetching goes wrong
We need a pop-up when something goes wrong while fetching the repository. This pop-up should then show the problem. Could be several things:

- no Internet connection
- folder does not exist
- folder is not a git repository
- ...

See #663 for more context information

plata:
set notification and failure reason whenever FailurePanel is shown

fixes #740

failure panel

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
@@ -42,6 +42,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Consumer;
 
 import static org.phoenicis.configuration.localisation.Localisation.tr;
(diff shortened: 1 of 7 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
```diff
@@ -64,7 +65,7 @@ public EnginesController(ViewEngines viewEngines, RepositoryManager repositoryMa
         this.themeManager = themeManager;
 
         this.repositoryManager.addCallbacks(this::populateView,
-                e -> Platform.runLater(() -> viewEngines.showFailure()));
+                e -> Platform.runLater(() -> viewEngines.showFailure(tr("Loading engines failed."), Optional.of(e))));
```
madoar: Just asking:
Am I overlooking something or are you always using `Optional.of(e)`? If this is the case why are we using an `Optional`, `Optional.of(null)` will throw an error?

plata: Yes, I'm currently using `Optional.of(e)` everywhere. However, I wanted to have the possibility to call it with `Optional.empty()`. It might happen that you want to show a `FailurePanel` but do not have an exception.

PR #1049, opened 2017-08-24
clean-up (use Apache StringUtils and formatter changes)

plata:


plata:
Why should I?

madoar:
To format the code again?

plata:
Hopefully that should not be necessary (if everybody built with `clean package` before committing).

madoar:
I sometimes forget it...

Issue #1010, opened 2017-08-21
Error when installing Notepad++


madoar:
When installing notepad++ I get the following exception:

```
[DEBUG] org.phoenicis.tools.config.CompatibleConfigFileFormat (l.88) - Error while reading the file. Will assume that the config file is EMPTY
java.io.FileNotFoundException: /home/marc/.Phoenicis/containers/wineprefix/Notepad/phoenicis.cfg (Datei oder Verzeichnis nicht gefunden)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at com.fasterxml.jackson.core.JsonFactory.createParser(JsonFactory.java:753)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2673)
	at org.phoenicis.tools.config.CompatibleConfigFileFormat.getMap(CompatibleConfigFileFormat.java:77)
	at org.phoenicis.tools.config.CompatibleConfigFileFormat.writeValue(CompatibleConfigFileFormat.java:53)
	at jdk.nashorn.internal.scripts.Script$Recompilation$68$2538A$\^eval\_.prefix([Engines, Wine, Engine, Object]:128)
	at jdk.nashorn.internal.scripts.Script$Recompilation$46$611$\^eval\_.go([Engines, Wine, QuickScript, InstallerScript]:72)
	at jdk.nashorn.internal.scripts.Script$24$\^eval\_.:program(<eval>:3)
	at jdk.nashorn.internal.runtime.ScriptFunctionData.invoke(ScriptFunctionData.java:637)
	at jdk.nashorn.internal.runtime.ScriptFunction.invoke(ScriptFunction.java:494)
	at jdk.nashorn.internal.runtime.ScriptRuntime.apply(ScriptRuntime.java:393)
	at jdk.nashorn.api.scripting.NashornScriptEngine.evalImpl(NashornScriptEngine.java:449)
	at jdk.nashorn.api.scripting.NashornScriptEngine.evalImpl(NashornScriptEngine.java:406)
	at jdk.nashorn.api.scripting.NashornScriptEngine.evalImpl(NashornScriptEngine.java:402)
	at jdk.nashorn.api.scripting.NashornScriptEngine.eval(NashornScriptEngine.java:155)
[...]

PR #1052, opened 2017-08-25
added libxext6:i386 dependency for .deb

plata:
see #1010

### DIFF
--- a/phoenicis-dist/src/deb/control/control
+++ b/phoenicis-dist/src/deb/control/control
@@ -10,4 +10,4 @@ Description: This program is a front-end for wine.
              Copyright 2011 PlayOnLinux team <contact@playonlinux.com>
 Maintainer: PlayOnLinux Packaging <packages@playonlinux.com>
 Homepage: www.phoenicis.org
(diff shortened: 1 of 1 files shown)


TriMoon:
Does this mean POL5 will be 32bits only?
I was hoping on a version for 64bit...

plata:
The dependency is a runtime dependency which is needed by Wine. It doesn't say anything about Phoenicis PlayOnLinux/PlayOnMac (which is written in Java, so 32/64bit really doesn't matter that much).

TriMoon:
@plata there also is a wine 64bit IIRC.
Besides It matters for those on 64bit linux, because the package won't even show-up in the package listing.
fe. i just checked while using 64bit Debian 9 (testing), and it isn't listed, and i'm not going to install 32bit architecture aside on a 64bit only system...

plata:
The thing is that the used Wine architecture (x86/amd64) depends on the installation script and not on the Phoenicis software. As many scripts are using 32bit Wine, adding this dependency to the .deb is an easy way to ensure that the average user can use the scripts without installing the dependency manually. I agree that it's not optimal, so if you have any better ideas, let me know.

Issue #1053, opened 2017-08-25
text in details view unreadable


plata:
![details text](https://user-images.githubusercontent.com/3178485/29640419-ceee5944-8856-11e7-924f-6e1a1adddc6f.png)

The text is cut off but it should be wrapped. Seems like the label is wider than the `DetailsView`.

madoar:
It looks like the toggle buttons in the sidebar don't have the correct height too, right?
In addition the sidebar seems to be cutoff too (see All Operating System).

plata:
The problem is that the details pane is never smaller than it's preferred width even if it does not fit in the window. I have no idea how to solve this.

PR #1057, opened 2017-08-25
fixed height of sidebar buttons

plata:
see #1053

before
before

now
now


madoar:
Does this effect only occur in the unity theme?

plata:
I didn't check.

madoar:
This also occurs with the default theme.

plata:
Ok. I will change it.

PR #1062, opened 2017-08-26
show engine installations in "Installations" tab

plata:
engine installation


madoar:
Just asking:
Did you test, that the cli still works afterwards?

plata:
No. Do we even have a command to install an engine from CLI? (I never did that)

plata:
Just installed Notepad++ with Phoenicis CLI. It calls the Wine installation implicitly (I deleted the Wine version before starting the installation). I guess that's enough to prove that it's still working.

PR #1064, opened 2017-08-27
unified setup wizards

### LINKED ISSUE #1063: unify setup wizards
The wizards `SetupWizard`, `SetupWizardWithMiniature` and `EngineProgressUi` in `SetupWizardInjector` can be unified to one common wizard.

plata:
fixes #1063

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/SetupWizardInjector.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/SetupWizardInjector.java
@@ -8,7 +8,6 @@ import org.phoenicis.scripts.wizard.UiSetupWizardImplementation;
 import java.net.URI;
 import java.util.Optional;
 import java.util.function.BiFunction;
-import java.util.function.Function;
 
 /**
  * Injects SetupWizard() function into a script engine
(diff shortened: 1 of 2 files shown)


madoar:
Can't we use `Java.type(...)` to hide the package path?
For more details take a look at https://docs.oracle.com/javase/8/docs/technotes/guides/scripting/nashorn/api.html.

plata:
We could do that but then we would need a variable somewhere in the script to hold this type. I don't think that's very obvious for people who just want to provide a script.

madoar:
You could provide the type like you provide `SetupWizardApps`, right?

plata:
Yes, could be possible. What do I do about the `Function` (I would need something like a `TriFunction`)?

plata:
```
var type = Java.type(org.phoenicis.scripts.ui.InstallationType);
```
Doesn't work (not even from JS). It gives `ClassNotFoundException: [JavaClass org.phoenicis.scripts.ui.InstallationType]`.

madoar:
Is it possible that you're missing some `"`?
```javascript
var InstallationType = Java.type("org.phoenicis.scripts.ui.InstallationType");
```

plata:
You're right...

phoenicis-scripts/src/main/java/org/phoenicis/scripts/nashorn/builtins/SetupWizardInjector.java
```diff
@@ -21,6 +21,8 @@ public SetupWizardInjector(UiSetupWizardFactory uiSetupWizardFactory) {
 
     @Override
     public void injectInto(NashornEngine nashornEngine) {
+        nashornEngine.eval("var InstallationType = Java.type(\"org.phoenicis.scripts.ui.InstallationType\")",
```
madoar: I think it would be better to get the package at runtime, in case we move it later on.
Can we get this information through the `Class` object?

PR #1068, opened 2017-08-27
Refactor filters + add filters for Library and Installation tabs

### LINKED ISSUE #1026: Search bar in Library calls empty (old) onSearch method
After entering a search term inside the Searchbar of the "Library" tab the wrong `onSearch` instance is used. To be more exact, `onSearch` is always `null`, because `setOnSearch(Consumer<String> onSearch)` can't be called before the constructor of the class and therefore before `this.searchBar = new SearchBox(onSearch, () -> {});`.

madoar:
This PR fixes #1026.

Things left to do:
- Modify `EnginesFilter` to extend `AbstractFilter`

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/LibraryFilter.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/LibraryFilter.java
@@ -1,0 +1,94 @@
+package org.phoenicis.javafx.views.mainwindow.library;
+
+import org.phoenicis.javafx.views.AbstractFilter;
+import org.phoenicis.library.dto.ShortcutCategoryDTO;
+import org.phoenicis.library.dto.ShortcutDTO;
+
+import java.util.Optional;
+
+/**
+ * Filter class for the "Library" tab
+ *
+ * @author Marc Arndt
+ */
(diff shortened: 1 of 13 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/library/LibraryFilter.java
```diff
@@ -0,0 +1,98 @@
...
+     * The entered search term.
+     * If no search term has been entered, this value is {@link Optional#empty()}.
+     */
+    private Optional<String> searchTerm;
+
+    /**
+     * The selected shortcut category.
+     * If no shortcut category has been selected, this value is {@link Optional#empty()}.
+     */
+    private Optional<ShortcutCategoryDTO> selectedShortcutCategory;
```
plata: Isn't this really the same as for Installations? Would it be possible to do something like:
```Java
public class FilterWithCategory<T> extends AbstractFilter {
...
    private Optioinal<T> selectedCategory;
```

madoar: The only similar thing are the fields (and method names) inside the classes.
The contents of the filter methods differs, if only in one word.

PR #1069, opened 2017-08-28
added possibility to configure new shortcut when creating it

### LINKED ISSUE #1060: Add a way to manually add a new shortcut
Now that it's possible to update/change a shortcut, we should also add a way to manually add new shortcuts to POL/POM 5.

plata:
requires PhoenicisOrg/Scripts#376

When you click "Create shortcut" in the library sidebar, a new details view opens:
details

If you do not fill the form properly, it marks the wrong fields and shows an error message in a tooltip:
error

fixes #1060

### DIFF
--- a/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/default/main.less
+++ b/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/default/main.less
@@ -602,6 +602,10 @@
   -fx-font-family: "Roboto Light";
 }
 
+.input:error {
+  -fx-background-color: #f31111;
+}
+
 .buttonWithIcon {
   -fx-background-color: transparent;
   -fx-label-padding: 0.2em 0 0.2em 2.2em;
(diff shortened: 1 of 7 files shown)


madoar:
I think it would be more intuitive if we add a red background or border to the textfields that are wrong.
In addition we could show the error message as a tooltip.
What do you think?

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/default/main.less
```diff
@@ -602,6 +602,10 @@
   -fx-font-family: "Roboto Light";
 }
 
+.input:error {
+  -fx-background-color: #f31111;
```
madoar: Just asking:
Maybe a red border looks better?

plata: I tried but it doesn't change the border color of the existing border. Instead it adds a new rectangular border around it. Looks really strange.

plata: Ah wait. I used the wrong field in the CSS.

PR #1070, opened 2017-08-28
set minimum window size

### LINKED ISSUE #1051: Window sizing
screenshot from 2017-08-23 22-55-18

I'm not sure about your sizing of the right panel here when the Phoenicis window is made smaller. Here the text gets cut off in the right-hand panel, would it not just be better to auto-resize so the text wraps? Then you should probably prevent the user from making the window any smaller when it starts hitting the bottom image (or allow the user to go a bit further by auto-resizing the bottom image).

Also, you can see that the entries in the top panel are cut off - I like how you've dealt with this by enabling the scroll wheel to scroll and providing a drop-down button, you may want to have small left-right arrow buttons like Firefox does when lots of tabs are open.

screenshot from 2017-08-23 23-04-55

But you need to set a minimum size because at the moment you can make ridiculous windows like

screenshot from 2017-08-23 23-04-15
screenshot from 2017-08-23 23-04-23
screenshot from 2017-08-23 23-04-28

Most programs don't allow the user to do this. And it looks very bad that when resizing too low you just end up cutting off most of the UI (at the very least, the user should be able to scroll across the UI elements to be able to access all of the UI elements) but

plata:
avoid 1x1 pixel window

fixes #1051

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindow.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/MainWindow.java
@@ -70,6 +70,9 @@ public class MainWindow extends Stage {
         this.getIcons().add(new Image(
                 JavaFXApplication.class.getResourceAsStream("/org/phoenicis/javafx/views/common/phoenicis.png")));
 
+        // avoid 1x1 pixel window
+        this.setMinHeight(200);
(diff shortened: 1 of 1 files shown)


madoar:
Can/Should we set this in the css theme?

plata:
I don't think it makes sense. The only purpose is really to avoid windows with 1x1 pixel.

PR #1080, opened 2017-08-28
deleted unused SimpleFilter

plata:


madoar:
I'm not sure if the class `SimpleFilter` is still used/required?

PR #1087, opened 2017-08-30
avoid losing icon and miniature when editing shortcut

plata:
If the icon and miniature did not change, they were deleted in deleteShortcut.

### DIFF
--- a/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
+++ b/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
@@ -29,6 +29,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.util.function.Consumer;
 
 @Safe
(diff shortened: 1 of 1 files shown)


madoar:
What do you mean with "did not change"?

plata:
If you do not choose a different miniature, the miniature URI points to the .miniature in shortcuts. This file is deleted in `deleteShortcut`. So you do not have a miniature anymore after editing the shortcut.

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
```diff
@@ -155,7 +156,46 @@ public void deleteShortcut(ShortcutDTO shortcutDTO) {
...
+                LOGGER.error("Could not backup miniature.");
+            }
+        }
+
         deleteShortcut(shortcutDTO);
         createShortcut(shortcutDTO);
+
+        // delete backups
+        if (iconBackup.delete()) {
+            iconBackup.delete();
```
madoar: I'm not sure I understand these if else blocks, you delete a file after it has been successfully deleted?

PR #1090, opened 2017-08-30
Improve the performance of PhoenicisFilteredList

### LINKED ISSUE #1076: Improve the performance of PhoenicisFilteredList
As far as I can tell we currently have no performance issues with this class, but the more scripts are added to POL/POM 5, the higher the probability to see some performance issues for this class.

The reason for this is the current implementation of:

- `int size()`
- `int getSourceIndex(int index)`, which is used in `E get(int index)` 

Both methods require the iteration through the whole list. Especially in the case of `get(index)`, this may be problematic, because you'll need a lot of iterations to get all elements inside the list if you need to get all elements contained.

For this reason I suggest to change the implementation to improve its performance.

madoar:
Fixes #1076. The `trigger` method is now based on the `update` method in `FilteredList`, with the only changes that it checks the whole list for updates and doesn't call `nextUpdate` if the value fulfilled the filter condition before and after the check.

In addition I've fixed the permutation range in some `permute` methods.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/lists/AdhocList.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/lists/AdhocList.java
@@ -60,7 +60,7 @@ public class AdhocList<E> extends PhoenicisTransformationList<E, E> {
                 perm[i + others.length] = c.getPermutation(i) + others.length;
             }
 
-            nextPermutation(from, to, perm);
+            nextPermutation(others.length + from, others.length + to, perm);
         }
     }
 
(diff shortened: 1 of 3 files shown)


plata:
Did you somehow verify that this really improves the performance?

madoar:
No I've done no real tests. The advantages of the new version are:

- the implementation is based on the one used for the "normal" `FilteredList` provided in javafx
- the implementation uses an array to store the filtered indexes. This makes the access through the  `get()` method very fast
- the implementation stores the size of the list in a separate field, which makes its access quite fast too

What may be slower are the change event operations. These events require some changes made to the filtered array. This shouldn't be a large problem for us, because I believe that we are accessing the values in the list more often than we change them.

plata:
Ok. As long as it works, it's fine for me. I was just asking because of the rules you probably know:
1. Don't optimize.
2. Don't optimize yet.

madoar:
What rule do you mean?

plata:
The two I wrote there (see <http://wiki.c2.com/?RulesOfOptimization>). It basically means that you should profile if you optimize because you might be wrong with your assumptions about the runtime behavior.

PR #1091, opened 2017-08-31
fixed JavaFX error message dialog size and set owner

### LINKED ISSUE #1089: Error screen doesn't automatically expand outwards to display details
screenshot from 2017-08-30 16-09-01
screenshot from 2017-08-30 16-09-05

plata:
fixes #1089 

normal

expanded

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/ErrorMessage.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/ErrorMessage.java
@@ -23,31 +23,62 @@ import javafx.scene.control.Label;
 import javafx.scene.control.TextArea;
 import javafx.scene.layout.GridPane;
 import javafx.scene.layout.Priority;
+import javafx.scene.layout.Region;
+import javafx.stage.Window;
 import org.apache.commons.lang.exception.ExceptionUtils;
+import org.phoenicis.javafx.views.mainwindow.ui.MainWindowView;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static org.phoenicis.configuration.localisation.Localisation.tr;
 
+/**
+ * error message dialog
+ */
 public class ErrorMessage {
     private final Logger LOGGER = LoggerFactory.getLogger(ErrorMessage.class);
     private final Alert alert;
 
+    /**
+     * constructor
+     * @param message error message
+     * @param exception exception which caused the error
+     */
     public ErrorMessage(String message, Exception exception) {
-        alert = new Alert(Alert.AlertType.ERROR);
-        alert.setTitle(tr("Error"));
-        alert.setHeaderText(message);
+        this(message, exception, (Window) null);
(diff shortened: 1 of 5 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/ErrorMessage.java
```diff
@@ -18,36 +18,69 @@
...
+import org.phoenicis.javafx.views.mainwindow.ui.MainWindowView;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static org.phoenicis.configuration.localisation.Localisation.tr;
 
+/**
+ * error message dialog
+ */
 public class ErrorMessage {
```
madoar: Why don't we let this class extend `Alert`?

plata: I'm not sure. I guess to restrict the possible interfaces such that we guarantee a common look and feel of the error dialogs.

madoar: I'm not sure I can follow you. As far as I understand the class, it's a wrapper for the `Alert` instance inside it. Therefore we should be able to remove the wrapper code by making it extend `Alert` directly, right? 
This would change `alert.something` to `this.something` and we can remove our `show()` method

plata: Yes, but if we subclass `Alert`, you can use all methods from `Alert`. So you might change the look which could result in different error messages in different parts of the software. With the current implementation you cannot do this. Therefore it ensures that the error message dialog looks the same everywhere.

madoar: Are you sure this is a big problem? I mean basically only developers working on POL/POM 5 are able to do that, and all changes that are made to the main repository (i.e. this one) are reviewed. Everything else is of no importance to us, right? I don't think this is such a big problem.

plata: I don't think it's a big problem. I just said that this probably the reason why it was chosen not to subclass `Alert`.

madoar: An alternative would be to let `ErrorMessage` extend `Alert` and make the constructor `private`.
Then we could implement a static `show()` method, that can be called like this:

```java
ErrorMessage.show(message, exception, window);
```

The result of the method would be void, so none of the fields inside `ErrorMessage` can be changed from outside.

plata: To be honest, I don't really see the point of changing the current implementation. There's no advantage if we subclass.

PR #1095, opened 2017-08-31
Change the implementation of EnginesFilter to extend AbstractFilter

### LINKED ISSUE #1072: Modify EnginesFilter to extend AbstractFilter
The implementation of the `EnginesFilter` class for the engines tab, should be changed to extend `AbstractFilter`. This process contains the change of the `FilteredList` instances to `PhoenicisFilteredList` instances.

madoar:
Fixes #1072

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginesFilter.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginesFilter.java
@@ -123,12 +123,15 @@ public class EnginesFilter extends AbstractFilter {
     public Predicate<EngineVersionDTO> createFilter(EngineCategoryDTO engineCategory,
             EngineSubCategoryDTO engineSubCategory) {
         return engineVersion -> {
-            return searchTerm
+            final boolean containsSearchTerm = searchTerm
                     .map(searchTerm -> engineVersion.getVersion().toLowerCase().contains(searchTerm.toLowerCase()))
-                    .orElse(true)
-                    && ((this.showInstalled.getValue() && isInstalled(engineCategory, engineSubCategory, engineVersion))
-                            || (this.showNotInstalled.getValue()
-                                    && !isInstalled(engineCategory, engineSubCategory, engineVersion)));
+                    .orElse(true);
+            final boolean fulfillsShowInstalled = this.showInstalled.getValue()
+                    && isInstalled(engineCategory, engineSubCategory, engineVersion);
+            final boolean fulfillsShowNotInstalled = this.showNotInstalled.getValue()
+                    && !isInstalled(engineCategory, engineSubCategory, engineVersion);
+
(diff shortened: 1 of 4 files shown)


madoar:
The approach in this PR may introduce some performance issues later on, when POL contains more than only wine engines.

The reason for this is, that every time the user changes the selected engine type new `EngineSubCategoryTab` objects are created. Each of these new objects will call `filter.addOnFilterChanged(filteredEngineVersions::trigger);`, which in turn can lead to a lot of filter trigger calls when a search term has been entered.

I believe that we can use this PR for now, but we maybe should create a new issue to remind us that we need to either:
- add a way to remove the added filter changed trigger, when the `EngineSubCategoryTab` gets removed from the list, or
- create all available `EngineSubCategoryTab` objects, for all available `EngineCategoryTab` objects, during the instantiation of the `EnginesView`, then we would only need to switch between the tabs later on.

plata:
I prefer the second option. Maybe we can store the `EngineSubCategories` in a `PhoenicisFilteredList`?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginesFilter.java
```diff
@@ -64,27 +106,42 @@ public EnginesFilter(EngineCategoryDTO engineCategory, EngineSubCategoryDTO engi
...
+     * @param engineCategory    The engine category
+     * @param engineSubCategory The engine sub category
+     * @return A new filter predicate
+     */
+    public Predicate<EngineVersionDTO> createFilter(EngineCategoryDTO engineCategory,
+            EngineSubCategoryDTO engineSubCategory) {
+        return engineVersion -> {
+            return searchTerm
+                    .map(searchTerm -> engineVersion.getVersion().toLowerCase().contains(searchTerm.toLowerCase()))
+                    .orElse(true)
```
plata: Can this be formatted differently?

madoar: How would you like it to look?
This is how maven formats it

plata: I think it should respect manual line breaks.

I'm not quite sure but currently it's a bit hard to see what belongs together. Maybe somehow like this:
```Java
return searchTerm.map(
    searchTerm -> engineVersion.getVersion().toLowerCase().contains(searchTerm.toLowerCase())).orElse(true)
    && ((this.showInstalled.getValue() && isInstalled(engineCategory, engineSubCategory, engineVersion))
    || (this.showNotInstalled.getValue() && !isInstalled(engineCategory, engineSubCategory, engineVersion))
);

plata: Maybe the better solution would be to create boolean variables for the different parts of the expression.

PR #1103, opened 2017-09-01
fixed callback for fetch engines

### LINKED ISSUE #963: fix callback for fetch engines
see #961

plata:
Tested with a renamed copy of Wine (in JS).

fixes #963

### DIFF
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/EnginesSource.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/EnginesSource.java
@@ -49,20 +49,22 @@ public class EnginesSource {
     public void fetchAvailableEngines(List<CategoryDTO> categoryDTOS, Consumer<List<EngineCategoryDTO>> callback) {
         final InteractiveScriptSession interactiveScriptSession = scriptInterpreter.createInteractiveSession();
 
-        List<EngineCategoryDTO> engines = new ArrayList<>();
+        List<EngineCategoryDTO> engines = Collections.synchronizedList(new ArrayList<>());
         for (CategoryDTO categoryDTO : categoryDTOS) {
             final String engineName = categoryDTO.getName();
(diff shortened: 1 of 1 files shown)


phoenicis-engines/src/main/java/org/phoenicis/engines/EnginesSource.java
```diff
@@ -49,20 +49,22 @@ public EnginesSource(ScriptInterpreter scriptInterpreter, ObjectMapper objectMap
...
+                                        .withName(engineName)
+                                        .withDescription(engineName)
+                                        .withSubCategories(unSerialize(output))
+                                        .build();
+                                engines.add(engineCategoryDTO);
                             }, this::throwError),
                     this::throwError);
         }
-
+        callback.accept(engines);
```
madoar: Is it always the case that the callback method is called **after** the `interactiveScriptSession` eval calls have been successfully terminated?

plata: You are right. It doesn't work like this.

madoar: Maybe we should add a synchronous way to evaluate a script?
Afterwards we could synchronously execute the scripts and call the callback afterwards.
In addition we could wrap all these steps in a single lambda which we execute in a new thread.

plata: I have a different idea. Let me try something.

plata: Ok. It works. I'm not sure if there is a better way than calling this function like I did it now.

plata:
I was thinking about this. But I decided not to do it on purpose because it's not really functionality. Maybe it is somehow possible to do this without an explicit function.

PR #1106, opened 2017-09-01
Improve and refactor EnginesView

### LINKED ISSUE #1096: Remove empty engines tabs during the filter process
When filtering the engines for a given search term some of the engine sub category tabs may be empty.
It would be quite nice if these tabs aren't shown.

### LINKED ISSUE #1097: Prevent eventual performance issues in container tab
The current implementation of the containers tab and its filter may introduce some performance issues later on, when POL contains more than only wine engines.

The reason for this is, that every time the user changes the selected engine type new `EngineSubCategoryTab` objects are created. Each of these new objects will call `filter.addOnFilterChanged(filteredEngineVersions::trigger);`, which in turn can lead to a lot of filter `trigger()` calls when a search term has been entered.

I can think of two approaches how to solve this, either:

- add a way to remove the added filter changed trigger, when the `EngineSubCategoryTab` gets removed from the list, or
- create all available `EngineSubCategoryTab` objects, for all available `EngineCategoryTab` objects, during the instantiation of the `EnginesView`, then we would only need to switch between the tabs later on.

madoar:
Fixes #1097, fixes #1096. In addition the engine sub categories are now ordered alphabetically.

This is quite fun to use. ;D

grafik
grafik
grafik

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EngineSubCategoryTab.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EngineSubCategoryTab.java
@@ -32,6 +32,8 @@ public class EngineSubCategoryTab extends Tab {
     private EngineSubCategoryDTO engineSubCategory;
     private String enginesPath;
 
+    private Predicate<EngineVersionDTO> filterPredicate;
+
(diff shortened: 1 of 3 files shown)


plata:
It's good that the code for the different tabs gets streamlined more and more. This makes it really easy to understand.

PR #1115, opened 2017-09-03
enabled icons for scripts

plata:
see PhoenicisOrg/Scripts#396


phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ScriptDTO.java
```diff
@@ -39,6 +39,7 @@
     private final Boolean free;
     private final Boolean requiresPatch;
     private final String script;
+    private final URI icon;
```
madoar: Can we give this field a more descriptive name? When I use a variable named `icon` I normally expect an `Image` object.

plata: It the same as for `ApplicationDTO`. It doesn't make sense to use a different name for both DTOs.

PR #1116, opened 2017-09-03
added ID for scripts

plata:


plata:
The `name` is translated, so we need `id` if we want to identify a script.

PR #1123, opened 2017-09-12
added shortcut ID

plata:
The shortcut files cannot use the shortcut name because it might contain invalid characters (especially "/" as in "TRON RUN/r").

requires PhoenicisOrg/Scripts#413

### DIFF
--- a/phoenicis-library/src/main/java/org/phoenicis/library/LibraryManager.java
+++ b/phoenicis-library/src/main/java/org/phoenicis/library/LibraryManager.java
@@ -98,6 +98,7 @@ public class LibraryManager {
 
     private ShortcutDTO fetchShortcutDTO(File shortcutDirectory, File file) {
         final String baseName = FilenameUtils.getBaseName(file.getName());
+        final File nameFile = new File(shortcutDirectory, baseName + ".name");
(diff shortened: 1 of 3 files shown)


phoenicis-library/src/main/java/org/phoenicis/library/LibraryManager.java
```diff
@@ -108,6 +109,12 @@ private ShortcutDTO fetchShortcutDTO(File shortcutDirectory, File file) {
             final URI miniature = miniatureFile.exists() ? miniatureFile.toURI()
                     : getClass().getResource("defaultMiniature.png").toURI();
 
+            String name = baseName;
+            if (nameFile.exists()) {
+                name = IOUtils.toString(new FileInputStream(nameFile), "UTF-8");
+                name = name.replace("\n", "");
```
madoar: Are there other "bad" characters?

plata: I think you misunderstood this. The "\n" is just removed because you might have a line break in the `shortcut.name`. It has nothing to do with "bad" characters. Those are filtered in the `ShortcutDTO`.

PR #1125, opened 2017-09-13
combine shortcut info in one json file

### LINKED ISSUE #1124: reduce number of shortcut files
Currently, there is a file for the description, a file for the category etc. These files should be combined in one `.info` file containing all required information in a json format.

For example:
```JSON
{
"name": "xyz",
"category": "Games",
"description": "lorem ipsum..."
}
```

plata:
fixes #1124 

requires PhoenicisOrg/Scripts#416

### DIFF
--- a/phoenicis-library/src/main/java/org/phoenicis/library/LibraryConfiguration.java
+++ b/phoenicis-library/src/main/java/org/phoenicis/library/LibraryConfiguration.java
@@ -18,7 +18,7 @@
 
 package org.phoenicis.library;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
+import org.phoenicis.configuration.PhoenicisGlobalConfiguration;
 import org.phoenicis.scripts.ScriptsConfiguration;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
@@ -30,6 +30,9 @@ public class LibraryConfiguration {
     @Autowired
     private ScriptsConfiguration scriptsConfiguration;
 
+    @Autowired
+    private PhoenicisGlobalConfiguration phoenicisGlobalConfiguration;
+
     @Value("${application.user.shortcuts}")
     private String shortcutDirectory;
 
@@ -38,7 +41,7 @@ public class LibraryConfiguration {
 
     @Bean
     public LibraryManager libraryManager() {
-        return new LibraryManager(shortcutDirectory, new ObjectMapper());
+        return new LibraryManager(shortcutDirectory, phoenicisGlobalConfiguration.objectMapper());
     }
 
     @Bean
@@ -47,7 +50,7 @@ public class LibraryConfiguration {
                 desktopShortcutDirectory,
                 libraryManager(),
                 scriptsConfiguration.scriptInterpreter(),
-                new ObjectMapper());
+                phoenicisGlobalConfiguration.objectMapper());
     }
 
     @Bean
--- a/phoenicis-library/src/main/java/org/phoenicis/library/LibraryConfiguration.java
+++ b/phoenicis-library/src/main/java/org/phoenicis/library/LibraryConfiguration.java
@@ -18,6 +18,7 @@
 
 package org.phoenicis.library;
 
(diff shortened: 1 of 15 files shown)


phoenicis-library/src/main/java/org/phoenicis/library/LibraryConfiguration.java
```diff
@@ -37,13 +38,16 @@
...
     public LibraryManager libraryManager() {
-        return new LibraryManager(shortcutDirectory);
+        return new LibraryManager(shortcutDirectory, new ObjectMapper());
     }
 
     @Bean
     public ShortcutManager shortcutManager() {
-        return new ShortcutManager(shortcutDirectory, desktopShortcutDirectory, libraryManager(),
-                scriptsConfiguration.scriptInterpreter());
+        return new ShortcutManager(shortcutDirectory,
```
madoar: As far as I remember we have a `Bean` somewhere for the `ObjectMapper`

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
```diff
@@ -39,42 +42,45 @@
...
             shortcutDirectoryFile.mkdirs();
         }
 
+        final Map<String, String> shortcutInfo = new HashMap<>();
+        shortcutInfo.put("name", shortcutDTO.getName());
+        shortcutInfo.put("category", shortcutDTO.getCategory());
+        shortcutInfo.put("description", shortcutDTO.getDescription());
+
         try {
+            String infoJson = this.objectMapper.writeValueAsString(shortcutInfo);
```
madoar: I would like it better, if we had a wrapper object with some annotations instead of a `HashMap`. This would make it easier to track what attributes are contained inside an `info` file.
An example for what I mean are the `RepositoryLocation` classes

phoenicis-library/src/main/java/org/phoenicis/library/LibraryConfiguration.java
plata: Where? There's one in `RepositoryConfiguration` but I don't think it makes sense to use that one. I followed the example of `EnginesConfiguration`.

madoar: Do we have access to `PhoenicisGlobalConfiguration`?

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
plata: Wouldn't that be almost duplicate of `ShortcutDTO`?

phoenicis-library/src/main/java/org/phoenicis/library/LibraryConfiguration.java
plata: Yes. The `ContainersConfiguration` uses this.

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
madoar: Then I would suggest tomove the stuff from inside `ShortcutDTO` to the new class with annotations. 
Afterwards we can add an instance of the new class `ShortcutInfo` to `ShortcutDTO`.

phoenicis-library/src/main/java/org/phoenicis/library/dto/ShortcutInfoDTO.java
```diff
@@ -0,0 +1,131 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Comparator;
+
+@JsonDeserialize(builder = ShortcutInfoDTO.Builder.class)
+public class ShortcutInfoDTO {
```
madoar: JavaDoc

PR #1131, opened 2017-09-16
add context to executed scripts

### LINKED ISSUE #1130: make scripts context aware
Currently, a script doesn't know anything about its context when it's executed (e.g. which application it belongs to).
This leads to problems like:
- PhoenicisOrg/Scripts#361
- PhoenicisOrg/Scripts#269

It should be possible to query the `ScriptDTO`, `ApplicationDTO` etc. for a running script.

The most simple approach for this would be to inject "environment" variables into the  script execution (from `AppsController` in `setOnSelectScript`).
For example, we could inject the application ID "MafiaII" for the "Mafia II (Demo)" script as `APPLICATION_ID`. This could then be used to get the correct miniature in `WineShortcut`.

To get this working, the repository DTOs need a notion of "parent". I don't think it is possible to refer to the parent DTO object directly but it should be possible to store the ID of it. For example, `ScriptDTO` would get a new field `ApplicationId` which is filled when the repository is fetched.

I'm unsure if this is the way we should go or if there is a better solution. Therefore I've opened this issue to get your feedback before implementing something.

plata:
fixes #1130

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
@@ -72,12 +72,29 @@ public class AppsController {
         });
 
         this.view.setOnSelectScript(
-                scriptDTO -> scriptInterpreter.runScript(scriptDTO.getScript(), e -> Platform.runLater(() -> {
-                    // no exception if installation is cancelled
-                    if (!(e.getCause() instanceof InterruptedException)) {
-                        new ErrorMessage(tr("The script ended unexpectedly"), e, this.view);
-                    }
-                })));
+                scriptDTO -> {
+                    final StringBuilder environmentBuilder = new StringBuilder();
+                    environmentBuilder.append("TYPE_ID=\"");
+                    environmentBuilder.append(scriptDTO.getTypeId());
+                    environmentBuilder.append("\";\n");
+                    environmentBuilder.append("CATEGORY_ID=\"");
+                    environmentBuilder.append(scriptDTO.getCategoryId());
+                    environmentBuilder.append("\";\n");
+                    environmentBuilder.append("APPLICATION_ID=\"");
+                    environmentBuilder.append(scriptDTO.getApplicationId());
+                    environmentBuilder.append("\";\n");
+                    environmentBuilder.append("SCRIPT_ID=\"");
+                    environmentBuilder.append(scriptDTO.getId());
+                    environmentBuilder.append("\";\n");
+                    final String environment = environmentBuilder.toString();
+                    final String execute = environment + scriptDTO.getScript();
+                    scriptInterpreter.runScript(execute, e -> Platform.runLater(() -> {
+                        // no exception if installation is cancelled
+                        if (!(e.getCause() instanceof InterruptedException)) {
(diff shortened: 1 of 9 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
```diff
@@ -72,12 +72,29 @@ public void loadApps() {
...
+                    environmentBuilder.append("CATEGORY_ID=\"");
+                    environmentBuilder.append(scriptDTO.getCategoryId());
+                    environmentBuilder.append("\";\n");
+                    environmentBuilder.append("APPLICATION_ID=\"");
+                    environmentBuilder.append(scriptDTO.getApplicationId());
+                    environmentBuilder.append("\";\n");
+                    environmentBuilder.append("SCRIPT_ID=\"");
+                    environmentBuilder.append(scriptDTO.getId());
+                    environmentBuilder.append("\";\n");
+                    final String environment = environmentBuilder.toString();
```
madoar: Why don't we inject the dtos directly?

plata: If we do that, you would have direct access to more or less the complete repository. I'm not sure that's a good idea. Also it feels a bit strange to me to store objects in environment variables. If a script needs to access the DTO, it can still use e.g. `getApplication()`.

madoar: I think we should inject a context object, that provides a getter method for all important context information, i.e. the typeId, scriptId, applicationId. These getter functions could then do the lookup. The main benefit of this is, that we can hide the lookup logic of the objects from the scripts.
I would suggest to provide this object via `NashornEngine#put`, instead of via `StringBuilder`.

plata: As I said above: I don't really like the idea of injecting global objects. @qparis what do you think?

madoar: I'm not sure what you mean? The access scope should be nearly the same as for your `environment` string approach, right?

plata: It's not about the scope. I just don't like that there would be an **object** which is "just there". The idea of environment variables as strings is quite common (e.g. in a shell). Therefore this approach somehow feels more "natural" for me.

madoar: I understand where you're coming from, but I still think it's nicer in our case to directly inject the objects, instead of letting the user fetch them himself, especially because we already know them on the java side.

I would like to wait for @qparis to see what he thinks about this.

madoar:
@qparis can you please take a look at this PR?

plata:
I think I will merge this. There was more than enough time for everybody to request changes if required.

madoar:
@plata I agree. If we can think of a better way to solve the problem later on we should still be able to change it then.

plata:
@madoar can you approve it then?

PR #1133, opened 2017-09-18
secure null icon/miniature when editing shortcut

### LINKED ISSUE #1132: NPE on edit applications settings
Hello, I have found an NPE when I want to save modification on an application
I juste want to remove the Wine Debug flag
https://pastebin.com/pVU6P4cb
playonlinuxaplicationbug

System informations:
linux mint 18.2 with kde
openjdk version "1.8.0_131"
OpenJDK Runtime Environment (build 1.8.0_131-8u131-b11-2ubuntu1.16.04.3-b11)
OpenJDK 64-Bit Server VM (build 25.131-b11, mixed mode)

plata:
fixes #1132

### DIFF
--- a/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
+++ b/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
@@ -172,7 +172,7 @@ public class ShortcutManager {
                     Files.move(iconFile.toPath(), iconBackup.toPath());
                     shortcutDTO = new ShortcutDTO.Builder(shortcutDTO).withIcon(iconBackup.toURI()).build();
                 } catch (IOException e) {
-                    LOGGER.error("Could not backup icon.");
+                    LOGGER.error("Could not backup icon.", e);
                 }
             }
         }
@@ -189,7 +189,7 @@ public class ShortcutManager {
                     Files.move(miniatureFile.toPath(), miniatureBackup.toPath());
                     shortcutDTO = new ShortcutDTO.Builder(shortcutDTO).withMiniature(miniatureBackup.toURI()).build();
                 } catch (IOException e) {
-                    LOGGER.error("Could not backup miniature.");
+                    LOGGER.error("Could not backup miniature.", e);
                 }
             }
         }
--- a/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
+++ b/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
@@ -31,6 +31,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URI;
 import java.nio.file.Files;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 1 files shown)


phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
```diff
@@ -162,28 +163,34 @@ public void updateShortcut(ShortcutDTO shortcutDTO) {
...
+        final URI shortcutIcon = shortcutDTO.getIcon();
+
+        if (shortcutIcon != null) {
+            final boolean keepIcon = shortcutIcon.getPath().equals(iconFile.getPath());
+            if (keepIcon) {
+                try {
+                    Files.move(iconFile.toPath(), iconBackup.toPath());
+                    shortcutDTO = new ShortcutDTO.Builder(shortcutDTO).withIcon(iconBackup.toURI()).build();
+                } catch (IOException e) {
+                    LOGGER.error("Could not backup icon.");
```
madoar: Maybe add the exception to `LOGGER.error`

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
```diff
@@ -162,28 +163,34 @@ public void updateShortcut(ShortcutDTO shortcutDTO) {
...
+        final URI shortcutMiniature = shortcutDTO.getMiniature();
+
+        if (shortcutMiniature != null) {
+            final boolean keepMiniature = shortcutMiniature.getPath().equals(miniatureFile.getPath());
+            if (keepMiniature) {
+                try {
+                    Files.move(miniatureFile.toPath(), miniatureBackup.toPath());
+                    shortcutDTO = new ShortcutDTO.Builder(shortcutDTO).withMiniature(miniatureBackup.toURI()).build();
+                } catch (IOException e) {
+                    LOGGER.error("Could not backup miniature.");
```
madoar: Same as last comment

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
```diff
@@ -162,28 +163,34 @@ public void updateShortcut(ShortcutDTO shortcutDTO) {
...
-            try {
-                Files.move(miniatureFile.toPath(), miniatureBackup.toPath());
-                shortcutDTO = new ShortcutDTO.Builder(shortcutDTO).withMiniature(miniatureBackup.toURI()).build();
-            } catch (IOException e) {
-                LOGGER.error("Could not backup miniature.");
+        final URI shortcutMiniature = shortcutDTO.getMiniature();
+
+        if (shortcutMiniature != null) {
+            final boolean keepMiniature = shortcutMiniature.getPath().equals(miniatureFile.getPath());
+            if (keepMiniature) {
```
madoar: I kind of think an `Optional` approach would be shorter (doesn't need so many indentions):

```java
Optional.ofNullable(shortcutDTO.getMiniature())
    .filter(shortcutMiniature -> shortcutMiniature.getPath().equals(miniatureFile.getPath()))
    .ifPresent(shortcutMiniature -> ...);
```

plata: If we want an `Optional`, I think we should do it properly and use `Optional` in the DTO already. However, I really wouldn't want to do this now/in this PR (not sure which effects it will have in the scripts repo etc.).

PR #1141, opened 2017-10-18
set default value if tools.linux-terminal does not exist

### LINKED ISSUE #1139: Unable to Build on MacOS
Hello, I'm currently attempting to install the latest version of PlayOnMac5, and it seems there is an issue when building: `Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project phoenicis-javafx: There are test failures.`

Here are the logs [WITH](https://paste.ee/p/H79Io) or [WITHOUT](https://paste.ee/p/CiujT) full debugging (-X switch) enabled.

Sorry I can't elaborate anymore; I'm not accustomed to using Maven, nor installing programs from source.

plata:
For example on Mac OS, `tools.linux-terminal` does not exist (also not required). Use an empty String as default in these cases.

fixes #1139

### DIFF
--- a/phoenicis-settings/src/main/java/org/phoenicis/settings/Setting.java
+++ b/phoenicis-settings/src/main/java/org/phoenicis/settings/Setting.java
@@ -19,7 +19,7 @@
 package org.phoenicis.settings;
 
 public enum Setting {
-    TERMINAL("tools.linux-terminal");
+    TERMINAL("${tools.linux-terminal:}");
 
     private final String propertyName;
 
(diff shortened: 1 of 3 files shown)


phoenicis-settings/src/main/java/org/phoenicis/settings/Setting.java
```diff
@@ -19,7 +19,7 @@
 package org.phoenicis.settings;
 
 public enum Setting {
-    TERMINAL("tools.linux-terminal");
+    TERMINAL("${tools.linux-terminal:}");
```
madoar: Are you sure about the `:`?

plata: Yes, why not? See <http://www.baeldung.com/spring-value-defaults>

madoar: Just asking, it looks strange to me

PR #1153, opened 2018-01-20
Cleaning up repository location code

qparis:


phoenicis-repository/src/main/java/org/phoenicis/repository/DynamicRepositoryLocationLoader.java
```diff
@@ -0,0 +1,20 @@
...
+import org.phoenicis.repository.location.RepositoryLocation;
+import org.phoenicis.repository.repositoryTypes.Repository;
+import org.springframework.context.ApplicationContext;
+
+import java.util.List;
+
+public class DynamicRepositoryLocationLoader implements RepositoryLocationLoader {
+    private final RepositoryLocationLoader delegated;
+
+    public DynamicRepositoryLocationLoader(ApplicationContext applicationContext, String beanName) {
```
madoar: Is there a way to use the `className` instead of the `beanName` here?

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryConfiguration.java
```diff
@@ -84,11 +81,22 @@ public RepositoryManager repositoryManager() {
...
+        return new DynamicRepositoryLocationLoader(applicationContext, "filesystemLocationLoader");
+    }
+
+    @Bean
+    public RepositoryLocationLoader filesystemLocationLoader() {
+        return new FilesystemJsonRepositoryLocationLoader(repositoryListPath,
+                phoenicisGlobalConfiguration.objectMapper());
+    }
+
     public void saveRepositories(List<RepositoryLocation<? extends Repository>> repositoryLocations) {
```
madoar: Can we add this method to `RepositoryLocationLoader`?

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryLocationLoader.java
```diff
@@ -0,0 +1,10 @@
+package org.phoenicis.repository;
+
+import org.phoenicis.repository.location.RepositoryLocation;
+import org.phoenicis.repository.repositoryTypes.Repository;
+
+import java.util.List;
+
+public interface RepositoryLocationLoader {
```
madoar: Please add some JavaDoc

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryConfiguration.java
```diff
@@ -84,11 +81,22 @@ public RepositoryManager repositoryManager() {
                 backgroundRepositoryFactory());
 
         // set initial repositories
-        repositoryManager.addRepositories(this.loadRepositoryLocations().toArray(new RepositoryLocation[0]));
+        repositoryManager.addRepositories(this.repositoryLocationLoader().loadRepositoryLocations().toArray(new RepositoryLocation[0]));
```
madoar: Can we alternatively pass the `RepositoryLocationLoader` to the `DefaultRepositoryManager` constructor?
I believe this may be a more elegant design that what we have currently (i.e. `repositoryManager.addRepositories`)

phoenicis-repository/src/main/java/org/phoenicis/repository/DynamicRepositoryLocationLoader.java
qparis: This is really the name of the bean that we consider here, not the name of the class

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryConfiguration.java
qparis: The would be more elegant, but I'm not sure that it won't break anything

qparis: And I also think that the interface of RepositoryManager needs to be changed

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryLocationLoader.java
qparis: Sure

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryConfiguration.java
qparis: Sure

phoenicis-repository/src/main/java/org/phoenicis/repository/DynamicRepositoryLocationLoader.java
madoar: I know.
I just think it would be nice to use the class name instead of the bean name.
Especially in the config/properties file I believe it to be more intuitive to see the class name than the bean name which is specified inside an on the first view arbitrary java file (i.e. `RepositoryConfiguration`).

qparis: The thing is that you could have several beans with the same class theoritically. For example something like systemFileSystemLocationLoader and a userHomeLocationLoader could both be FilesystemJsonRepositoryLocationLoader

madoar: Couldn't we use the fully qualified name (i.e. package + class name)?
This name should be unique too, right?

qparis: No because we can have two beans with different names but using the same class. They will only differ by the parameters you pass on their respective constructors

qparis: A bean represents an instance of a given class. 

In most cases, these instance are singletons because they are statically declared (i.e. in a configuration file), but this is not always true. This is why spring introduces the concept of bean name

PR #1166, opened 2018-02-15
added gemnasium Maven plugin

plata:


madoar:
It seems like the tests are failing

plata:
Yes. I'm not sure why it cannot update gemnasium currently.

madoar:
Can you try running the tests with `-X`? Maybe we get some better error message this way.
In addition: Is it possible to test the plugin without travis locally, to see if it's maybe a problem with the travis setup?

plata:
I try locally but then I get "API Error: customer has no free slot". I'm following http://support.gemnasium.com/knowledgebase/articles/1834507-does-gemnasium-support-maven-multi-modules-project in case you want to try as well.

madoar:
```
NB: This setup leverages the gemnasium.properties files and only requires you to put the plugin configuration in the parent pom.xml.
This also means that you can't configure the projectSlug in the parent pom.xml file because it will take precedence over the gemnasium.properties of each module. 

If you prefer to go with pom.xml only configuration, then you'll need to override the plugin config in each module's pom.xml to set the correct projectSlug.
```

I believe we need to set the `projectSlug` inside each `pom.xml` file, right?
To execute the tests on travis with the `-X` option we'll need to override the test command, this can be done via: https://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Build-Step

plata:
I've also seen that. But it doesn't work if you remove the slug either.

madoar:
I'm not sure if we get a more clear error message without modifing the travis script.
I guess the api key is somehow bound otherwise you wouldn't get a different error message locally, right?

plata:
I think what's happening is that Travis tries to run the command for the same slug for all submodules. Therefore I tried to create several gemnasium projects locally as described in the how-to. However, I have the feeling that it tries to create private projects which is not possible with the free subscription.

plata:
If I just run `mvn verify` locally, I get the same message as on Travis.

PR #1167, opened 2018-02-16
run formatter

plata:


plata:
Why did Travis fail? I didn't change anything.

madoar:
The error message says, the test failed because it timed out. Can you restart the tests?
I don't think I have the necessary permissions.

plata:
I saw that there was a timeout but why? Can we change anything to avoid this in the future?

madoar:
I'm not sure how the timeout was produced, but I believe the cause was junit.
The question is do we have test methods that are marked with a timeout, or a rule? (see also [here](https://github.com/junit-team/junit4/wiki/timeout-for-tests))

plata:
Should be AppConfigurationInitialisationTest.java.

Issue #618, opened 2017-02-20
Check possible impact of Java 9


plata:

madoar:
Depending on how far the Java dev team gets when they release Java 9, it's possible that we can use ES6 class definitions in Java 9, which would make some things a lot easier to read.

plata:
First of all we have to be sure that Java 9 does not break anything.

plata:
Java 9 is tested with Travis since #1172

PR #1172, opened 2018-02-16
Travis CI: change config to use multiple operating system / jdk combinations

madoar:
Please be aware, that the osx java version seems to depend on the version of xcode chosen.
The xcode version in turn depends on the chosen osx version...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ addons:
       - oracle-java8-installer
       - oracle-java9-installer
 
+before_install:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+
 matrix:
   include:
     # The normal tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,5 @@
 language: java
-
-dist: trusty
 sudo: false
-
-os:
-- linux
 
 addons:
   apt:
@@ -12,19 +7,16 @@ addons:
       - oracle-java8-installer
       - oracle-java9-installer
 
-jdk:
-- oraclejdk8
-- oraclejdk9
-- openjdk8
-
-env:
-
 matrix:
   include:
-  - language: java
-    os: osx
-
-jobs:
-  include:
-    - stage: Check format
-      script: ./check_formatted.sh
+    - jdk: oraclejdk8
+      os: linux
+      dist: trusty
+    - jdk: oraclejdk9
+      os: linux
+      dist: trusty
+    - jdk: openjdk8
+      os: linux
+      dist: trusty
+    - os: osx
+      osx_image: xcode9.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,6 @@ addons:
       - oracle-java8-installer
       - oracle-java9-installer
 
-before_install:
-    # enable gui, see https://docs.travis-ci.com/user/gui-and-headless-browsers and https://github.com/SublimeText/UnitTesting/issues/74
-    - if [ "$TRAVIS_OS_NAME" == "linux"  ]; then
-      export DISPLAY=:99.0;
-      sh -e /etc/init.d/xvfb start;
-      sleep 3;
-      fi
-
 matrix:
   include:
     # The normal tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,11 @@ addons:
 
 before_install:
     # enable gui, see https://docs.travis-ci.com/user/gui-and-headless-browsers and https://github.com/SublimeText/UnitTesting/issues/74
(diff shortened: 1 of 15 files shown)


madoar:
Ok. So far so well, here the current progress:
- travis doesn't support openjdk 9 yet
- travis doesn't support openjfx (i.e. JavaFX for OpenJDK)

In addition it seems like Java 9 added a new method to the `TransformationList` class: `int getViewIndex​(int index)`, because we're not implementing this method (yet), the build for OracleJDK 9 fails.

@plata what do you think?

plata:
I guess we should ensure that Java 9 works. Then we can also close #618. I would suggest to build Java 8 and 9 for Linux and OSX (that's 4 builds already). If we should get issues with OpenJDK/openjfx, we can still decide to add it later on.

plata:
What's the purpose of that plugin?

madoar:
It was a test, where I tried to tell maven to compile the project with java 8 specs, but it doesn't seem to work like I thought. 

I'm currently not sure if it's a good idea to add the `getViewIndex​` method.
The reason being that normally we should also add `@Override` before it.
But as far as I know this would produce an error/warning message with java 8, because Java 8 doesn't define this method, right?

plata:
I'm not sure. But there should be a way to make it work for Java 8 *and* 9, right?

madoar:
I don't know.

When I add the following method to an arbitrary class:
```java
    @Override
    public void test() {

    }
```

I get the following error message:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project phoenicis-javafx: Compilation failure
[ERROR] .../POL-POM-5/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/MainWindow.java:[89,5] method does not override or implement a method from a supertype
```

This leads me to the belief that this could become problematic, if handled wrong.

plata:
I understand. However, I believe that in this particular case it would be o.k. to leave the `@Override` away or comment it out.

madoar:
Some dependencies are making problems, when used with Java 9.
The ones I could identify are powermock and jacoco, but it's possible that some other dependencies are incompatible with Java 9 too.

I guess we should first update all dependencies to a more current version, before we continue working on this.

madoar:
After solving the mockito problem it seems like I've stumbled over the next one.
Not we seem to have a problem with `openjfx-monocle` which I believe is required to test javafx applications without a graphical device, which travis doesn't provide.
The problem is quite clear, we're using `openjfx-monocle` version `8u76-b04`, which should be used for Java 8 and not Java 9.
I guess we need to use `jdk-9+181` to make it work with Java 9, but I'm not sure if this will work with Java 8...

Has anyone an idea how to solve this?

plata:
Can you try if the JDK 9 monocle version works with Java 8?

madoar:
Ok this should work for our current purposes.

I've solved the monocle problem by removing it.
This additional required a change from headless tests to tests showing a GUI window.
Luckily it's possible to make travis work with tests that can't work in headless mode.

plata:
But this doesn't mean that a JavaFX window pops up when you run the tests, right?

madoar:
Yes, a javafx window pops up when you run the tests

plata:
Do we want that? I'm thinking how it would feel if windows pop up every time I build.

madoar:
I don't know.
I'm not sure if we have an alternative.
As far as I understand `openjfx-monocle` the used version depends on your underlying Java version (at least the readme at https://github.com/TestFX/Monocle suggests that to me)

plata:
Is it possible to configure Travis such that it uses different virtual machines for the Java 8 and 9 build (like it does for Linux and OSX)?

madoar:
What exactly do you mean by different virtual machines?

As far as I understand travis it runs each "job" inside its own environment.
This PR introduces 4 jobs 2 on linux with java 8 and 9, one on mac and another one on linux for the formatting check.

plata:
I meant the environment. If we have 2 for Java 8 and 9, can't we just install monocle for Java 8/9 in the respective environment?

madoar:
We're currently installing monocle via maven, I'm not sure if we can switch the version with maven depending on some system variable.
In addition I'm not sure if we want that, because this also affects the build process of us developers.

plata:
Have a look at TestFX/TestFX#433 (2nd to last entry): they use profiles to get it working for Java 8 and 9. This seems to be the "official" way to solve this issue.

madoar:
Ok, I've readded monocle for headless tests.
When looking at the finished tests, it looks like it's working.
This is quite hard to test, when you're not using Java 9 :/

madoar:
Now that the OSX tests finished too, it seems like it's working as intended :)

Later on I think we should try to get travis working with OpenJDK, so that we'll have some automated tests for that jdk platform too, but I think that still has time.
Otherwise I believe this PR has been fully completed and can be merged, what do you think @plata?

PR #1192, opened 2018-03-15
added .eslintrc.yml

plata:
We have .js files for testing in phoenicis-tests/src/main/resources/org/phoenicis/tests. Therefore, we must adjust ESLint settings for Codacy.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,0 +1,25 @@
+globals:
+  include: true  
+  # java
+  java: true
+  Bean: true
+  # files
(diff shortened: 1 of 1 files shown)


plata:
It's pretty simple: We have some global (classes/functions) which ESLint doesn't know. To avoid that it complains, we just list the globals.

madoar:
So these globals are keywords (variable-, function-, class- names), that eslint then ignores?

plata:
Yes, or better: it assumes that they exist in the global scope.

PR #1197, opened 2018-03-17
use GitHub Pages for documentation

plata:
GitHub Pages have some advantages compared to the wiki, e.g.:
* Documentation can be updated with pull requests so anybody can contribute.
* Better tables of content

How it looks like:
https://plata.github.io/POL-POM-5/


madoar:
This looks quite nice.
I have a few questions:
- how is the html code generated?
- where do you suggest to host the github pages (i.e. in a docs folder or on a new branch)?
- how do we build the pages (I guess with travis)?

plata:
* The HTML is generated by GitHub Pages internally (via Jekyll).
* The pages are stored in the "docs" folder.
* Do you want to use Travis to check pull requests for the pages? I'm not sure if that's possible. However, the pages are written in Markdown so the standard GitHub preview should be enough to show the differences and whether or not it will work.

madoar:
I think we can go with this, what do you think @qparis?

plata:
Travis checks are also working now, so ok to merge from my side.

plata:
@qparis please tell if we want to use this or not. If yes, I will also do it for the Scripts repository.

plata:
Do you know if the wiki will be deleted if I uncheck it in the settings?

PR #1199, opened 2018-03-18
Travis CI: do not run maven build for check format

### LINKED ISSUE #1198: Travis: do not run maven build for check format

plata:
fixes #1198

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,4 +27,4 @@ matrix:
       # do not use default install (runs formatter)
       install: true
       script:
-        - mvn -Pcheck-formatted validate formatter:validat
+        - mvn -Pcheck-formatted validate formatter:validate
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,6 @@ matrix:
       dist: trusty
       install:
         # do not use default install (runs formatter)
-        - mvn dependency:go-offline
+        - mvn install:install
       script:
         - mvn -Pcheck-formatted validate formatter:validat
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ matrix:
       jdk: oraclejdk8
       os: linux
       dist: trusty
+      install: mvn dependency:go-offline
       script:
         - mvn -Pcheck-formatted validate formatter:validate
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,11 +18,12 @@ matrix:
         - mvn -Pcheck-formatted validate formatter:validate
 
     # The normal tests
-    - jdk: oraclejdk8
-      os: linux
-      dist: trusty
-    - jdk: oraclejdk9
-      os: linux
-      dist: trusty
-    - os: osx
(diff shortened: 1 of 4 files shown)


madoar:
I'm not sure we can do this, because the format check may run on a "fresh" container, which means, that the whole project hasn't been build previously.

plata:
You're right. But can we avoid that it performs a complete build and only run the formatter?

madoar:
In what way?
I guess you can run it together with the `Check github pages` check, but I wouldn't recommend adding it to the normal tests.

When exactly is the formatting done in the maven lifecycle?
Maybe an `maven clean compile` is enough?

plata:
There's a "format" goal in the top level pom. I would like to execute only this goal if possible.

madoar:
This should be possible by replacing `mvn test -B`, through the format goal maven command.
Alternatively we could also think about adding this command to the `check_formatted.sh` file?

plata:
I think this should work. Wasn't aware that you cannot use the fully qualified plugin name...

plata:
No. Doesn't work. It skips all the modules.

plata:
Just figured out that we do not even need our custom script to check if there are non-formatted files. We can use `formatter:validate`. However, I still have the problem that it doesn't work with multi modules (see revelc/formatter-maven-plugin#261).

plata:
I don't understand why the build doesn't fail. It's working on my local machine. Can anyone else try it?

madoar:
I've tried and it works for me too.
I've run `mvn -Pcheck-formatted validate formatter:validate` on the branch and got:
`[ERROR] Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.0.1:validate (validate) on project phoenicis-javafx: File '../POL-POM-5/phoenicis-javafx/src/main/java/org/phoenicis/javafx/UiMessageSenderJavaFXImplementation.java' format doesn't match! -> [Help 1]
`

madoar:
Taking a look at the log I see the following strange line `724`:
```bash
$ mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```

plata:
If Travis realizes at some point in time that it should run, we might see if the `install` was really the problem.

.travis.yml
```diff
@@ -9,21 +9,24 @@ addons:
...
+        - mvn dependency:go-offline
       script:
-      - mvn test -B
-      - ./check_formatted.sh
\ No newline at end of file
+        - mvn -Pcheck-formatted validate formatter:validate
+
+    # The normal tests
+    - stage: Build + test
+      - jdk: oraclejdk8
```
madoar: This line possibly contains a small problem for travis.
Because the `-` defines a new job inside the matrix.
Therefore we possibly have an empty job inside the `Build + Test` stage.
I suggest removing the `-` at the beginning of this line

plata:
It builds again but the `go-offline` doesn't work.

plata:
It's working now!

PR #1215, opened 2018-04-01
Fix ID handling for scripts

plata:


phoenicis-engines/src/main/java/org/phoenicis/engines/EngineToolsManager.java
```diff
@@ -48,19 +48,19 @@ public EngineToolsManager(ScriptInterpreter scriptInterpreter) {
...
-     * @param engineName
+     * @param engineId
      * @param container
-     * @param toolName
+     * @param toolId
      * @param doneCallback
      * @param errorCallback
      */
-    public void runTool(String engineName, String container, String toolName, Runnable doneCallback,
+    public void runTool(String engineId, String container, String toolId, Runnable doneCallback,
```
madoar: Can you please add some descriptions about what the parameters should contain to the JavaDoc

phoenicis-repository/src/main/java/org/phoenicis/repository/types/ClasspathRepository.java
```diff
@@ -296,14 +297,23 @@ private ScriptDTO buildScript(String typeId, String categoryId, String applicati
...
-                    .withScript(new String(IOUtils.toByteArray(scriptFile)))
-                    .withIcon(icon)
-                    .build();
+            ScriptDTO.Builder scriptDTOBuilder = new ScriptDTO.Builder(
+                    objectMapper.readValue(scriptJsonInputStream, ScriptDTO.class))
+                            .withTypeId(typeId)
+                            .withCategoryId(categoryId)
+                            .withApplicationId(applicationId)
+                            .withScript(new String(IOUtils.toByteArray(scriptFile)))
+                            .withIcon(icon);
```
madoar: The indentation looks kind of strange here

plata: The formatter indents it like this. I cannot change it.

phoenicis-repository/src/main/java/org/phoenicis/repository/types/LocalRepository.java
```diff
@@ -111,9 +111,9 @@ public RepositoryDTO fetchInstallableApplications() {
 
                     if (StringUtils.isBlank(jsonTypeDTO.getId())) {
                         if (!StringUtils.isBlank(jsonTypeDTO.getName())) {
-                            typeDTOBuilder.withId(jsonTypeDTO.getName().replaceAll("[^a-zA-Z0-9]", ""));
+                            typeDTOBuilder.withId(jsonTypeDTO.getName().replaceAll("[^a-zA-Z0-9_]", ""));
```
madoar: I suggest, that we move the regex (`[^a-zA-Z0-9_]`) to a separate constant field, because we're using it in multiple places and I think it's quite prone to misspellings.

PR #1240, opened 2018-04-12
Object oriented engine tools

plata:
requires https://github.com/PhoenicisOrg/scripts/pull/575


phoenicis-engines/src/main/java/org/phoenicis/engines/EngineTool.java
```diff
@@ -0,0 +1,12 @@
+package org.phoenicis.engines;
+
+/**
+ * interface which must be implemented by all engine tools in Javascript
+ */
+public interface EngineTool {
+    /**
+     * runs the tool in the given container
+     * @param container container where the tool shall be run
```
madoar: Is this the name of the container? An id? Or the path leading to the container?

PR #1241, opened 2018-04-12
Object oriented installers

plata:
Open issues: Aborting an installation does not work.

requires https://github.com/PhoenicisOrg/scripts/pull/577


plata:
@madoar any idea why aborting does not work?

madoar:
No, do you get an exception/error when you try to interrupt a script?

plata:
Yes. I think I know the problem: If I retrieve the Java object and use this to call `run`, the `InterruptedException` is not thrown in the context of the script execution. I can catch it but then the wizard will not close (because the script interpreter didn't even realize that there was an exception).
I will implement a work around but maybe we should rethink this.

Issue #1219, opened 2018-04-05
extend git repository to use certain tag/revision


plata:
Would be useful for releases: tag the scripts repository and use that tag. Currently we would have to use a branch for this.

PR #1243, opened 2018-04-14
Use alpha branch of scripts repository

plata:


plata:
I don't quite understand. How exactly do you want to do it?

plata:
If you are talking about the `.properties`: @madoar had to move the repository configuration from the `.properties` to `repositories.json` because it wasn't possible to define the repositories in the `.properties` properly.

phoenicis-repository/src/main/java/org/phoenicis/repository/FilesystemJsonRepositoryLocationLoader.java
```diff
@@ -45,7 +45,7 @@ public FilesystemJsonRepositoryLocationLoader(String repositoryListPath, ObjectM
             try {
                 result.add(new GitRepositoryLocation.Builder()
                         .withGitRepositoryUri(new URL("https://github.com/PhoenicisOrg/scripts").toURI())
-                        .withBranch("master").build());
+                        .withBranch("5.0-alpha").build());
```
qparis: I'm talking about this string

plata: I know, but how should this be solved with a configuration? Do you want do specify only the default git branch in `.properties`? I don't think this makes a lot of sense (e.g. the repository URL is still hard coded).

qparis:
Can you remind me why can't we put all these string in the configuration file? Isn't it better than hardcoding them?

plata:
The problem is that you cannot specify this kind of information in `.properties`. For example, for a git repository, we need URL, branch and maybe revision. If we want to specify this in `.properties`, we would have to encode it in a rather obscure string (something like "git+url:https://github.com/PhoenicisOrg/scripts+branch:master+revision:1234").

Or are you talking only about the fixed set of defaults, like this:
```properties
application.repository.default.git.url = "https://github.com/PhoenicisOrg/scripts"
application.repository.default.git.branch = "master"
application.repository.default.classpath = "/org/phoenicis/repository"
```

plata:
It's in #1249.

madoar:
I think it's better to solve this via #1219, see also my comment on #1252.

plata:
I put an explanation why I believe that we need a release branch to #1252.

PR #1248, opened 2018-04-14
Fix install/delete of engines

plata:


madoar:
What is the difference between this PR and #1247?

plata:
This is for the alpha release branch and #1247 is for master.

PR #1252, opened 2018-04-15
Describe required steps for a release in docs

plata:


docs/_data/navigation.yml
```diff
@@ -25,3 +25,5 @@ docs:
         url: /themes/
       - title: "Translation"
         url: /translation/
+      - title: "Release"
```
madoar: I believe the title to be misdirecting.
When reading `Release` I think about a list of past releases, and the corresponding changelog.

docs/_pages/release.md
```diff
@@ -0,0 +1,19 @@
+---
+title: "Release"
+permalink: /release/
+toc: false
+---
+
+The following steps must be executed to release a new version of Phoenicis:
+* Create release branches (e.g. "5.0-alpha") for phoenicis and scripts
```
madoar: Do we really want to use branches for the releases?
Normally you use the `Releases` tab (together with git tags) on GitHub to manage releases.

docs/_data/navigation.yml
plata: How should I name it?

docs/_pages/release.md
plata: We will use the GitHub Release tab. You can trigger releases for branches as well. The big advantage is that you can have bugfix releases in the release branch and continue development in master.

[Here](https://datasift.github.io/gitflow/IntroducingGitFlow.html) is a description of GitFlow. As we do not have a development branch (would be a bit difficult with forking etc), you should read that description as if our master branch is the development branch. 
Now the question: Which branch shall be used for releases? Obviously, the master branch cannot be used because it is used as development branch and therefore receives changes rather often which can possibly break it. Moreover, the release branch may receive only bugfixes which also rules out master.
So the best and simplest approach we can use in my opinion is to create release branches, merge bug fixes to the release branch and to master and use the release branch to trigger a GitHub release.

docs/_pages/release-process.md
```diff
@@ -0,0 +1,19 @@
...
+The following steps must be executed to release a new version of Phoenicis:
+* Create release branches (e.g. "5.0-alpha") for phoenicis and scripts
+* Protect release branch
+    * Require pull request reviews before merging
+    * Require status checks to pass before merging (Travis CI and Codacy)
+* on the release branch:
+    * Add release branch to `branches` section of `.travis.yml` such that Travis CI executes checks for the branch
+    * Specify scripts release branch in configuration (`application.repository.default.git.url`)
+    * Set release version for Maven in `pom.xml` files
+    * Set release version in .deb control files
```
madoar: This line will be hard to maintain I guess.
Currently we're only writing about `.deb` files, but what about a later mac os x installer or other formats.
I guess they will require a modification of their own control files.
Therefore I suggest to write this more generic.

docs/_pages/release-process.md
```diff
@@ -0,0 +1,19 @@
...
+* Test
```
madoar: Where should what be tested?

docs/_pages/release-process.md
```diff
@@ -0,0 +1,19 @@
...
+* Create GitHub release
```
madoar: Only Phoenicis itself or the scripts too?

docs/_pages/release-process.md
```diff
@@ -0,0 +1,19 @@
...
+* Announce release on phoenicis.org
```
madoar: Are there any required information to be published in the release announcement?

docs/_pages/release-process.md
```diff
@@ -0,0 +1,19 @@
+---
+title: "Release process"
+permalink: /release-process/
+toc: false
+---
+
+The following steps must be executed to release a new version of Phoenicis:
+* Create release branches (e.g. "5.0-alpha") for phoenicis and scripts
+* Protect release branch
```
madoar: I'm not sure I like it, that the release branch can be modified later on via pull requests.
The way I learned it, a release should be fixed, i.e. a frozen branch, which can't/shouldn't be modified.

When releasing a bugfix, or the next version later on, a new branch should be created.

docs/_pages/release-process.md
plata: I don't want to be more generic. The purpose of this page is that you have a check-list and don't forget anything. Therefore I'm willing to update it with any additional control file required. I don't think it will be that many. maybe flatpak, snap, rpm, OSX installer but that should be about it.

docs/_pages/release-process.md
plata: I will describe this as part of #1260.

docs/_pages/release-process.md
plata: Both, description is updated.

docs/_pages/release-process.md
plata: Yes, updated.

docs/_pages/release-process.md
plata: As you said: We use GitHub releases based on tags. So every tag is fixed. Therefore we can just release all releases which are based on the same master from the same branch instead of creating a new branch from the release branch. Apart from this, you see that we currently change the release branch because we detect bugs before actually triggering the release (the alternative to this would be to freeze master, which might stop development).

I'm open to change the way we do releases in the future but I think for now (alpha) we're ok.

PR #1259, opened 2018-04-19
Add issue template

plata:


madoar:
Maybe we should also ask for a screenshot, if it's a graphical issue?

plata:
I don't now. We'll find out if it should be added.

PR #1299, opened 2018-04-25
Revert "Bump net.revelc.code.formatter:formatter-maven-plugin from 2.0.1 to 2.7.2"

plata:
Reverts PhoenicisOrg/phoenicis#1213

Running `mvn clean package` changes the formatting of many files. The config file does not seem to be considered.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -168,7 +168,7 @@
             <plugin>
                 <groupId>net.revelc.code.formatter</groupId>
                 <artifactId>formatter-maven-plugin</artifactId>
-                <version>2.7.2</version>
+                <version>2.0.1</version>
                 <executions>
                     <execution>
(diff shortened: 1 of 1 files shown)


madoar:
Interesting, it seems like `mvn -Pcheck-formatted validate formatter:validate` works differently from `mvn clean package`.

plata:
Any ideas how we continue with this? I didn't find a description how the config should look like now.

madoar:
I'm not sure if an eclipse code format configuration is the right approach for this.

I think we should think about changing to a linter tool for Java, which works with maven.
Maybe http://checkstyle.sourceforge.net/ would work for us?

plata:
I don't quite understand. The current formatter works with maven already.

Issue #1236, opened 2018-04-08
Find better way to get class names for engine tools


plata:
Currently, the class name is computed in a pretty hacky way in `WinePrefixContainerWineToolsTab#populate`. It works as long as the ID is the snake case form of the class name. There must be a better solution for this.

plata:
The problem with objects is that you must know the class name to create them. I also don't see a nice way currently to get this information in a generic way.

Please have a look at the PR which is linked above. It provides a solution which is at least not as dirty as the current implementation.

madoar:
About the class name:
I think we can expect, that the class is named the same as the file, it's contained in, i.e. the same as is done in Java.

About the PR:
I think we can merge the PR but I think we should let this issue remain open, so that we can find a better long term solution.

plata:
That's fine for me.

The problem is: we do not know the file name. Also, one file can contain more than one class.

plata:
Sounds like a plan. Can we be sure that we always have a file name? Could it happen e.g. that we have have some weird repository which does not have file names (like command line input)?

madoar:
I would expect that we always have some kind of URI, which we use to access the script.
In the case of a command line or something similar, we need to generate an URI.
From an URI we should be able to extract something like a file name (even if the resource is not a file).

plata:
To summarize:
* When the repository is loaded, a `className` is stored for every `ScriptDTO`.
* By convention, the class exported by the file has the same name as the file.
* This `className` can be used in `eval` to create the desired object.

To proof the concept, we can update the engine tools implementation.

madoar:
Yes.
The remaining part is how to use the created object from the Java side of things.
I can think of three ways:
1) we can generate some javascript code, which executes the corresponding run method of the object.
2) we implement the javascript classes/objects by extending a Java interface via `Java.extend`. This should be the cleanest, but I think it can be quite hard to add extra functions to the javascript class, that aren't defined in the interface (which for example is required for wine)
3) we implement the javascript classes/objects like we do now, just that we define somewhere in the docu, what methods are required. Afterwards we need to access the created object via a `ScriptObjectMirror`, then we can access the methods of the object via the `ScriptObjectMirror` instance.

plata:
There's one big issue: If we follow the approach with our current repository structure, all classes will be named "script" (because all scripts are named "script.js").

madoar:
Yes, that's right. We would need to rename the files.

plata:
[...]

PR #1303, opened 2018-04-27
Change verb and application implementations to return JSON objects

madoar:
This PR targets the changes to the application and verb scripts on the Java application side, as described in #1236.

As a short note:
This PR should only be merged after a corresponding PR for the script changes has been made.

In addition I've asked myself if the CLI client works currently, because the passed parameters to the scripts are kind of different in the two application (i.e. JavaFX vs. CLI).
In the JavaFX application we pass the four ids: `TYPE_ID`, `CATEGORY_ID`, `APPLICATION_ID` and `SCRIPT_ID`.
For the CLI application I couldn't find code that sets these ids, before executing an application script.

Maybe we should try to make both method implementations more uniform.

### DIFF
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/EngineToolsManager.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/EngineToolsManager.java
@@ -39,6 +39,7 @@ public class EngineToolsManager {
 
     /**
      * constructor
+     *
      * @param scriptInterpreter
      */
     public EngineToolsManager(ScriptInterpreter scriptInterpreter) {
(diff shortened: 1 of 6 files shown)


plata:
Please link the PR in scripts when it's ready.

madoar:
I'm not sure if we should continue with the approach suggested in this PR.
If we decide to move to graal.js after its release, I think it is a better idea to first get graal.js running with phoenicis.
Afterwards we can decide how we want to improve the script representation.

It seems like the release candidates for graal.js have been published to [maven central](https://mvnrepository.com/artifact/org.graalvm).
In addition, there is a first version of a repository which demonstrates [graal.js with maven](https://github.com/graalvm/graal-js-jdk11-maven-demo).

plata:
I'm fine with that. Then feel free to close this. Can you copy this information to the "nashorn deprecated" issue?

PR #1311, opened 2018-05-05
Implement engine settings

### LINKED ISSUE #1309: generic engine settings for a container
Currently, the engine settings for a container (GLSL etc.) are implemented specifically for Wine in Java. We need a generic way to implement this in Javascript with a Java interface.

Rough idea:
Java interface `EngineSetting` with the methods:
* `String[] getOptions()`
* `String getCurrentOption()`
* `void setOption(String container, String option)`

Affected classes:
* `WinePrefixContainerDisplayTab`
* `WinePrefixContainerController`
* `org.phoenicis.containers/wine/**`, `WinePrefixContainerDTO` (should be deleted)

Open questions:
* How to handle translations? The options may not be translated when they are passed to/from Javascript but shall be translated in the GUI.

plata:
fixes #1309 

requires PhoenicisOrg/scripts#611


phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
```diff
@@ -0,0 +1,94 @@
...
+        for (TypeDTO typeDTO : repositoryDTO.getTypes()) {
+            if (typeDTO.getId().equals("engines")) {
+                categoryDTOS = typeDTO.getCategories();
+            }
+        }
+        for (CategoryDTO engine : categoryDTOS) {
+            final String engineId = engine.getId();
+            settings.put(engineId, new ArrayList<>());
+            for (ApplicationDTO applicationDTO : engine.getApplications()) {
+                if (applicationDTO.getId().equals("settings")) {
```
madoar: This id looks like insider knowledge to me.
Is there maybe another way to identify settings applications?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerEngineSettingsTab.java
```diff
@@ -60,14 +60,31 @@ private void populate() {
         final GridPane displayContentPane = new GridPane();
         displayContentPane.getStyleClass().add("grid");
 
+        int row = 0;
+        for (EngineSetting setting : engineSettings) {
+            final ComboBox<String> comboBox = new ComboBox<>();
+            comboBox.setMaxWidth(Double.MAX_VALUE);
+            comboBox.setValue(setting.getCurrentOption());
```
madoar: If we want to we could use `tr()` here.
The question is how do we collect the different options into the translation related files?

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
plata: I don't know. However, this issue is not new/special for settings. We have the same for engine tools and engines in general. I agree that a different solution would be preferable. Can you open a new issue for it?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerEngineSettingsTab.java
plata: I was thinking do solve the issue a bit differently: Instead of just passing a `String[]` we could use a `SettingValue` object which has a `value` and a `text`. Only the `text` would be translated.
Alternatively, we could also pass the index of the selected combobox item instead of the String to `setOption`.
Not sure what's the best solution.

madoar: I think it is important to implement this with the goal that writing a new `EngineSetting` is as easy as possible.

I think if we use your first option it will require that each settings defines multiple `SettingValue` objects, which need to be returned.
The second idea sounds good to me.

plata:
I have updated the pr accordingly. Can you check if this is the way we want to go before I start implementing the other settings?

plata:
There's a problem: the `regedit` plugin cannot delete registry entries currently. We need this to go back to default.

plata:
And even more important: I cannot retrieve the list of settings. No idea how I can aggregate the result of the lambda calls.

madoar:
What do you mean by aggregate the result of the lambda calls?

plata:
`fetchAvailableEngineSettingss` shall return all tools but currently the same tool is stored in the list several times. I assume it is because the different lambda calls to `getSetting` try to fill the same local variable.

plata:
I think I found it: the JSON files had all the same ID. It's horrible to debug the Javascript interfaces really.

PR #1321, opened 2018-05-08
Update "run in container"

### LINKED ISSUE #1261: Run executable file chooser: show HOME by default
### Expected behavior
File chooser opens in HOME or `~/.Phoenicis/containers`.

### Actual behavior
File chooser opens in the directory where Phoenicis has been started.

### Steps to reproduce
1. Start `phoenicis.sh` e.g. from `phoenicis-dist`.
2. Containers → select container → Tools → click "run executable"

plata:
open file chooser in container directory

fixes #1261

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerToolsTab.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerToolsTab.java
@@ -60,6 +60,13 @@ public class ContainerToolsTab extends Tab {
             this.lockAll();
             FileChooser fileChooser = new FileChooser();
             fileChooser.setTitle(tr("Choose executable"));
+
+            // open in container directory if it exists
+            File containerDir = new File(this.container.getPath());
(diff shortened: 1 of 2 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerToolsTab.java
```diff
@@ -60,6 +60,13 @@ private void populate() {
             this.lockAll();
             FileChooser fileChooser = new FileChooser();
             fileChooser.setTitle(tr("Choose executable"));
+
+            // open in container directory if it exists
+            File containerDir = new File(this.container.getPath());
+            if (containerDir.canRead()) {
```
madoar: How about we use the home directory of the user as a fallback?

plata: We could do that but to be honest, I don't think it makes much sense. I mean to get to this point, you must already choose an existing container so how can it be that the container directory does not exist?

PR #1326, opened 2018-05-08
Test .deb

### LINKED ISSUE #1297: Test .deb in Travis

plata:
fixes #1297


plata:
The caching does not seem to work. I have no idea why.

madoar:
I'm not too sure I understand what exactly travis should execute via this script.

What exactly are you caching?

madoar:
Can we maybe use docker for this instead?
I mean we could just use an ubuntu docker container, in which we execute the `.deb` file and the final application.
What do you think?

plata:
The problem is not the execution. The problem is that the .deb is not available in the cache. I've contacted Travis regarding this but they couldn't find a solution yet as well (it's working in a smaller test setup).

plata:
The caching works now but the installation does not (unmet dependencies).

plata:
I'm now using a docker container with Ubuntu 16.04 to install the .deb. That's working. @madoar @qparis anything else we can/should test here?

madoar:
I'm fine with this PR.

The only suggestion I have is to use a Dockerfile, instead of adding the commands directly to the travis configuration.
The benefit when using a Dockerfile is, that we can execute the image/container locally, instead of only being dependent on travis to execute it.

plata:
What should be in that Dockerfile? The .deb test currently only calls `apt install`.

plata:
To be honest, I don't quite get what you have in mind. Feel free to provide a PR with your idea.

PR #1329, opened 2018-05-09
Add sortpom Maven plugin

plata:


madoar:
This PR basically orders the POM files, right?

plata:
The sortpom plugin is something like a formatter/linter for pom.xml files. It ensures a certain order but also makes sure that the tags are indented properly.

PR #1333, opened 2018-05-10
Update view after engine install/delete

### LINKED ISSUE #1246: engine list is not updated after install/delete
The engine is not marked as installed/delete. Must restart Phoenicis to see the change.

plata:
fixes #1246 

Maybe not the most elegant solution but it gets the job done.

### DIFF
--- a/i18n/keys.pot
+++ b/i18n/keys.pot
@@ -51,27 +51,27 @@ msgstr ""
 msgid "Error while creating shortcut"
 msgstr ""
 
-#: phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java:63
-#: phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java:64
+#: phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java:65
+#: phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java:66
 msgid "Loading engines failed."
 msgstr ""
 
-#: phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java:81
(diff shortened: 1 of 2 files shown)


madoar:
As an alternative approach:
Can we add a redraw/repaint method to `PhoenicisTransformationList`, which triggers a redrawing of the elements inside the list?
I think this is the cleanest solution.

plata:
I don't think it would be that much cleaner (and I don't have an idea currently how to implement it). The optimal way would be some kind of dynamic list elements which change depending on their properties but that would probably mean some massive changes.
In general I'm not so happy with the strange duality we have there: On the one hand, there's the DTO tree and then there's the Javascript implementations of the interfaces. Right now, that's two parallel worlds.

madoar:
My idea is to view the JavaScript implementations of the interfaces as an extension of the DTO tree, i.e. the interfaces are part of the DTO tree.

The point with the DTOs is, that they should be immutable, so a change of state, i.e. an installation or uninstallation is hard to reflect.

What could be an alternative solution would be to use two observable lists:
- one containing all installed engines, and
- another containing all not installed engines

Both lists need to be merged into a single list, which is then sorted alphabetically.

Whenever an engine is installed it is moved from the list of not installed engines to the list of installed engines, and the other way around when the engine is uninstalled.

What do you think?

plata:
One problem is that we cannot sort the lists in Java because only the Javascript engine knows how the sorting shall be done.
I would suggest to merge this now to have the bugfix and then open a new issue to discuss a general concept for the interfaces.

madoar:
Why does only the Javascript engine know, how the sorting should be done?
I mean when we visualize the engines in the engines list, we visualize Java objects, these objects needs to have a getter method for a title field, which we use for sorting, right?

plata:
The sorting is done based on the version string. This version string is highly dependent on the specific engine implementation. For example, Wine uses "3.6" but another engine might use a completely different version scheme (like "a1", "a2"). So it's not possible to do the sorting in Java (it only works right now because we simply assume that all engines use the same versioning as Wine).

PR #1334, opened 2018-05-10
Travis CI: test with OpenJDK 8

### LINKED ISSUE #1188: Travis: test with OpenJDK

plata:
fixes #1188

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,9 @@ jobs:
       os: linux
       dist: trusty
       jdk: openjdk8
+      addons:
+        apt:
+          update: true
     # build + test
     - stage: Build + test
       os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ jobs:
         - docker exec -t ubuntu-16-04 bash -c "apt -qq update;
                                                apt install -y git;
                                                apt install -y maven;
-                                               apt install -y openjdk-8-jdk;
                                                apt install -y openjfx;
                                                apt install -y gettext"
       script:
@@ -45,7 +44,6 @@ jobs:
         - docker exec -t ubuntu-16-04 bash -c "apt -qq update;
                                                apt install -y git;
                                                apt install -y maven;
-                                               apt install -y openjdk-9-jdk;
                                                apt install -y openjfx;
                                                apt install -y gettext"
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ jobs:
 
     # build + test
     - stage: Build + test
+      jdk: openjdk8
       sudo: required
       services:
         - docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,32 @@ jobs:
                                                git clone https://github.com/PhoenicisOrg/phoenicis.git;
                                                cd phoenicis;
                                                mvn test -B"
+    - stage: Build + test
+      jdk: openjdk9
+      sudo: required
+      services:
+        - docker
+      before_install:
+        - docker run -d --name ubuntu-18-04 -v $(pwd):/travis ubuntu:18.04 tail -f /dev/null
+      install:
+        - docker exec -t ubuntu-18-04 bash -c "apt -qq update;
+                                               apt install -y git;
+                                               apt install -y locales;
+                                               apt install -y maven;
+                                               apt install -y openjdk-9-jdk;
+                                               apt install -y openjfx;
+                                               apt install -y gettext"
+        # fix encoding for locales, set environment correctly in next step
+        - docker exec -t ubuntu-18-04 bash -c "locale-gen en_US.UTF-8"
+      script:
+        - docker exec -t ubuntu-18-04 bash -c "cd /travis;
+                                               export LANG=\"en_US.UTF-8\";
+                                               export LANGUAGE=\"en_US:en\";
+                                               export LC_ALL=\"en_US.UTF-8\";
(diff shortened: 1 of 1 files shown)


plata:
The build works but there are character encoding problems.

madoar:
Do you have an idea how we can fix the encoding problems?

plata:
No. I don't even have an idea why they occur in the first place.

madoar:
@plata after a short search I've found [this](https://stackoverflow.com/questions/27931668/encoding-problems-when-running-an-app-in-docker-python-java-ruby-with-u), maybe the solution there works for us.

madoar:
I think it would be best to specify the commands we execute in the Dockerfile.
We can then build and run the container via `docker build` and `docker run`, see https://docs.docker.com/get-started/part2/#build-the-app for more details.

plata:
Now we get
```
package javafx.geometry does not exist
```

madoar:
I think `openjfx` is not installed?

plata:
You are right. Actually I was looking for it but then somehow mixed openjfx and openjdk and got confused because that's installed already :P

plata:
It's working now.

plata:
You cannot name jobs in Travis at the moment (see https://github.com/travis-ci/travis-ci/issues/5898).

OpenJDK 9 should just be copy+paste.

plata:
Seems like it's possible to set the `jdk` variable (it's not used but shows up in the table correctly).

plata:
@madoar do you have Ubuntu 18.04 installed? The OpenJDK 9 build doesn't like the `JAVA_HOME`.

madoar:
No, I'm using Linux Mint 18.3.

plata:
Any ideas what the problem could be?

madoar:
Are you a hundred percent sure, the JAVA_HOME path set in the travis file is the correct one?
It may be useful to add a `ls`, just to be sure.

plata:
Ok... Seems like there is no `openjdk-9-jdk` package.

plata:
Don't know why it's not working.

madoar:
Just asking:
Why don't we use an openjdk docker image? 
See https://hub.docker.com/_/openjdk/

plata:
We can try that. Which one should I use?

madoar:
I think we should start with `openjdk:8`.
If it works we can try `openjdk:9` and lastly `openjdk:10`.
The benefit of using these images should be, that you should not need to set the (java) paths.

plata:
Do we need a custom docker file for that? I'm not that familiar with Docker really. I thought you were talking about an Ubuntu image containing OpenJDK 8.

madoar:
I am talking about a docker image.
The docker image is called `openjdk`, while the tag afterwards specifies the to be used openjdk version.
So if you need an ubuntu image containing OpenJDK 8 you can use the image `openjdk:8`.

plata:
Seems to be the same problem.

madoar:
When using the `openjdk` image you shouldn't need to set `JAVA_HOME`.
In addition you can try removing the language parameters.

Have you tried creating and executing the containers locally?

plata:
No, I didn't try locally.

.travis.yml
```diff
@@ -15,6 +15,45 @@ jobs:
...
+      sudo: required
+      services:
+        - docker
+      before_install:
+        - docker run -d --name ubuntu-16-04 -v $(pwd):/travis openjdk:8 tail -f /dev/null
+      install:
+        - docker exec -t ubuntu-16-04 bash -c "apt -qq update;
+                                               apt install -y git;
+                                               apt install -y maven;
+                                               apt install -y openjdk-8-jdk;
```
madoar: This line isn't needed, when using the image `openjdk:8`

.travis.yml
```diff
@@ -15,6 +15,45 @@ jobs:
...
+      sudo: required
+      services:
+        - docker
+      before_install:
+        - docker run -d --name ubuntu-16-04 -v $(pwd):/travis openjdk:9 tail -f /dev/null
+      install:
+        - docker exec -t ubuntu-16-04 bash -c "apt -qq update;
+                                               apt install -y git;
+                                               apt install -y maven;
+                                               apt install -y openjdk-9-jdk;
```
madoar: This line isn't needed when using the image `openjdk:9`

plata:
Ok. Let's simply not support OpenJDK9 for now (this is clearly not a Phoenicis issue).

qparis:
Anyway, JDK 10 is out. I think we should support this one because there the memory management is optimized and also we could take advantage of type hintings. We do you think?

madoar:
@qparis the issue is not really JDK 9 or JDK 10.
The problem seems to be the way JavaFX (i.e. OpenJFX) works together with OpenJDK 9+.

As far as I understand it, up to OpenJDK 8 it was possible to overlay an OpenJFX build over an existing JDK, i.e. you could install OpenJFX after installing OpenJDK 8 on your system.

Starting with OpenJDK 9 this isn't possible anymore.
Now you need to build the (Open)JDK 9+ after installing OpenJFX.
When building OpenJDK 9+ you seem to need to set a reference to the OpenJFX installation/build. 

More details can be found [here](https://wiki.openjdk.java.net/display/OpenJFX/Building+OpenJFX#BuildingOpenJFX-IntegrationwithOpenJDK).

plata:
To add to @madoar: the whole issue seems to be related to the module system which has been introduced. Of course, we can decide to support *Oracle* JDK 10 but that's completely unrelated to this PR.

PR #1336, opened 2018-05-11
Implement ls

plata:


phoenicis-tools/src/main/java/org/phoenicis/tools/files/FileUtilities.java
```diff
@@ -38,6 +38,14 @@
     @Value("${application.user.tmp}")
     private String tmpDirectory;
 
+    public String[] ls(File directory) {
```
madoar: JavaDoc :)

plata: I've update the whole file.

Issue #1157, opened 2018-02-13
openjdk/openjfx on fedora


Tutul-:
I saw that one coming from a while but : it seem that openjdk/openjfx on fedora have some difficulty to run phoenicis. As far as I know, it's probably linked with the "No proprietary code" mojo of Fedora.
I'm missing webkit javafx library `libjfxwebkit.so` and get error symbol with native library. It's strange because I get one error about a native library symbol but the lib is installed
```
/usr/lib/jvm/openjfx/rt/lib/amd64/libjavafx_iio.so: undefined symbol: jpeg_resync_to_restart
```
I suppose libjfxwebkit.so use not full open source code :/
Can someone confirm ?

Tutul-:
Ok find something, webkit depend on media, and media depend on ffmpeg. And ffmpeg isn't trully open source so it isn't on the main repo. :/
I guess, one day, that many problem will make me change to some other distribution >>

plata:
Do you see any solution for this on our side? I wouldn't like to tell Fedora users that they just can't use Phoenicis.

Tutul-:
I try but for now rpmfusion (a repo with non-free and free-with-non-free-lib) is down so I can't check if openjfx is within. An other solution is to build openjfx on fedora. I use to build it when v5 started because openjfx wasn't enougth free

plata:
Have you tried: https://github.com/gferon/openjfx-fedora?

If you got it working, let me know how so that we can add an installation description for Fedora in the wiki.

Tutul-:
webkit compile but without media. I suppose it can work as long as we don't need it.
btw, I'll move to manjaro in the week, too many restriction with fedora for some software.

plata:
So you got it running without the media stuff?

ekultails:
Tutul, you're absolutely right. Fedora strives to be a fully open-source operating system for the community and that means not providing proprietary software in the default repositories. [1] Red Hat understands the usefulness of having additional repositories so in Fedora 28 users can opt-in to enable a few verified third-party repositories. [2]

The only reliable way to get Phoenicis working on Fedora is to use Oracle Java instead of OpenJDK. This includes the full JavaFX library. The official Fedora wiki page on how to install Oracle Java should be linked where ever the build documentation for Phoenicis is. [3]  I have tested this out and installed a few Windows applications with Phoenicis and it works great.

1. https://fedoraproject.org/wiki/Forbidden_items?rd=ForbiddenItems
[...]

PR #1351, opened 2018-06-14
Document that currently Oracle Java is required to build and run Phoenicis on Fedora.

ekultails:
As discussed here: https://github.com/PhoenicisOrg/phoenicis/issues/1157


qparis:
Hum, this is a problem because we promesed that POL would stay fully open source. This problem needs to be solved

plata:
As I said in #1157: if we know which component is required, we can check if we really need it.

PR #1353, opened 2018-06-17
Fixed deb package

qparis:


qparis:
I don't understand how is my build making any difference for this test...

qparis:
@plata Any idea on what's happening here? I need this PR for alpha release

madoar:
The issue seems to be a hiccup of travis.
After restarting the tests they successfully finished.

About the shell script in general:
Do we really need such a script?
Isn't it easier to simply include all required java dependencies in the jar file and then simply execute it via `java -jar Phoenicis.jar`?

qparis:
Once Phoenicis is installed, users expect a path command to run it in general

madoar:
I don't understand.
How does setting the classpath and providing a shell script help to run Phoenicis?
The only thing that is really required is the command `java -jar Phoenicis.jar`.
The shell script only functions as a kind of wrapper for this single command, right?

qparis:
Yep, exactly, and it avoids classpath problems also. This is really a standard ; when you install a package, you expect to have a command to run your program in your path. It is not user friendly if you need to guess how to run it.

qparis:
Can we merge this pull request?

plata:
I think this fix should be done for the alpha branch as well.

PR #1358, opened 2018-06-23
Fix .deb package

plata:
see #1353


plata:
@qparis Travis passed, so `mvn test` must have been successful. I don't understand why you had a problem with it.

plata:
With this merged, the alpha branches should now really be ready for release @qparis.

PR #1367, opened 2018-07-27
Add build description for Ubuntu 18.04 and Linux Mint 19

RattyDAVE:
Added Ubuntu 18.04 and Linux Mint 19 install instructions.
Added Fix for steam not downloading.


docs/_docs/Users/build.md
```diff
@@ -5,6 +5,95 @@ order: 1
...
+
+* Clean up the existing config
+
+`rm -R ~/.Phoenicis`
+
+* Run! This is the only line you need from now on.
+
+`~/build/phoenicis/phoenicis-dist/target/phoenicis-dist/phoenicis.sh`
+
+There is an issue with the latest steam client not downloading. See https://bugs.winehq.org/show_bug.cgi?id=45329
```
plata: These Steam instructions do not belong here (this is only about building Phoenicis). We should consider updating [SteamScript](https://github.com/PhoenicisOrg/scripts/blob/master/Engines/Wine/QuickScript/Steam%20Script/script.js) instead.

docs/_docs/Users/build.md
```diff
@@ -5,6 +5,95 @@ order: 1
 toc: true
 ---
 
+## Ubuntu 18.04 and Linux Mint 19
```
plata: In general: can you follow the description for 16.04 more closely (just to have a common "style")? Especially: use the same steps/wording if possible and use code blocks.

docs/_docs/Users/build.md
```diff
@@ -5,6 +5,95 @@ order: 1
...
+
+* Remove exising version of wine
+
+`apt-get purge wine* libwine`
+
+* Update the repository information
+
+`sudo apt-get update`
+
+* Install wine packages (This is needed to stop the empty dialog boxes)
```
plata: I would prefer: "Install Wine devel packages (fix empty dialog boxes)".

@qparis shouldn't the dependencies be provided by our Wine builds?

docs/_docs/Users/build.md
```diff
@@ -5,6 +5,95 @@ order: 1
...
+
+`sudo /var/lib/dpkg/info/ca-certificates-java.postinst configure`
+
+* Add the Wine repository
+
+`sudo apt-add-repository https://dl.winehq.org/wine-builds/ubuntu/`
+
+`wget -nc https://dl.winehq.org/wine-builds/Release.key && sudo apt-key add Release.key`
+
+* Remove exising version of wine
```
plata: Why is removing the existing Wine version required?

RattyDAVE: So that the dialog boxes will be displayed

RattyDAVE:
Feel free to use these instructions as you see fit.

If you look the various forums for Ubuntu 18.04, Xbuntu 18.04 and Linux Mint 19 you will see issues. This is how I got it to work. 

It seems 18.04.1 was released yesterday so I cant comment on that.

docs/_docs/Users/build.md
plata: @madoar @qparis what do you think?

docs/_docs/Users/build.md
```diff
@@ -105,6 +146,18 @@ bash ./phoenicis-dist/phoenicis.sh
 
 ## Troubleshooting
 
+### Maven dependencies cannot be downloaded
```
madoar: As a short note:
Sometimes it also helps to delete the `.m2` folder in your home folder.
But this is a drastic measure, which basically resets maven and forces it to redownload all dependencies.

docs/_docs/Users/build.md
RattyDAVE: I have also tested with Ubuntu 18.04.1 and it still needs all this work.

plata: Ok. Anyways, I've opened an issue to figure out why this is necessary. Until that, the current description containing all steps is merged to master.

PR #1389, opened 2018-09-15
Build self-contained .jar for JavaFX application

plata:
Uses [javafx-maven-plugin](https://github.com/javafx-maven-plugin/javafx-maven-plugin) to build a .jar in `phoenicis-javafx/target/jfx/app`. This can be started simply with `java -jar`.

Might also be useful for snap/flatpak.

### DIFF
--- a/phoenicis-javafx/pom.xml
+++ b/phoenicis-javafx/pom.xml
@@ -151,13 +151,14 @@
                     <execution>
                         <phase>package</phase>
                         <goals>
-                            <goal>jar</goal>
+                            <goal>native</goal>
                         </goals>
                     </execution>
                 </executions>
                 <configuration>
                     <vendor>PhoenicisOrg</vendor>
                     <mainClass>org.phoenicis.javafx.JavaFXApplication</mainClass>
+                    <bundler>linux.app</bundler>
                 </configuration>
             </plugin>
         </plugins>
--- a/phoenicis-javafx/pom.xml
+++ b/phoenicis-javafx/pom.xml
@@ -108,6 +108,15 @@
             <version>1.1.10</version>
         </dependency>
     </dependencies>
+    <pluginRepositories>
+        <pluginRepository>
+            <snapshots>
+                <enabled>true</enabled>
+            </snapshots>
+            <id>oss-sonatype-snapshots</id>
+            <url>https://oss.sonatype.org/content/groups/public/</url>
+        </pluginRepository>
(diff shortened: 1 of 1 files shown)


madoar:
The plugin seems to have some problems in combination with Java 9

plata:
For whatever reason, it cannot find javafxpackager.

plata:
@FibreFoX any hints what could be the problem?

FibreFoX:
@plata the javafx-maven-plugin is not compatible with that JDK version, in the current state there is only proper support for JDK 8, but you might want to check using the somewhat dusty SNAPSHOT-version `8.9.0-SNAPSHOT` of the plugin.

One of the problem was the changed location of `ant-javafx.jar` file, in JDK 10 this became a lot of harder (because the classes were packaged in the new `jmod`-thing. (Getting even absurd with JDK 11 because it's not bundled within the OracleJDK anymore).

plata:
@FibreFoX Thanks for the info. Any idea about bundling only JRE/JFX without using a native bundler (I'd like to use this for flatpak).

FibreFoX:
@plata I'm not sure I understood that correctly, so you want to just have the package bundled like calling `mvn jfx:jar`, right? Then you have several options:
* You might be able to change the dependency inside the plugin, but this has to be done via maven profiles (the ugly way, because inside plugins, exclusions aren't working that easy, https://issues.apache.org/jira/browse/MNG-2163).
* You might want to try using the SNAPSHOT-version (recommended), as this has this done already.
* Use a different maven-plugin which does this all.

plata:
@FibreFoX yes, like `mvn jfx:jar` just with an JRE which I can bundle. Can you suggest a plugin which does this? I've searched quite a bit on the Internet but didn't find anything.

FibreFoX:
Before doing more suggestions:
* So do you need a stripped down JRE (using jlink)?
* Do you need any native launcher?
* Why even switching from `javafx-maven-plugin`? Calling `mvn jfx:jar` with the SNAPSHOT-version should work without any need to change a lot.

Because JDK 9 was only a small intermediate release (oracle changed their release-cycle), please avoid using JDK 9 any more, if possible directly switch to JDK 10.

plata:
@FibreFoX so it seems to be working with the snapshot version. However, I don't get the jlink part. Can I execute it with `javafx-maven-plugin` or do I need something additional like `maven-jlink-plugin`?

PR #1391, opened 2018-09-16
Catch launch error

plata:
avoid that application fails silently

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
@@ -30,7 +30,11 @@ import org.springframework.context.annotation.AnnotationConfigApplicationContext
 public class JavaFXApplication extends Application {
 
     public static void main(String[] args) {
-        Application.launch(JavaFXApplication.class);
+        try {
+            Application.launch(JavaFXApplication.class);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
     }
 
     @Override


phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
```diff
@@ -30,7 +30,11 @@
 public class JavaFXApplication extends Application {
 
     public static void main(String[] args) {
-        Application.launch(JavaFXApplication.class);
+        try {
+            Application.launch(JavaFXApplication.class);
+        } catch (Exception e) {
+            e.printStackTrace();
```
madoar: I think we should log the exception to a log file instead of printing it.
The current behavior (not catching the exception) should lead to an automatic print of the exception to the terminal/console.

plata: I've added this because I got an error while playing around with flatpak. The only output I got was:
```
phoenicis-javafx-5.0-SNAPSHOT Error invoking method.
phoenicis-javafx-5.0-SNAPSHOT Failed to launch JVM
```
Only after adding this `try catch`, the exception got printed.

Apart from this, I'm not sure that using a logger is possible here. The errors that can happen at this point are so basic that the logger might not even be available yet.

madoar: I don't really see why a logger shouldn't be usable here?
Everything you specify on a `static` scope should be usable in the main method.

plata: Maybe you're right. I just remember that I saw in Stackoverflow that they were checking if the logger is available. Might not be required in our case. Anyways, we can always add the logger later on.

PR #1398, opened 2018-09-25
Travis: build with Oracle JDK 10

plata:


plata:
@madoar any ideas regarding this issue with monocle? I think you fixed it for JDK9 as well.

madoar:
If I remember correctly, the trick was to define a profile for JDK10, which specifies the correct version of `openjfx-monocle` for Java 10.
The only problem: I can't find a version of `openjfx-monocle` for Java 10, only for Java 9 and 10.

plata:
So no JavaFX tests with Java > 9? That cannot be true...

madoar:
The main issue seems to be, that no monocle dependency could be found.
The reason for this is most likely, because no maven profile has been specified for it.
Can you please go to https://github.com/PhoenicisOrg/phoenicis/blob/master/phoenicis-javafx/pom.xml and copy the profile for jdk9 and modify it to be specific for jdk10.

plata:
Like you said, there seems to be no monocle version for Java 10...

madoar:
Have you tried the one for JDK9?

plata:
Seems to work.

PR #1400, opened 2018-09-25
Travis: build with OpenJDK 9

plata:


plata:
So here we have our famous OpenJDK 9 + JavaFX issue. Do we even want to support JDK9 or do we go to 10/11 directly?

madoar:
All in all I think we support Java 9 and 10.
The main problem is in my opinion, that there is no good documentations on how to install OpenJFX with OpenJDK 9 and 10.

I believe we should directly jump to Java 11.
With this thought in my it is good news that Java 11 released today.
The only problem we'll face is, like you previously mentioned somewhere, that most (all?) current linux distributions don't contain Java 11 in their default packages.

plata:
Isn't JDK11 available e.g. in Ubuntu 18.04? JFX11 can be installed via maven. The bigger problem is Nashorn. I think it has been removen in JDK11.

madoar:
Nashorn should still be contained in JDK11.
The JEP says, that it is only to be deprecated.

I have no clue about when JDK11 will be available in the linux distributions.

plata:
Ok. Then let's try it.

PR #1404, opened 2018-09-26
Add support for JDK 11

### LINKED ISSUE #1368: Java 11
Starting with JDK 11 OpenJFX won't be included in the JDK anymore.
Therefore, we should think about how we can support OpenJFX 11 in the future, and how we can provide support for Java 10- and Java 11+ at the same time.

I've also created an issue asking the last question at the OpenJFX github repository, see [here](https://github.com/javafxports/openjdk-jfx/issues/141).

plata:
fixes #1368

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -193,6 +193,10 @@
                 <version>3.8.0</version>
                 <configuration>
                     <release>10</release>
+                    <compilerArgs>
+                        <arg>--add-modules=java.xml.ws.annotation</arg>
+                        <arg>--add-modules=java.activation</arg>
+                    </compilerArgs>
                 </configuration>
             </plugin>
             <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -284,25 +284,6 @@
             </build>
         </profile>
         <profile>
-            <id>jdk9</id>
-            <activation>
-                <jdk>9</jdk>
-            </activation>
-            <build>
-                <plugins>
-                    <plugin>
-                        <groupId>org.apache.maven.plugins</groupId>
-                        <artifactId>maven-compiler-plugin</artifactId>
-                        <version>3.8.0</version>
-                        <configuration>
-                            <source>1.8</source>
-                            <target>1.8</target>
-                        </configuration>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
-        <profile>
             <id>jdk8</id>
             <activation>
                 <jdk>1.8</jdk>
--- a/pom.xml
+++ b/pom.xml
@@ -245,13 +245,19 @@
                         <configuration>
                             <release>11</release>
                             <compilerArgs>
-                                <arg>--add-modules=java.xml.ws.annotation</arg>
                                 <arg>--add-modules=java.activation</arg>
                             </compilerArgs>
                         </configuration>
                     </plugin>
                 </plugins>
             </build>
+            <dependencies>
+                <dependency>
+                    <groupId>org.glassfish.jaxb</groupId>
+                    <artifactId>jaxb-runtime</artifactId>
+                    <version>2.4.0-b180830.0438</version>
+                </dependency>
+            </dependencies>
         </profile>
         <profile>
             <id>jdk10</id>
--- a/pom.xml
+++ b/pom.xml
@@ -250,9 +250,9 @@
             </build>
             <dependencies>
                 <dependency>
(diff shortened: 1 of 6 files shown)


madoar:
This won't be enough I think.
You'll need to add JavaFX via maven, to get it working, see https://openjfx.io/openjfx-docs/ for details.

plata:
I know. I will update this step by step. Any help welcome because sometimes the errors are quite strange.

pom.xml
```diff
@@ -192,8 +192,7 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.0</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <release>10</release>
```
madoar: What does the 10 mean?

plata: It's the Java version which shall be used. It should be 11 but somehow that doesn't work for me. See https://openjfx.io/openjfx-docs/#maven

madoar: Won't this fail automatically for Java 8 and 9?

plata: I guess so. Maybe I can add a profile for this as well.

plata: Right now, I would like to get Java 11 working. Then we can think about what we want to support (Java 8 will have EOL in 2019 anyways, so...).

pom.xml
```diff
@@ -240,6 +231,96 @@
...
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-compiler-plugin</artifactId>
+                        <version>3.8.0</version>
+                        <configuration>
+                            <release>10</release>
+                            <compilerArgs>
+                                <arg>--add-modules=java.xml.ws.annotation</arg>
```
madoar: Is there a reason, why you only add this to JDK10 and not to JDK11?

plata: It doesn't work for JDK11. JDK11 uses the dependencies instead (see last 2 commits).

qparis: Is it different from including java.xml.bind and javax.activation modules dependencies?

madoar: An additional question:
Which of our maven dependencies requires these modules?
I mean as far as I know these modules are part of some Java EE APIs, which have been removed in JDK11 from Java SE.
This indicates, that we have some maven dependencies, which require Java EE APIs.
Are these dependencies Java 11 compatible?

plata: @qparis the functionality should be the same. However, the modules are available in Java EE until JDK 10 (though deprecated) and have been removed in JDK 11. So technically, the maven dependencies are probably a bit different from the modules in Java 10.

@madoar as far as I understand, the maven dependencies are made to provide the removed functionality of the Java EE APIs for Java versions where those have been removed. So for Java 11, we do not use the Java EE APIs anymore but instead rely on those maven dependencies.

pom.xml
```diff
@@ -240,6 +231,77 @@
...
+            </build>
+            <dependencies>
+                <dependency>
+                    <groupId>org.glassfish.jaxb</groupId>
+                    <artifactId>jaxb-runtime</artifactId>
+                    <version>2.4.0-b180830.0438</version>
+                </dependency>
+                <dependency>
+                    <groupId>com.sun.activation</groupId>
+                    <artifactId>javax.activation</artifactId>
```
madoar: Another idea:
Try to add both dependencies to the project, where they are missing, i.e. to `phoenicis-multithreading`.

plata: Ok.

plata:
There's a problem with OracleJDK 10 and OpenJDK 10: they need different profiles. The current config works for OpenJDK 10, the one on master works for OracleJDK 10.

plata:
I think it's the best solution to drop Oracle JDK 10 support. I could create different profiles but it's not possible to detect Oracle JDK or OpenJDK in the profiles automatically (at least as far as I know). So you would have to select the profile manually and could not build Phoenicis automatically on e.g. Ubuntu 18.04. Apart from that, you can still use Oracle JDK 8 or 11 if you want Oracle JDK.

madoar:
What is the reason for removing JDK9?

plata:
It's only an intermediate release. End of support is March 2018. So we should really concentrate on LTS (JDK 11) and go with Java 10 until that's really available (should be in the next days for Ubuntu 18.04 etc).

madoar:
No problem, I just wanted to know the reason :)

PR #1406, opened 2018-09-29
Add Flatpak support

### LINKED ISSUE #1190: provide flatpak
general flatpak tutorial: http://docs.flatpak.org/en/latest/first-build.html

old flatpak tutorial for Java apps: https://github.com/flatpak/flatpak.github.io/blob/81bb006fd96b9a13170bd876316eb5a4eb48244b/source/bundling-a-java-app.html.haml.markdown

good example (pycharm): https://github.com/flathub/com.jetbrains.PyCharm-Community/blob/f9e57ce185a4c2f6a171f9e5b5f3ba2acd36a475/com.jetbrains.PyCharm-Community.json

plata:
The manifest file is located in `phoenicis-dist/src/flatpak`.

Required extensions:
- org.freedesktop.Platform.Compat.i386
- org.freedesktop.Platform.GL32.nvidia-410-73 (or similar)

To build:
```
cd phoenicis-dist/src/flatpak/
flatpak-builder build-dir org.phoenicis.javafx.json --force-clean
```
Run with:
```
cd phoenicis-dist/src/flatpak/
flatpak-builder build-dir org.phoenicis.javafx.json --force-clean --user --install
flatpak run org.phoenicis.javafx phoenicis-javafx
```

fixes #1190


plata:
@mbooth101 I'm installing OpenJFX 11 to use it with your OpenJDK 10 extension. It works but I think it would be nice to have an OpenJFX extension as well. Do you think that would be feasible?

madoar:
In my opinion we should wait until a `flatpak` module for Java 11 exists.
Such a module has the big advantage, that we won't need a separate dependency to JavaFX 11, because it can be included in our application (jar-file) via maven.

plata:
Would be nice, yes. Did you take a look at the JDK 10 extension? How hard can it be do adopt this for 11?
In any case the open issues of this pr (see description) must be solved.

madoar:
Yes, I've taken a look at the JDK 10 extension at https://github.com/flathub/org.freedesktop.Sdk.Extension.openjdk10, which truth be told is quite hard to find on google.
I'm not sure how hard it is to adapt the script to JDK 11.
My problem with the extensions are that they look very hacky to me, where some commands are modified to work specifically with the given JDK version.
I'm not sure how transferable these commands are to JDK 11.

You're right, the most important open task is the open issue.
Could the problem be missing permissions for the sandbox?
For example to download wine?

plata:
I assume that it's a permission issue as well. However, I've no idea what exactly the problem is. https://github.com/PhoenicisOrg/phoenicis/blob/1abb8bb81c6f94699eac4f3b31d657053f246893/phoenicis-dist/src/flatpak/org.phoenicis.javafx.json#L14 should grant the required permissions (at least to my understanding). I've also tried the switch for permissions for home specifically, but that didn't help either. Phoenicis can read the exisiting library so it cannot be that it doesn't have rights at all. Downloading can also not be the issue because the required Wine version is installed already. Might be that execution permissions (for the Wine executable) require a special setting but I don't know what.

madoar:
I may have found the source for the problem.
According to http://docs.flatpak.org/en/latest/sandbox-permissions.html, `host` blacklists the `/tmp/` folder.
Isn't this the folder, where we download wine and other files?

plata:
I think we have a tmp inside `.Phoenicis` but I have to check.

Maryse47:
> flatpak-builder --run build-dir org.phoenicis.javafx.json phoenicis-javafx

`flatpak-builder --run` is known to be broken (it doesn't match flatpak environment correctly. Use `flatpak-builder --force-clean  --install --user build-dir org.phoenicis.javafx.json` instead then test with `flatpak run org.phoenicis.javafx.json phoenicis-javafx`.

plata:
@Maryse47 I've noticed that there are differences already but even with the installed flatpak the issue remains.

mbooth101:
> @mbooth101 I'm installing OpenJFX 11 to use it with your OpenJDK 10 extension. It works but I think it would be nice to have an OpenJFX extension as well. Do you think that would be feasible?

Hi @plata 

Maybe -- I have no exposure to openjfx. How is it built?

BTW, there will be openjdk11 extension very soon: https://github.com/flathub/flathub/pull/669

plata:
@mbooth101 great! Looking forward to the OpenJDK 11 extension. I don't know how OpenJFX is built but if you look at my manifest in this PR, I'm just downloading and unpacking it which seems to be working (at least the GUI starts etc). I was thinking that it might be possible to do just that in an own extension for OpenJFX. Correct me if I'm wrong.

mbooth101:
I would just bare in mind that if it has native parts it may need to be built from source in order to support all the Flatpak architectures.

madoar:
Did you receive an error message/exception?

plata:
It said something about "<NAK". Not sure if that's an the error but I couldn't see anything else really (fetching the branches etc seemed to work ok). Can you give it a try? I think you know the git repository quite well.

LeandroStanger:
@plata Update runtime to 18.08

plata:
@LeandroStanger doesn't help, unfortunately.

madoar:
@plata I'm not sure I can really help here.
I haven't used flatpak yet, so I'm not really sure how to use it correctly.
Does Phoenicis provide any kind of error message?

Out of interest, can a flatpak installation conflict with a `deb` installation?

plata:
Strange. It worked now. I think it just took longer.

plata:
Now I get:
```
java.lang.RuntimeException: java.io.IOException: Cannot run program "/home/plata/.var/app/org.phoenicis.javafx/data/.Phoenicis/engines//wine/upstream-linux-x86/3.0.3/bin/wine" (in directory "/home/plata/.var/app/org.phoenicis.javafx/data/.Phoenicis/containers/wineprefix/Notepad"): error=2, file not found
```
The file exists.

plata:
I've tried to investigate the exception: it happens when `luna` tries to run `regedit`. However, the prefix exists so `wineboot` must have worked. I don't really understand what the difference should be between `wine wineboot` and `wine regedit`.

Maryse47:
I recommend to open PR in https://github.com/flathub/flathub/pulls . The most capable people to help are there.

phoenicis-dist/src/flatpak/org.phoenicis.javafx.appdata.xml
```diff
@@ -0,0 +1,57 @@
...
+  <summary>Install and run Windows software on Linux</summary>
+  <summary xml:lang="de">Installiere und starte Windows Software unter Linux</summary>
+  <description>
+    <p>Phoenicis is a front-end for Wine.</p>
+  </description>
+  <description xml:lang="de">
+    <p>Phoenicis ist ein Front-End für Wine.</p>
+  </description>
+  <categories>
+    <category>Game</category>
```
madoar: I'm not sure if `Game` is the correct category for Phoenicis.
Maybe something like `Utility` fits better.

phoenicis-dist/src/flatpak/org.phoenicis.javafx.appdata.xml
```diff
@@ -0,0 +1,57 @@
...
+  <id type="desktop">org.phoenicis.javafx</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <name>Phoenicis</name>
+  <summary>Install and run Windows software on Linux</summary>
+  <summary xml:lang="de">Installiere und starte Windows Software unter Linux</summary>
+  <description>
+    <p>Phoenicis is a front-end for Wine.</p>
+  </description>
+  <description xml:lang="de">
+    <p>Phoenicis ist ein Front-End für Wine.</p>
```
madoar: Are the translations manageable via Crowdin?

phoenicis-dist/src/flatpak/org.phoenicis.javafx.desktop
```diff
@@ -0,0 +1,12 @@
+[Desktop Entry]
+Name=Phoenicis JavaFX
+Comment=Front-end for Wine
+Exec=phoenicis-javafx %U
+StartupNotify=true
+Terminal=false
+Type=Application
+MimeType=application/x-ms-dos-executable;application/x-msi;application/x-ms-shortcut;
+Icon=org.phoenicis.javafx
+Categories=Game
```
madoar: Same as above

phoenicis-dist/src/flatpak/org.phoenicis.javafx.desktop
```diff
@@ -0,0 +1,12 @@
...
+Keywords=Phoenicis;
```
madoar: PlayOnLinux, Wine

phoenicis-dist/src/flatpak/org.phoenicis.javafx.json
```diff
@@ -0,0 +1,123 @@
...
+    },
+    {
+      "name": "openjdk",
+      "buildsystem": "simple",
+      "build-commands": [
+        "/usr/lib/sdk/openjdk11/install.sh"
+      ]
+    },
+    {
+      "name": "openjfx",
```
madoar: Is it still required to install OpenJFX separately?
Starting with Java 11 it should be possible to install JavaFX/OpenJFX via maven.

plata: Doesn't work somehow. I get:
```
Error occurred during initialization of boot layer
java.lang.module.FindException: Module javafx.controls not found
```

phoenicis-dist/src/flatpak/org.phoenicis.javafx.appdata.xml
plata: I don't think so (it's in the same file and not listed in their [supported formats](https://support.crowdin.com/supported-formats/)).

phoenicis-dist/src/flatpak/org.phoenicis.javafx.desktop
madoar: I'm not sure how the keywords need to be split.
If they are split with a `;`, the separator between PlayOnLinux and Wine is wrong.

phoenicis-dist/src/flatpak/org.phoenicis.javafx.json
madoar: I think this indicates a problem with the `java` command and/or the project setup for Java 11.
Have you only removed the 
```
{
      "name": "openjfx",
      ...
}
```
block?
Have you also removed the `-p /app/javafx-sdk-11/lib` argument to the `java` command?

phoenicis-dist/src/flatpak/org.phoenicis.javafx.desktop
plata: Fixed.

phoenicis-dist/src/flatpak/org.phoenicis.javafx.json
plata: The problem is: I have only JDK10 here (Ubuntu 18.04). If you check #1408, the maven dependencies do not seem to be enough (according to the tests of @ImperatorS79).

phoenicis-dist/src/flatpak/org.phoenicis.javafx.json
```diff
@@ -0,0 +1,123 @@
+{
+  "app-id": "org.phoenicis.javafx",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "18.08",
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions": [
+    "org.freedesktop.Sdk.Extension.openjdk11"
+  ],
+  "add-extensions": {
```
jjedelsky: I think there should be extension for 32bit libGL `org.freedesktop.Platform.GL32`. Without it, 32bit wine may be unable to use NVidia libGL.
See how [steam flatpak](https://github.com/flathub/com.valvesoftware.Steam/blob/47015a67c4a79a7dc98d29f79c31f00388a9801b/com.valvesoftware.Steam.json#L36) adds 32bit extensions.

plata: I've actually tried that but I always got an error when I tried to `flatpak install` this. Maybe I just did it wrong. Can you try?

madoar:
@plata if this works as intended then I'm fine with merging this PR now.
Still I think we should at least open a new issue for the JavaFX 11 issue (i.e. an installed SDK is required, which looks like a bug/configuration problem to me) or fix it in this PR.

plata:
I would like to merge this. The JavaFX issue is also there if you try to run Phoenicis built with JDK 11. I don't think it's related to Flatpak.

PR #1431, opened 2018-10-20
Prevent cloning git repository from different threads in parallel

### LINKED ISSUE #1402: Folder '...' is no git-repository
### Expected behavior
Phoenicis starts correctly

### Actual behavior
Phoenicis doesn't start correctly with an error message containing: `Folder '...' is no git-repository`. (see also #1401)

### Steps to reproduce
This issue occurs sometimes. I'm not really sure what the reason for the bug is, but in all cases I know of, a manual deletion of the folder and a restart of Phoenicis has solved the issue.

plata:
fixes #1402

### DIFF
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.util.concurrent.Semaphore;
 
 public class GitRepository implements Repository {
     private final static Logger LOGGER = LoggerFactory.getLogger(GitRepository.class);
@@ -44,6 +45,7 @@ public class GitRepository implements Repository {
     private final LocalRepository.Factory localRepositoryFactory;
 
     private final File localFolder;
+    static Semaphore mutex = new Semaphore(1);
 
     public GitRepository(URI repositoryUri, String branch, String cacheDirectoryPath,
             LocalRepository.Factory localRepositoryFactory, FileUtilities fileUtilities) {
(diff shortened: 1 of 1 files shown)


plata:
@pmorch can you validate if this works for you?

plata:
I agree. However, the code fixed the issue for me. Might also be a very strange coincidence.

plata:
Just read again: are we sure that there is only one `GitRepository` object? In any case, we must use `static synchronized` (the directory on the filesystem is not bound to a Java object).

plata:
I cannot reproduce the issue anymore. @pmorch would be great if you could also check if `static synchronized` is sufficient (it should be).

madoar:
Then wouldn't it be better to [lock](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#lock()) the file/folder to prevent other threads and applications from overriding the file/folder?

madoar:
The problem with the solution provided by this PR is, that I don't see how we can test the fix.
In addition the fix provided by this PR prevents Phoenicis from cloning multiple different repositories at the same time, which could lead to an increase in performance.

What I think would be a better approach is an approach comprising of a map consisting of a unique lock object for each distinct git repository.
We can then lock the procedure for each distinct git respository by calling:
```java
synchronized (getDistinctLockObject(repositoryUrim, ...)) {
   // clone git repository
}
```

Still this approach has the same problem in that we can't really ensure that it fixes the issue, but I think it is cleaner than the semaphore one.
The most ideal I think would be an approach securing the local repository folder across multiple applications.
For this we would require an additional file in the local git folder, for which we can acquire a file-lock.

plata:
We should go for the following approach:
1. Try to reproduce the issue in a VM as described by @pmorch.
2. Check if `static synchronized` or the current solution in this pr really fixes the issue and merge it as a quick fix.
3. Open a new issue to implement a secure solution also for multiple applications running in parallel.

I will not have time to do this in the near future, so it would be great if someone could take over.

qparis:
I don't know if it was you had in mind: instead of using a static synchronized method, I would do as madoar said, but I would use a static Object as a lock object. It should help

pmorch:
Hi,
Sorry for the late reply.
I just tried to build it from source, and when I run it I get:
```
$ ./phoenicis-dist/phoenicis.sh 
Error: JavaFX runtime components are missing, and are required to run this application
```
I have to admit I'm not really interested in pursuing this any further. I've already outlined how to reproduce this from vanilla 18.04, and I'll leave it at that.

plata:
@qparis @madoar can you test this in a VM? My virtual box supports only 32bit OSs, so I cannot try it.

PR #1437, opened 2018-10-26
Update deb generation

ImperatorS79:
Fix a lot of Lintian errors/warnings.

### DIFF
--- a/phoenicis-dist/pom.xml
+++ b/phoenicis-dist/pom.xml
@@ -72,6 +72,7 @@
                         </goals>
                         <configuration>
                             <outputDirectory>${project.build.directory}/lib</outputDirectory>
+                            <!--<excludeArtifactIds>javafx-web, javafx-media, javafx-graphics, javafx-controls, javafx-base</excludeArtifactIds> Avoid to copy those libs when system have native openjfx-11-->
                         </configuration>
                     </execution>
                 </executions>
@@ -93,8 +94,8 @@
                                     <type>file</type>
                                     <mapper>
                                         <type>perm</type>
-                                        <prefix>/usr/local/bin</prefix>
-                                        <filemode>775</filemode>
+                                        <prefix>/usr/bin</prefix>
+                                        <filemode>755</filemode>
                                     </mapper>
                                 </data>
                                 <data>
@@ -110,8 +111,8 @@
                                     <type>file</type>
                                     <mapper>
                                         <type>perm</type>
-                                        <prefix>/usr/local/bin</prefix>
-                                        <filemode>775</filemode>
+                                        <prefix>/usr/bin</prefix>
+                                        <filemode>755</filemode>
                                     </mapper>
                                 </data>
                                 <data>
@@ -127,7 +128,7 @@
                                     <type>directory</type>
                                     <mapper>
                                         <type>perm</type>
-                                        <prefix>/usr/local/lib/phoenicis</prefix>
+                                        <prefix>/usr/share/phoenicis-playonlinux</prefix>
                                     </mapper>
                                 </data>
                                 <data>
--- a/phoenicis-dist/src/deb/control/control
+++ b/phoenicis-dist/src/deb/control/control
@@ -1,13 +1,13 @@
-Package: Phoenicis-PlayOnLinux
+Package: phoenicis-playonlinux
 Version: 5.0
 Section: misc
 Priority: optional
 Architecture: all
 Description: This program is a front-end for wine.
-              It permits you to install Windows Games and softwares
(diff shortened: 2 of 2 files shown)


phoenicis-dist/pom.xml
```diff
@@ -72,6 +72,7 @@
                         </goals>
                         <configuration>
                             <outputDirectory>${project.build.directory}/lib</outputDirectory>
+                            <!--<excludeArtifactIds>javafx-web, javafx-media, javafx-graphics, javafx-controls, javafx-base</excludeArtifactIds> Avoid to copy those libs when system have native openjfx-11-->
```
plata: What about this commented part?

phoenicis-dist/src/deb/control/control
```diff
@@ -1,13 +1,13 @@
...
-              on Linux. It allows you to manage differents virtual hard drive,
+              It permits you to install Windows Games and software
+              on Linux. It allows you to manage different virtual hard drive,
               and several wine versions.
              Copyright 2011 PlayOnLinux team <contact@playonlinux.com>
 Maintainer: PlayOnLinux Packaging <packages@playonlinux.com>
 Homepage: www.phoenicis.org
-Depends: default-jre | java8-runtime, openjfx, unzip, wget, xterm | x-terminal-emulator, imagemagick, cabextract, icoutils, p7zip-full, curl, libxext6:i386
\ No newline at end of file
+Depends: default-jre | java8-runtime, openjfx, unzip, wget, xterm | x-terminal-emulator, imagemagick, cabextract, icoutils, p7zip-full, curl, libwine-development, libwine-development:i386
```
plata: Why is libxext6 removed and wine-devel added?

ImperatorS79:
Since .deb dependencies and launch script depends on which OS you are, maybe profile could be uses here. Compiling as follows:
```
mvn -Dubuntu1810 clean package
```
would then build a .deb for ubuntu 18.10. Putting nothing would not build any .deb. Different control file and launch scripts could be then specified for each supported distribution (which could only be latest stable and latest version).

What do you think about this ?

plata:
Like I've pointed out in some other issue already, I would like to reduce our testing effort and deploy via flatpak. If we now add profiles for different OSs, our build matrix will explode.

phoenicis-dist/src/deb/control/control
plata: Yes, listing the required dependencies would be better.

ImperatorS79:
Yes I know, but you cannot continue to provide a .deb and hope it will work in every distribution. So supporting only latest LTS (+ latest one maybe) and saying which one you are supporting in the wiki  seems to be quite reasonable.

plata:
Ok. So let's say we support always the recent Ubuntu LTS (which will automatically include e.g. Mint). In this case we don't need the profiles, right?

ImperatorS79:
Yes indeed

ImperatorS79:
Do not merge this, still some work to do.

ImperatorS79:
This is working for me.

plata:
@ImperatorS79 shall we merge this now?

phoenicis-dist/src/deb/control/control
```diff
@@ -1,13 +1,13 @@
...
-              on Linux. It allows you to manage differents virtual hard drive,
+              It permits you to install Windows Games and software
+              on Linux. It allows you to manage different virtual hard drive,
               and several wine versions.
              Copyright 2011 PlayOnLinux team <contact@playonlinux.com>
 Maintainer: PlayOnLinux Packaging <packages@playonlinux.com>
-Homepage: www.phoenicis.org
-Depends: default-jre | java8-runtime, openjfx, unzip, wget, xterm | x-terminal-emulator, imagemagick, cabextract, icoutils, p7zip-full, curl, libxext6:i386
\ No newline at end of file
+Homepage: http://www.phoenicis.org
```
plata: Should be https.

ImperatorS79:
Ready to be merge.

plata:
@ImperatorS79 this breaks the Travis build in master somehow. No idea way because the pull request build was ok. Can you please check?

ImperatorS79:
The .deb is for ubuntu 18.04, and travis uses ubuntu 16.04 to test .deb
installation. Maybe it comes from that.

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Garanti
sans virus. www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

PR #1455, opened 2018-11-07
Travis: do not use docker to test OpenJDK 8 build

plata:


madoar:
We need the openjfx installation to test OpenJDK 8.
Does travis provide openjfx now on its package repository?

plata:
You are right. I forgot why we're using docker for the OpenJDK 8 build.

PR #1456, opened 2018-11-07
Travis: try specific OpenJDK 8 version

plata:


plata:
This was only to investigate the reason for the OpenJDK 8 build error in Travis.

PR #1457, opened 2018-11-07
Implement workaround for surefire bug 1588

### LINKED ISSUE #1447: OpenJDK 8 build fails in pull requests

plata:
see https://issues.apache.org/jira/browse/SUREFIRE-1588
This should be reverted as soon as the surefire bugfix is available.

fixes #1447

I do not understand why the change in the pom.xml does not work. It is working for me locally. Looking at the logs, it seems that Travis uses surefire 2.17 and not 2.22.1. Not sure if that's ok...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ jobs:
         - docker exec -t ubuntu-16-04 bash -c "cd /travis;
                                                git clone https://github.com/PhoenicisOrg/phoenicis.git;
                                                cd phoenicis;
-                                               mvn -B clean package"
+                                               _JAVA_OPTIONS=-Djdk.net.URLClassPath.disableClassPathURLCheck=true mvn -B clean package"
       before_cache:
         - docker exec -t ubuntu-16-04 bash -c "cp /travis/phoenicis/phoenicis-dist/target/*.deb /travis/travis_cache/"
     # build + test
--- a/pom.xml
+++ b/pom.xml
@@ -124,6 +124,14 @@
                         </lifecycleMappingMetadata>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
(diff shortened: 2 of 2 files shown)


madoar:
It looks to me like this PR only fixes the build bug on Travis and not for everyone who uses Java 8 to build it locally on their computer.

I think a better short time solution would be to use a maven profile, which has been suggested [here](https://issues.apache.org/jira/browse/SUREFIRE-1588?focusedCommentId=16671777&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16671777).
I think, this should solve the problem for both Travis and local builds.

plata:
That's not true. I had the issue locally as well and my change in the pom fixed it. See also: https://issues.apache.org/jira/browse/SUREFIRE-1588?focusedCommentId=16670105&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16670105

madoar:
The profile solution was meant as a way to remove the need to modify the `docker` command.

The `maven-surefire-plugin` version difference between a travis build and a local build is really strange.
I think we should send a message to the people from travis and ask them if this is a bug.

plata:
It doesn't work. The profile is not activated by default. This is stated in the Maven documentation:
> This profile will automatically be active for all builds unless another profile in the same POM is activated using one of the previously described methods. All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config. 
So our JDK profiles deactivate the `activeByDefault`. I've also tried to activate it manually (which really isn't an option) but in this case, the tests failed. I assume that this is because `<argLine>` conflicts with the `<argLine>` config for the JavaFX tests.

In any case this is only a temporary solution until the Maven plugin is updated with the fix so can we please merge this such that the other pull requests are not blocked anymore?

ovari:
@madoar can you please accept this PR so that other PRs can be merged? Thank you

plata:
@madoar I will merge this now.

PR #1464, opened 2018-11-14
Revert "Implement workaround for surefire bug 1588"

plata:
Reverts PhoenicisOrg/phoenicis#1457 (temporary workaround)


plata:
Merge this when the official surefire bugfix is available in maven (dependabot will tell us).

PR #1471, opened 2018-11-18
Update maven-surefire-plugin to 3.0.0-M1

plata:


plata:
@dependabot[bot] why didn't you do this?

greysteil:
Apologies for not answering this here sooner! It must have got lost in my inbox.

Dependabot didn't create this update because it thinks that milestone releases are pre-releases (I think correctly?). As a result, it won't update you from the stable version to a pre. It *would* have created an update once 3.0.0 was released, and now that you've updated to a milestone release you'll get PRs for new milestone releases of this dependency.

plata:
I see. I don't know exactly how surefire handles their updates but your explanation perfectly makes sense to me and I would agree that the dependabot behavior is correct.

PR #1481, opened 2018-11-24
Refactor the SearchBox UI component with MVC in mind

madoar:
This PR is meant as the first part of a large GUI refactoring.
The problem with the current GUI implementation is, that most components are implemented using a single class, in which the model, view and controller code has been implemented.

A more modern and in my opinion much better readable and maintainable approach is the MVC (model, view, controller) pattern, in which the different competences of a visual component are split into multiple classes.
I've tried to implement this in this PR for the `SearchBox` component.

Background:
- http://www.guigarage.com/2012/11/custom-ui-controls-with-javafx-part-1/
- https://github.com/dlemmermann/CalendarFX
- https://github.com/javafxports/openjdk-jfx/issues/241


plata:
@qparis I would prefer if you could check this (i.e. does it fit in your general concept how the JavaFX GUI should be implemented).

madoar:
From my side this is now ready to be merged.
Do you have any improvement ideas or other suggestions?

plata:
Like I said, I would prefer if you could review this @qparis.

qparis:
looks clean for me

PR #1482, opened 2018-11-25
OSX build goals

qparis:
This merge request add a goal to build a .dmg for Mac OS X

* For the moment, this require Java 1.8. 
* It uses mavan-javafx which is deprecated. I do not have a better solution so far


madoar:
@qparis an alternative at least for the JavaFX application could be [javapackager](https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javapackager.html), but I'm not sure how easy it is to use and configure.

qparis:
Basically, this is what is used behind javafx goal. 
By the way, I cannot run POM 5 with Java >= 9.

madoar:
Ah ok, the used maven plugin calls javapackager underneath.
To solve the issue with Java 9+, I think we need to use javapackager manually.
I believe it's likely that we require different commands depending on the used Java version, i.e. one for Java 8 and one for Java 9+.

qparis:
>  I believe it's likely that we require different commands depending on the used Java version, i.e. one for Java 8 and one for Java 9+.

I cannot run Phoenicis on Mac with Java 9+ at all (even on my IDE). I get Error: JavaFX runtime components are missing, and are required to run this application

qparis:
That makes sense.

phoenicis-cli/pom.xml
```diff
@@ -21,11 +21,11 @@
     <parent>
         <groupId>org.phoenicis</groupId>
         <artifactId>phoenicis</artifactId>
-        <version>${phoenicis.version}</version>
+        <version>5.0-SNAPSHOT</version>
```
madoar: Is this required to make the OSX build work?
If not I think it would be better to move these unrelated pom changes to its own PR.

qparis: Yes it is. Without this, the module maven-javafx will not work on a submodule

madoar: So basically the plugin doesn't work with maven parameters (in our case the version of the module)?

qparis: It is normally not allowed by maven to use parameters in the <parent> section

plata: Yes, IntelliJ complains about this all the time as well. If you think about it, it really doesn't make sense to do this either. You could very well update one module while leaving the other one untouched. Maybe we could use one of the solutions provided [here](https://stackoverflow.com/questions/17312681/maven-multi-module-project-version-management)?

README.md
```diff
@@ -26,6 +27,17 @@ unzip phoenicis-dist.zip
 ./phoenicis-dist/phoenicis.sh
 ```
 
+### OSX
+Build
+```
+mvn clean package install
```
plata: Why do `install` here if it's not used to run Phoenicis (see below)?

phoenicis-blue.svg
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
```
plata: This should be in https://github.com/PhoenicisOrg/resources

phoenicis-javafx/pom.xml
```diff
@@ -143,6 +143,16 @@
                     <mainClass>org.phoenicis.javafx.JavaFXApplication</mainClass>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>com.zenjava</groupId>
```
plata: Don't you want to package this like here: https://github.com/PhoenicisOrg/phoenicis/pull/1389/files ?

phoenicis-javafx/src/main/deploy/package/macosx/phoenicis-blue.svg
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
```
plata: For me it's a bit strange that we have this deploy now in phoenicis-javafx because all other deployment is done in phoenicis-dist. I think we should be consistent.

pom.xml
```diff
@@ -43,8 +43,9 @@
         <connection>scm:git:https://github.com/PhoenicisOrg/phoenicis.git</connection>
     </scm>
     <properties>
-        <phoenicis.version>5.0-SNAPSHOT</phoenicis.version>
-        <project.settings>${project.basedir}/settings</project.settings>
+        <phoenicis.version>${project.version}</phoenicis.version>
```
plata: Does it make sense to have `phoenicis.version`?

README.md
qparis: Because the plugin is fetching dependencies from the maven repository

phoenicis-javafx/src/main/deploy/package/macosx/phoenicis-blue.svg
qparis: Agreed but the plugin is not compatible with that :(

pom.xml
qparis: It is just for backward compatiblity. We shall remove it

phoenicis-javafx/pom.xml
qparis: This is basically what is done here, except that I am not using.a snapshot version of the plugin (the downside is that it won't work with Java 9+)

phoenicis-blue.svg
qparis: I will add it.

README.md
plata: I don't quite understand. The required dependencies should be downloaded during build anyways.

pom.xml
plata: Is it used anywhere? Can you either open an issue to remove it or remove it in this pr if it's unused?

phoenicis-blue.svg
plata: Or move? We shouldn't have it in 3 different places.

phoenicis-javafx/src/main/deploy/package/macosx/phoenicis-blue.svg
plata: I see. Nevertheless: Are the files in this directory even used anywhere?

pom.xml
qparis: I think it is used for cross dependencies. I'll open a issue

phoenicis-blue.svg
qparis: I think that it makes sense to have it somewhere in a submodule, so that it could be inserted anywhere (Readme, about window, etc..)

phoenicis-javafx/pom.xml
qparis: Yes

phoenicis-javafx/src/main/deploy/package/macosx/phoenicis-blue.svg
qparis: Yes, they are read by the plugin to generate the .app icon

README.md
qparis: I mean phoenicis-* dependencies

phoenicis-blue.svg
plata: What about phoenicis-javafx/src/main/resources/org/phoenicis/javafx/views/common/? At least we have the `phoenicis.png` there already so it would fit somehow.

phoenicis-javafx/src/main/deploy/package/macosx/phoenicis-blue.svg
plata: So these files are used by default? Just wondering because I can't find them anywhere in a pom.

phoenicis-javafx/pom.xml
plata: Ok.

README.md
plata: I thought the purpose was to get a fat .jar with all dependencies?

phoenicis-javafx/src/main/deploy/package/macosx/phoenicis-blue.svg
qparis: Yes

phoenicis-blue.svg
qparis: Good to me

README.md
qparis: a .dmg to be more precise. However the dependencies are not grabbed from the current source tree but from your local maven repository

phoenicis-javafx/src/main/deploy/package/macosx/phoenicis-blue.svg
plata: Fine.

README.md
plata: I see. So we cannot include the .dmg build in the normal build process because phoenicis must be installed first? I somehow assumed that the plugin could pick up the project internal dependencies without having to install them.

qparis: We cannot, not because of mvn install (I think that most of the CI platform will not have problem with that). However, OSX tools are needed to create the .dmg, so we must have a Apple machine so far to do it

plata: You can select `os: osx` in Travis. Don't you think that is sufficient?

phoenicis-cli/pom.xml
plata: Sure. Can you do that?

madoar:
Now that we have updated to use Java 9 modules why don't we use the javapackager directly?
I mean, in addition to working with all Java versions it also allows us to build `deb` and `rpm` installers.
The main issue would be to configure it correctly.

README.md
qparis: I don't think so because I also need to provide my personal certificate to sign the package...

qparis: I know, this is really a pain

plata:
I wouldn't like anything manual. The build should be completely automatic with maven.

madoar:
It can be automated with maven, it would just be the case that maven executes the script for us, see https://stackoverflow.com/questions/29273279/maven-build-native-installer-of-javafx-application-with-inno-setup-script

README.md
plata: Oh.

qparis:
> I wouldn't like anything manual. The build should be completely automatic with maven.

We sure have to do it on the long run. However, if we want to handle this cleanly, the best approach is to ... write a maven plugin. So we have to decide if we want to start from scratch or if we want to fork the project.

In the mean time, we need this temporary solution so that we can release the next alpha version for OSX users

plata:
I don't think it's a good idea to maintain our own maven plugin... However, that's a completely different topic now. If you need your credentials to build the .dmg, this will never work completely automatic anyways (and Mac OS is our only use case for this plugin).

qparis:
> I don't think it's a good idea to maintain our own maven plugin... However, that's a completely different topic now. If you need your credentials to build the .dmg, this will never work completely automatic anyways (and Mac OS is our only use case for this plugin).

We don't need my credentials, we need some someone to have some credentials. So yes, you can consider that he build is automated 👍 We are not going to rely on travis to make releases though

qparis:
Also, the compilation will work without the certificate. Gatekeeper will just complain that's the point

madoar:
@qparis off-topic: can you please give #1481 a short look over?

plata:
Ok. Then I don't see a problem with the current implementation (except that it only works for Java 8 which might be fixed in future versions of the plugin).

plata:
@qparis Can you add a .dmg build (and test if possible) to the Travis "Packages" stage?

qparis:
> @qparis Can you add a .dmg build (and test if possible) to the Travis "Packages" stage?

Yes! Can I do it on another ticket?

plata:
Sure.

PR #1484, opened 2018-11-26
Provide phoenicis.sh for Java 8 and Java 10+

### LINKED ISSUE #1408: phoenicis.sh must be different for Java 8 and 10/11
For Java 10/11, run with:
```
java -cp 'phoenicis-dist/target/lib/*':phoenicis-dist/target/phoenicis-dist-5.0-SNAPSHOT.jar -p /<path>/javafx-sdk-11/lib --add-modules javafx.controls org.phoenicis.javafx.JavaFXApplication
```

How can we handle this?

plata:
fixes #1408

### DIFF
--- a/phoenicis-dist/src/scripts/phoenicis-java8.sh
+++ b/phoenicis-dist/src/scripts/phoenicis-java8.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+POL_HOME="$(dirname "$0")"
+cd "$POL_HOME"
+POL_HOME="$PWD"
+CLASSPATH=${CLASSPATH}:$POL_HOME/lib/*
+
+java -classpath "$CLASSPATH" org.phoenicis.javafx.JavaFXApplication "$@"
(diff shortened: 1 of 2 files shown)


madoar:
Is it possible to add a java version check to the shellscript to prevent the need of two different launcher files?

plata:
I wanted to do that originally but the file is not copied in the pom but from the assembly plugin. Check #1485.

madoar:
Ah, I forgot that.
Just to recap: It is currently the case, that the required end-user java version depends on the java version used to build the jar?

plata:
Yes, but that's normal. You cannot run anything built for Java 10 with Java 8 (the other way around should work). In our case we have the additional issue with JavaFX.

madoar:
This adds the question: which JDK do we use to build our release version?

plata:
The default for the target system. So Java 8 for Ubuntu 16.04, Java 10 for Ubuntu 18.04, Java 11 for flatpak etc.

plata:
@madoar is there anything left that should be done for this PR?

plata:
I fully agree with that.

PR #1488, opened 2018-11-26
Move style definitions from the Java code to CSS

madoar:


madoar:
Problematic components:
- all `GridPane`s using css class `.grid`: 
   Some specify their own hgap and vgap (e.g. `AboutPanel`).
   Others define custom column constraints, which don't seem to be changable/settable by css (e.g. `EditShortcutPanel`).
   Maybe we should split the css class into two classes based on the use-case?

plata:
Yes, I think it would make sense to do that.

madoar:
@plata can you please approve this PR, I need it for #1489 :)

plata:
Ok. It's impossible to review this in code anyways. I guess you checked that it still looks ok, so it's fine for me.

madoar:
Yes, I've tested the application to see if its visual impression has changed, which didn't happen.
Sorry, I should have mentioned that earlier.

PR #1489, opened 2018-11-27
Refactor the list widget selector with MVC in mind

madoar:
This PR is still work in progress

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
@@ -1,0 +1,71 @@
+package org.phoenicis.javafx.components.behavior;
+
+import javafx.event.ActionEvent;
+import javafx.scene.control.ToggleButton;
+import org.phoenicis.javafx.components.control.ListWidgetSelector;
+import org.phoenicis.javafx.components.skin.ListWidgetSelectorSkin;
+import org.phoenicis.javafx.views.common.widgets.lists.ListWidgetType;
+
+import java.util.Optional;
+import java.util.function.Consumer;
+
+public class ListWidgetSelectorBehavior
+        extends BehaviorBase<ListWidgetSelector, ListWidgetSelectorSkin, ListWidgetSelectorBehavior> {
+    /**
+     * Constructor
+     *
+     * @param control The control associated with this behavior
+     * @param skin The skin associated with this behavior
+     */
+    public ListWidgetSelectorBehavior(ListWidgetSelector control, ListWidgetSelectorSkin skin) {
+        super(control, skin);
+    }
+
+    @Override
+    public void initialise() {
+        getSkin().getIconsListButton().addEventFilter(ActionEvent.ANY, this::eventFilter);
+        getSkin().getIconsListButton()
+                .setOnAction(event -> {
+                    getControl().setSelected(ListWidgetType.ICONS_LIST);
+                    getOnSelect().ifPresent(consumer -> consumer.accept(ListWidgetType.ICONS_LIST));
+                });
+
+        getSkin().getCompactListButton().addEventFilter(ActionEvent.ANY, this::eventFilter);
+        getSkin().getCompactListButton()
+                .setOnAction(event -> {
+                    getControl().setSelected(ListWidgetType.COMPACT_LIST);
+                    getOnSelect().ifPresent(consumer -> consumer.accept(ListWidgetType.COMPACT_LIST));
+                });
+
+        getSkin().getDetailsListButton().addEventFilter(ActionEvent.ANY, this::eventFilter);
+        getSkin().getDetailsListButton()
+                .setOnAction(event -> {
+                    getControl().setSelected(ListWidgetType.DETAILS_LIST);
+                    getOnSelect().ifPresent(consumer -> consumer.accept(ListWidgetType.DETAILS_LIST));
+                });
+
+        // ensure that if at startup a selected type is set select it
+        Optional.ofNullable(getControl().getSelected())
+                .ifPresent(selected -> getSkin().getListButton(selected).fire());
+
+        // change the selection based on a change in the selected property
+        getControl().selectedProperty()
+                .addListener(invalidatedEvent -> getSkin().getListButton(getControl().getSelected()).fire());
+    }
+
+    /**
+     * An event filter to prevent the unselectation of all buttons
+     *
+     * @param event The input event to be filtered
+     */
+    private void eventFilter(ActionEvent event) {
+        ToggleButton source = (ToggleButton) event.getSource();
+        if (source.getToggleGroup() == null || !source.isSelected()) {
+            source.fire();
+        }
+    }
+
+    private Optional<Consumer<ListWidgetType>> getOnSelect() {
+        return Optional.ofNullable(getControl().getOnSelect());
+    }
+}
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/skin/ListWidgetSelectorSkin.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/skin/ListWidgetSelectorSkin.java
@@ -1,0 +1,80 @@
+package org.phoenicis.javafx.components.skin;
+
+import javafx.geometry.Pos;
+import javafx.scene.control.ToggleButton;
+import javafx.scene.control.ToggleGroup;
+import javafx.scene.layout.HBox;
+import org.phoenicis.javafx.components.behavior.ListWidgetSelectorBehavior;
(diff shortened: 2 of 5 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
```diff
@@ -0,0 +1,85 @@
...
+import org.phoenicis.javafx.components.skin.ListWidgetSelectorSkin;
+import org.phoenicis.javafx.views.common.widgets.lists.ListWidgetType;
+
+import java.util.Optional;
+import java.util.function.Consumer;
+
+/**
+ * The behavior for the {@link ListWidgetSelector} component
+ */
+public class ListWidgetSelectorBehavior
```
plata: This is for our miniature list in apps etc., right? I'm asking because the name seems to be a bit generic to me.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
```diff
@@ -0,0 +1,85 @@
...
+
+        // set the behavior for the details list toggle button
+        getSkin().getDetailsListButton().addEventFilter(ActionEvent.ANY, this::eventFilter);
+        getSkin().getDetailsListButton()
+                .setOnAction(event -> {
+                    getControl().setSelected(ListWidgetType.DETAILS_LIST);
+                    getOnSelect().ifPresent(consumer -> consumer.accept(ListWidgetType.DETAILS_LIST));
+                });
+
+        // ensure that if at startup a selected type is set select it
```
plata: I don't get this comment

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
```diff
@@ -0,0 +1,85 @@
...
+        Optional.ofNullable(getControl().getSelected())
+                .ifPresent(selected -> getSkin().getListButton(selected).fire());
+
+        // change the selection based on a change in the selected property
+        getControl().selectedProperty()
+                .addListener(invalidatedEvent -> getSkin().getListButton(getControl().getSelected()).fire());
+    }
+
+    /**
+     * An event filter to prevent the unselectation of all buttons
```
plata: selection?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/skin/ListWidgetSelectorSkin.java
```diff
@@ -0,0 +1,104 @@
...
+import javafx.scene.control.ToggleGroup;
+import javafx.scene.layout.HBox;
+import org.phoenicis.javafx.components.behavior.ListWidgetSelectorBehavior;
+import org.phoenicis.javafx.components.control.ListWidgetSelector;
+import org.phoenicis.javafx.views.common.widgets.lists.ListWidgetType;
+
+/**
+ * The skin for the {@link ListWidgetSelector} component
+ */
+public class ListWidgetSelectorSkin
```
plata: Naming: Is it really a "skin" or rather a "ui"? For me, skins just look different while this can be implemented with completely different controls (if I understand it correctly).

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
madoar: No, it is the component used to select the list widget, that shows the apps etc.
grafik

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
madoar: The following command performs the initial list widget selection based on your settings

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
madoar: deselection seems to be the better English word instead of unselection.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/skin/ListWidgetSelectorSkin.java
madoar: As I understand it, "skin" is the JavaFX name for "ui".
For more details take a look at the [Skin](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Skin.html) class, which is the upper most parent class of all our `Skin` classes.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
plata: Ok, but my point remains: "ListWidget" is somewhat generic given that it refers to this special thing. I must admit that I don't have an idea for a better name currently.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
plata: Can you change it in the code?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
plata: Was just about the "selectation" part because that just doesn't seem to exist.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/skin/ListWidgetSelectorSkin.java
plata: I wasn't aware of this but if it's the normal JavaFX naming, it's fine for me.

plata:
Maybe better in GitHub pages. I wouldn't know where to put it in JavaDoc.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
madoar: I don't understand what exactly I should change?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
madoar: Hmm.
Every list widget we have implemented, i.e. ``, `` and ``, all implement the `ListWidget` interface.
So I think the name `ListWidgetSelector` is quite fitting.
What I can change is, that I can add a short reference to `ListWidget` to the `ListWidgetSelector` interface.
What do you think?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
plata: The comment. For me, it's not clear. What you wrote here is much more understandable.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/behavior/ListWidgetSelectorBehavior.java
plata: That's exactly the point. Every list widget **we** have implemented. With the name "ListWidget" it sounds like a very generic list widget (which it is not). In my opinion, the name should make clear that this is about the special list widget in Phoenicis which can display items in 3 different styles.

madoar: We could extend the component in the future to be able to show an arbitrary number of list widgets, which are somehow passed to the component (most likely via an `ObservableList`).
Then it would be more generic.

plata: Yes, I know. Like I said: I have no better name currently. So let's keep it.

madoar:
I created a new issue #1503 to track the progress on our JavaFX MVC documentation.
I think we should add the JavaFX MVC documentation in its own PR.

PR #1504, opened 2018-11-30
Add splash screen

plata:
splash

based on: https://gist.github.com/jewelsea/2305098

This doesn't quite achieve what I want yet. My idea was to preload the repository etc. The main issue is that there are many dependencies to JavaFX parts and those cannot be loaded if we are not on the GUI thread.
Ideas welcome.

Things to preload:
- repository (required for translations and default icons)
- library
- themes


madoar:
For the GUI thread problem you can use `Platform#runLater()`.
Instead of a `Task` + `Thread` combination I would recommend using `Executors.newSingleThreadExecutor()#submit(Runnable)`.
The result is a `Future` object, which you can then query if it is done.

plata:
`Platform#runLater` will not work. I want to run the commands explicitly during the load phase. It doesn't help if this is delayed.
What's the advantage of thread executor? How must `showSplash` be changed if I do not have a `Task`?

plata:
I think this is ok as an initial version. Currently, it loads only the repositories. Maybe we can extend it later on to preload also the library and engines.

qparis:
Does this mean that the interface cannot be used until the repository is loaded?

plata:
Yes. I know that you'll probably argue that we don't want that. However, what I see currently (especially on slow machines) is that Phoenicis starts and then stays unusable for quite some time. You could even get the impression that it hung up completely.

qparis:
I think that the "right" behavior is to at least allow to run a local shortcut while the list of installer is being updated. I would try to copy the same behavior than POLv4

plata:
I agree. But that doesn't work currently. Apart from that even loading the library takes quite some time.

qparis:
I think we should focus on this. It is very important to understand that a large part of our users do not really care about the scripts. They just need a good wine manager (and this explains why POLv4 is still used today)

plata:
Ok. I can remove the synchronous repository load again. Nevertheless, I think loading some stuff before we start makes sense (e.g. the library). I couldn't manage to load the library while not on the JavaFX thread because there are some dependencies through the Beans. Could you have a look at that?
Apart from that I would suggest to lazy load the repository and the engines only when they are needed.

madoar:
Ignore my previous comment about `Executors.newSingleThreadExecutor()#submit(Runnable)`, I didn't see that `Task` is a JavaFX class, which provides useful functionality that we need here.

About the library:
I agree with you, it would be nice if the library, which is displayed directly after startup, doesn't show up empty, because the shortcuts are fetched lazy.
Do you know where the dependencies lead to?

plata:
I think the dependencies come from the wizard etc. (i.e. JavaFX classes you can use from the scripts). I will check it in more detail.

plata:
I've pushed the library load code. It crashes without error message when trying to create the application context...

plata:
I do not have a solution for preloading stuff on the JavFX thread (themes, library). However, with the lazy loading of apps, container and engines and the preloading of the repository, I have the feeling that Phoenicis is now usable as soon as the main window pops up.
Please give it a try.

I'm not really happy with the design yet and it would be nice to have more feedback during the repository load but I think this can be improved later.

madoar:
I like the look of the new "loading screen" (i.e splash screen).
My only issue is, that we afterwards still need to wait for the applications and engines to be loaded.
I think this is intentional but still strange in my opinion.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
```diff
@@ -19,15 +19,43 @@
...
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
 
+import java.util.Arrays;
+import java.util.List;
+
+import static org.phoenicis.configuration.localisation.Localisation.tr;
+
 public class JavaFXApplication extends Application {
+    private final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(JavaFXApplication.class);
```
madoar: Please import the `Logger` class in the header.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
```diff
@@ -38,10 +66,106 @@ public static void main(String[] args) {
...
+        splashHeight = splashWidth * (1 / splashImageAspectRatio);
+
+        final ImageView splash = new ImageView(splashImage);
+        splash.setFitWidth(splashWidth);
+        splash.setFitHeight(splashHeight);
+        loadProgress = new ProgressBar();
+        loadProgress.setPrefWidth(splashWidth);
+        progressText = new Label(tr("Loading ..."));
+        progressText.setFont(new Font(12));
+        splashLayout = new VBox();
```
madoar: This and the next line can be unified via the `new VBox(nodes)` constructor.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
```diff
@@ -19,15 +19,43 @@
...
+import java.util.List;
+
+import static org.phoenicis.configuration.localisation.Localisation.tr;
+
 public class JavaFXApplication extends Application {
+    private final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(JavaFXApplication.class);
+
+    private double splashWidth;
+    private double splashHeight;
+    private Pane splashLayout;
```
madoar: Are all these properties required on an object scope?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
```diff
@@ -38,10 +66,106 @@ public static void main(String[] args) {
...
+
+                return null;
+            }
+        };
+
+        showSplash(initStage, loadTask, this::showMainStage);
+        new Thread(loadTask).start();
+    }
+
+    private void showSplash(final Stage initStage, Task<?> task, InitCompletionHandler initCompletionHandler) {
```
madoar: I would like some JavaDoc in this method to understand what happens inside

phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
```diff
@@ -51,13 +175,11 @@ public void start(Stage primaryStage) {
...
         });
+        mainStage.toFront();
     }
 
-    private void loadFonts() {
-        Font.loadFont(getClass().getResource("views/common/mavenpro/MavenPro-Medium.ttf").toExternalForm(), 12);
-        Font.loadFont(getClass().getResource("views/common/roboto/Roboto-Medium.ttf").toExternalForm(), 12);
-        Font.loadFont(getClass().getResource("views/common/roboto/Roboto-Light.ttf").toExternalForm(), 12);
-        Font.loadFont(getClass().getResource("views/common/roboto/Roboto-Bold.ttf").toExternalForm(), 12);
+    public interface InitCompletionHandler {
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/LoadingException.java
```diff
@@ -0,0 +1,29 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.javafx;
+
+public class LoadingException extends RuntimeException {
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
plata: They are used in `init` and `showSplash`.

qparis:
I don't understand what you decided to do eventually with the splash screen. Be careful, I suspect that people will fell that that our technical stack is "bloated" if we add a splashscreen compared to POLv4

qparis:
> We could of course use some fallback icon and English strings but I don't think that's an approach we should consider at all. For example, it would look completely unprofessional if parts of your GUI suddenly jump from English to French.

This is what I was thiking about. Because 90% of the interface (Categories name, tabs, ...) will exist once for all, right?

plata:
I don't quite understand the question. Yes, many parts of the GUI are defined purely in the Java software and do not require the scripts.

qparis:
> I don't quite understand the question. Yes, many parts of the GUI are defined purely in the Java software and do not require the scripts.

So we can provide translation for this main part, show in english not found translation while POL is being loaded. The user will be served sooner and will not need to have a working internet connection to launch its software

plata:
Just to clarify: you do not need a working Internet connection right now either. If you have a git repository defined but no Internet available, it will simply use the cache.

Like I said: If I open any application and it's loaded, I expect that it is fully functional. Therefore, I personally would find it very strange if some icons/strings change while I'm already doing stuff.

madoar:
Can't we just try out how the splash screen is perceived by the users?
I mean we're currently in alpha, so I as a user would expect that the developers are still trying stuff out to see whether a feature is beneficial or if it needs to be replaced with something else.

qparis:
> Like I said: If I open any application and it's loaded, I expect that it is fully functional. Therefore, I personally would find it very strange if some icons/strings change while I'm already doing stuff.

Really depends on what changes. If it is core stuff, I agree with you. If we are talking about something like legal mentions, we honestly don't care. I really believe that fonctionnally speaking, we can bundle pretty much what is really needed. 

Look at how Google define website performance here: https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/metrics/speed-index. We are not creating a webpage, but I really think that 
we should take inspiration on the concept

qparis:
> Can't we just try out how the splash screen is perceived by the users?
> I mean we're currently in alpha, so I as a user would expect that the developers are still trying stuff out to see whether a feature is beneficial or if it needs to be replaced with something else.

The problem is that because of the choice of Java, some people are waiting for the chance to say that the new version is overbloated. 
Even if this has nothing to do with Java, I just think that we should be irreproachable on that point because we really don't want to deal with new polemics.

plata:
Regarding the website rules: Generally, yes. But I think we all know these pages where you click somewhere and then end up in a completely different place because the website decided to put a different button under your cursor... We should avoid similar "responsiveness".

Your point about Java etc. is completely valid. I'm just not convinced that changing icons/strings later on will not cause the same polemics.

qparis:
Is the splash screen displayed at first load only?
Maybe we can try to look at the interface and imagine what will change later on to see what is the impact?

plata:
No, not only on first load. The problem is that loading the repository (even from your local disk) is slower than you would think (if your computer is not that powerful/you have no SSD). The main reason for this is that we cannot load the repository partly, i.e. all applications must be parsed already, and the `ObjectMapper` takes some time.

Do you have anything special in mind that will change later?

plata:
Technically, we only need the translations and the category icons. We cannot put this in a separate repository because we cannot know which repositories are used by the users.

What we would need is something like "shallow" repository load. On the other hand, I'm not sure it's really worth it because that will probably be error prone and induce crazy issues where you would never expect it.
Can you try this branch on your machine to get a feeling how long it takes until the splash finishes on a reasonably fast computer?

qparis:
I have a really fast machine so I guess it will not be a problem on my computer. I think we need to see how long it is on an average disk

madoar:
I also have a reasonable fast machine with an i7 CPU, DDR4 RAM and a SSD, but the startup still feels "long" (1 - 3 seconds? I haven't measured it) in my opinion.
The progress bar on the splash screen stays at the halfway point for a while.

Additionally you have the waiting times later on, when for example the applications tab lazy loads the applications.

plata:
It stays at 50% because we do not have progress feedback during repository load.

madoar:
What could maybe work is to parallelize some stuff during the repository loading, but I'm not sure what is parallelizable during this phase.

plata:
I don't know if we could parallelize the `LocalRepository` load. It might work to load the applications in parallel. However, if the bottleneck is the disk, this will not help (I think it's not).

qparis:
Our problem is that we are treating the same way critical (translations, icons) and non critical (remote scripts) resources the same way. If you think about it, it takes much more time to load everything on POLv4. The differences is that the translations are already there, so you don't really care. 

I think our only solution is to separate thing by level of criticity

plata:
I don't see how this could work. In our design, we have the flexibility that everybody can create her own repositories with own categories etc. So how can you decide what is critical and what not?

qparis:
> I don't see how this could work. In our design, we have the flexibility that everybody can create her own repositories with own categories etc. So how can you decide what is critical and what not?

I guess critical things include by order: 
* Translation of the static interface 
* Things needed to run programs
* Container management 

This just be available locally and very quickly when POL starts

qparis:
Also, I understand that people can create their own categories, but if we can provide translation for the most common ones (i.e. the ones we support) directly in phoenicis, this will be good enough

plata:
With this definition, we already have the critical things available. It still doesn't solve that icons/translations would change after loading.
Thinking about it: the main reason for all this is that the library is not completely local. So maybe the category icons should be part of the library?

qparis:
What do you mean by completely local?

plata:
The library depends on the repository (e.g. for category icons). We could argue that the library should be completely detached from the repository and contain everything it requires.
So instead of having a flat shortcut directory like now, we could do this:
- shortcuts
    - games
        - icon.png
        - notepadplusplus
            - Notepad++.info
            - Notepad++.shortcut

qparis:
So you would decorelate library icons from repository icons?

plata:
Yes.

qparis:
And the icon would be downloaded locally during the installation phase, correct?

plata:
Yes (if it doesn't exist already).

The library icon hierarchy would be:
- standard theme
- shortcut directory
- selected theme (if theme specifies icon for given category)

qparis:
This is very smart. I like the idea

plata:
I will open a new issue for this.

In any case, this doesn't really solve the splash/no splash issue: If your library is very big, it will also take some time to read it. It is impossible to avoid that and we cannot mitigate because without the library, Phoenicis is useless.

ImperatorS79:
IMHO the fact that the loading bar stays at 50% is really problematic: one could thing something is not optimized.

plata:
@qparis no idea. I do not know how it works in PoL4.

@ImperatorS79 I'm not sure how to get more information about the repository loading currently. Shall I simply use a spinning wheel instead?

ImperatorS79:
Yes you should then.

plata:
@ImperatorS79 Updated. Give it a try now.

ImperatorS79:
@plata could you merge master here so that I could build a working .deb ?

ImperatorS79:
The design is really better like this. However I get:
```
java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1542)
	at org.phoenicis.repository.DefaultRepositoryManager.lambda$triggerCallbacks$1(DefaultRepositoryManager.java:131)
	at org.phoenicis.repository.types.Repository.fetchInstallableApplications(Repository.java:49)
	at org.phoenicis.repository.types.BackgroundRepository.lambda$fetchInstallableApplications$0(BackgroundRepository.java:51)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
```
in the engines tab at first launch.

plata:
Only on this branch?

ImperatorS79:
Ok so I removed my `.Phoenicis` dir and everything works (the engine does not seem to have been pre-loaded), but I have this in the log:
```
[INFO ] org.springframework.context.annotation.ConfigurationClassParser (l.458) - Properties location [file:${application.user.settings}] not resolvable: /home/test/.Phoenicis/config.properties (Aucun fichier ou dossier de ce type)
[INFO ] org.phoenicis.repository.DefaultRepositoryManager (l.85) - Adding repositories: [org.phoenicis.repository.location.GitRepositoryLocation@16716145[https://github.com/PhoenicisOrg/scripts,master], org.phoenicis.repository.location.ClasspathRepositoryLocation@30f54484[/org/phoenicis/repository]] at index 0
[INFO ] org.phoenicis.repository.types.MultipleRepository (l.46) - Fetching applications for: org.phoenicis.repository.types.MultipleRepository@1f14e759[repositories=[git-repository https://github.com/PhoenicisOrg/scripts, local folder: /home/test/.Phoenicis/cache/git-331995911, branch: master, org.phoenicis.repository.types.ClasspathRepository@7a5c05ae[packagePath=/org/phoenicis/repository]]]
[INFO ] org.phoenicis.repository.types.GitRepository (l.76) - Begin fetching process of git-repository https://github.com/PhoenicisOrg/scripts, local folder: /home/test/.Phoenicis/cache/git-331995911, branch: master
[INFO ] org.phoenicis.repository.types.GitRepository (l.82) - Creating lock file for git-repository https://github.com/PhoenicisOrg/scripts, local folder: /home/test/.Phoenicis/cache/git-331995911, branch: master
[INFO ] org.phoenicis.repository.types.GitRepository (l.107) - Creating local folder for git-repository https://github.com/PhoenicisOrg/scripts, local folder: /home/test/.Phoenicis/cache/git-331995911, branch: master
[INFO ] org.phoenicis.repository.types.GitRepository (l.120) - Cloning git-repository https://github.com/PhoenicisOrg/scripts, local folder: /home/test/.Phoenicis/cache/git-331995911, branch: master
[INFO ] org.phoenicis.repository.types.LocalRepository (l.76) - Reading directory : /home/test/.Phoenicis/cache/git-331995911
[INFO ] org.springframework.context.annotation.ConfigurationClassParser (l.458) - Properties location [file:${application.user.settings}] not resolvable: /home/test/.Phoenicis/config.properties (Aucun fichier ou dossier de ce type)
[INFO ] org.springframework.context.annotation.ConfigurationClassParser (l.458) - Properties location [file:${application.user.root}/javafx.properties] not resolvable: /home/test/.Phoenicis/javafx.properties (Aucun fichier ou dossier de ce type)
[INFO ] org.phoenicis.repository.DefaultRepositoryManager (l.85) - Adding repositories: [org.phoenicis.repository.location.GitRepositoryLocation@7f716441[https://github.com/PhoenicisOrg/scripts,master], org.phoenicis.repository.location.ClasspathRepositoryLocation@6fac2291[/org/phoenicis/repository]] at index 0
[INFO ] org.phoenicis.javafx.views.common.ThemeManager (l.74) - Setting theme to 'Standard theme'
[INFO ] org.phoenicis.javafx.views.common.ThemeManager (l.110) - Refreshing currently shown theme
[INFO ] org.phoenicis.javafx.views.common.ThemeManager (l.116) - Loading standard theme at 'jar:file:/usr/share/phoenicis/app/phoenicis-javafx-5.0-SNAPSHOT.jar!/org/phoenicis/javafx/themes/standard/main.css'
[INFO ] org.phoenicis.javafx.views.common.ThemeManager (l.125) - Loading WebView stylesheet at 'jar:file:/usr/share/phoenicis/app/phoenicis-javafx-5.0-SNAPSHOT.jar!/org/phoenicis/javafx/themes/standard/description.css'
[INFO ] org.phoenicis.repository.types.MultipleRepository (l.46) - Fetching applications for: org.phoenicis.repository.types.MultipleRepository@2b580480[repositories=[git-repository https://github.com/PhoenicisOrg/scripts, local folder: /home/test/.Phoenicis/cache/git-331995911, branch: master, org.phoenicis.repository.types.ClasspathRepository@259fef9b[packagePath=/org/phoenicis/repository]]]
[INFO ] org.phoenicis.repository.types.GitRepository (l.76) - Begin fetching process of git-repository https://github.com/PhoenicisOrg/scripts, local folder: /home/test/.Phoenicis/cache/git-331995911, branch: master
[INFO ] org.phoenicis.repository.types.GitRepository (l.145) - Opening git-repository https://github.com/PhoenicisOrg/scripts, local folder: /home/test/.Phoenicis/cache/git-331995911, branch: master
[INFO ] org.phoenicis.repository.types.GitRepository (l.149) - Pulling new commits from git-repository https://github.com/PhoenicisOrg/scripts, local folder: /home/test/.Phoenicis/cache/git-331995911, branch: master
[INFO ] org.phoenicis.repository.types.LocalRepository (l.76) - Reading directory : /home/test/.Phoenicis/cache/git-331995911
[ERROR] org.phoenicis.repository.types.Repository (l.51) - Fetching installable applications failed!
java.util.ConcurrentModificationException: null
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1542)
        at org.phoenicis.repository.DefaultRepositoryManager.lambda$triggerCallbacks$1(DefaultRepositoryManager.java:131)
        at org.phoenicis.repository.types.Repository.fetchInstallableApplications(Repository.java:49)
        at org.phoenicis.repository.types.BackgroundRepository.lambda$fetchInstallableApplications$0(BackgroundRepository.java:51)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
```

plata:
I have to check that.

Kreyren:
Recommeds making the splash screen to be compatible with QT+GTK for further customization on linux.

ImperatorS79:
This project is using javafx, not qt or gtk.

Kreyren:
@ImperatorS79 I see.. Sorry i still have zero knowledge in java :X aldo it would be cool to have some way of customization alike using QT/GTK..

plata:
@ImperatorS79 just tested. Doesn't happen for me.

madoar:
@plata @qparis what do you think about this PR? It is quite old. In addition with the "recent" changes to our JavaFX components some feature panels are loading a lot faster compared to before. Do we still want to add a splash screen?

qparis:
I still think that a splash screen is bad in terms of user experience. We should more focus on bring as soon as possible the interface and use skeleton loaaders if necessary

plata:
I'm fine with closing this. What do you mean by "skeleton loaders"?

qparis:
image

I think we should get inspiration from the web performance metrics: the speed index. The sooner the interface is closer to what the users wants at the end, the happier he is. 

Examples: 
image

image

madoar:
How is this speed index calculated?

Kreyren:
i really hate splash screens.. can you at least make it optional?

qparis:
@madoar : you measure the visual progression of your interface from 0% and 100% through the time, and you take the area that is above the curve. The lower, the better

plata:
Ok, I will close this then. We can discuss this in a separate issue.

PR #1505, opened 2018-12-01
Lock git folder during clone/update

### LINKED ISSUE #1463: Check current solution to avoid parallel git cloning
see #1431

plata:
NOTE: The lock is acquired only for the complete JVM, so it's not a replacement for the `synchronized` block.

fixes #1463

### DIFF
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
@@ -33,6 +33,7 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.URI;
+import java.nio.channels.FileLock;
 
 public class GitRepository implements Repository {
     private final static Logger LOGGER = LoggerFactory.getLogger(GitRepository.class);
@@ -48,7 +49,7 @@ public class GitRepository implements Repository {
     // lock file to avoid concurrent access to the git clone
     private final File lockFile;
 
-    private static Object mutex = 0;
+    private static final Object mutex = new Object();
 
     public GitRepository(URI repositoryUri, String branch, String cacheDirectoryPath,
             LocalRepository.Factory localRepositoryFactory, FileUtilities fileUtilities) {
@@ -60,7 +61,7 @@ public class GitRepository implements Repository {
         this.localRepositoryFactory = localRepositoryFactory;
 
         this.localFolder = createRepositoryLocation(cacheDirectoryPath);
-        this.lockFile = new File(this.localFolder.getAbsolutePath() + "_lock");
+        this.lockFile = new File(this.localFolder.getAbsolutePath() + ".lock");
     }
 
     private File createRepositoryLocation(String cacheDirectoryPath) {
@@ -87,13 +88,10 @@ public class GitRepository implements Repository {
                         throw new RepositoryException("Couldn't create lock file " + this.lockFile.getAbsolutePath());
                     }
                 }
-                FileOutputStream lockFileStream = new FileOutputStream(lockFile, true);
-
-                try {
-                    java.nio.channels.FileLock lock = lockFileStream.getChannel().lock();
-                    cloneOrUpdateWithLock(lock);
-                } finally {
-                    lockFileStream.close();
+                try (FileOutputStream lockFileStream = new FileOutputStream(lockFile, true)) {
+                    try (FileLock lock = lockFileStream.getChannel().lock()) {
+                        cloneOrUpdateWithLock(lock);
+                    }
                 }
             } catch (IOException e) {
                 throw new RepositoryException("An unknown error occurred", e);
@@ -176,7 +174,7 @@ public class GitRepository implements Repository {
 
             LOGGER.info("Deleted " + this);
         } catch (IOException e) {
-            LOGGER.error(String.format("Couldn't delete " + this), e);
+            LOGGER.error("Couldn't delete " + this, e);
         }
     }
 
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
@@ -70,7 +70,7 @@ public class GitRepository implements Repository {
         return new File(cacheDirectoryPath + "/git" + hashcode);
     }
 
-    private void cloneOrUpdate() throws RepositoryException {
+    private void cloneOrUpdateWithLock() throws RepositoryException {
         synchronized (mutex) {
             try {
                 LOGGER.info("Begin fetching process of " + this);
@@ -89,8 +89,8 @@ public class GitRepository implements Repository {
                     }
                 }
                 try (FileOutputStream lockFileStream = new FileOutputStream(lockFile, true)) {
-                    try (FileLock lock = lockFileStream.getChannel().lock()) {
-                        cloneOrUpdateWithLock(lock);
+                    try (FileLock ignored = lockFileStream.getChannel().lock()) {
+                        cloneOrUpdate();
                     }
                 }
             } catch (IOException e) {
@@ -99,7 +99,7 @@ public class GitRepository implements Repository {
         }
     }
 
-    private void cloneOrUpdateWithLock(java.nio.channels.FileLock lock) throws RepositoryException, IOException {
+    private void cloneOrUpdate() throws RepositoryException {
         final boolean folderExists = this.localFolder.exists();
 
         // check that the repository folder exists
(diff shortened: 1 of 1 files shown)


phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -65,72 +69,103 @@ private File createRepositoryLocation(String cacheDirectoryPath) {
...
-                    LOGGER.info("Creating local folder for " + this);
+                // check that the repository lock file exists
+                if (!lockFileExists) {
+                    LOGGER.info("Creating lock file for " + this);
 
-                    if (!this.localFolder.mkdirs()) {
-                        throw new RepositoryException("Couldn't create local folder for " + this);
+                    try {
+                        this.lockFile.getParentFile().mkdirs();
+                        this.lockFile.createNewFile();
```
qparis: This (i.e. persistent lock management) is not the responsibility of GitRepository.
You should create another class

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -65,72 +69,103 @@ private File createRepositoryLocation(String cacheDirectoryPath) {
...
                 LOGGER.info("Begin fetching process of " + this);
 
-                boolean folderExists = this.localFolder.exists();
+                boolean lockFileExists = this.lockFile.exists();
 
-                // check that the repository folder exists
-                if (!folderExists) {
-                    LOGGER.info("Creating local folder for " + this);
+                // check that the repository lock file exists
+                if (!lockFileExists) {
```
qparis: When is the file deleted?

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
plata: How should that class look like? I think it would have to know internal information of `GitRepository`.

qparis: What information would it need?

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
plata: Never. Why should it be deleted?

qparis: So it is not a lock then, is it?

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
plata: E.g. when the lock is required and for which file. I could create a class which locks a resource on request but that would give pretty much the same code as I have now (because that's exactly what the Java API does already).

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
plata: No, the file itself is not the lock. I acquire a lock **for** the file with `java.nio.channels.FileLock` (on OS level).

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
qparis: So basically, you need some kind of Map associating a file path with a semaphore, right?

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
qparis: Ok, it seems that I have missed something

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
plata: I don't think so. But not sure I understood you correctly.

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -45,7 +45,10 @@
     private final LocalRepository.Factory localRepositoryFactory;
 
     private final File localFolder;
-    private static Semaphore mutex = new Semaphore(1);
+    // lock file to avoid concurrent access to the git clone
+    private final File lockFile;
+
+    private static Object mutex = 0;
```
madoar: I think I would prefer:
```java
private static final Object mutex = new Object();
```
In my opinion `0` is normally used as a primitive value which can't be locked.

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -57,6 +60,7 @@ public GitRepository(URI repositoryUri, String branch, String cacheDirectoryPath
         this.localRepositoryFactory = localRepositoryFactory;
 
         this.localFolder = createRepositoryLocation(cacheDirectoryPath);
+        this.lockFile = new File(this.localFolder.getAbsolutePath() + "_lock");
```
madoar: Why don't we use the file ending `.lock` for our lock files?

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -65,72 +69,103 @@ private File createRepositoryLocation(String cacheDirectoryPath) {
...
-                        throw new RepositoryException("Couldn't create local folder for " + this);
+                    try {
+                        this.lockFile.getParentFile().mkdirs();
+                        this.lockFile.createNewFile();
+                    } catch (IOException e) {
+                        throw new RepositoryException("Couldn't create lock file " + this.lockFile.getAbsolutePath());
                     }
                 }
-                Git gitRepository = null;
+                FileOutputStream lockFileStream = new FileOutputStream(lockFile, true);
```
madoar: I would prefer:
```java
try (FileOutputStream lockFileStream = new FileOutputStream(lockFile, true)) {
   try (FileLock lock = lockFileStream.getChannel().lock()) {
      cloneOrUpdateWithLock(lock);
   }
}
```

In addition it could be useful to replace the `FileOutputStream` and `lockFile` with a `RandomAccessFile`, which directly provides a `getChannel` method.

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -65,72 +69,103 @@ private File createRepositoryLocation(String cacheDirectoryPath) {
         return new File(cacheDirectoryPath + "/git" + hashcode);
     }
 
-    @Override
-    public synchronized RepositoryDTO fetchInstallableApplications() {
-        RepositoryDTO result = null;
-        try {
-            mutex.acquire();
+    private void cloneOrUpdate() throws RepositoryException {
+        synchronized (mutex) {
```
madoar: So basically we can only clone **one** git repository at a time, independent on the repository we want to clone, right?

plata: Yes. I don't think this limitation is important for us.

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -65,72 +70,100 @@ private File createRepositoryLocation(String cacheDirectoryPath) {
...
             } finally {
-                mutex.release();
+                // close repository to free resources
+                if (gitRepository != null) {
+                    gitRepository.close();
+                }
             }
-        } catch (InterruptedException e) {
-            throw new RepositoryException("InterruptedException occurred", e);
+        } finally {
```
madoar: The `try { ... } finally { lock.release(); }` shouldn't be required anymore, because the calling method automatically does this in:
```java
                    try (FileLock lock = lockFileStream.getChannel().lock()) {
                        cloneOrUpdateWithLock(lock);
                    }
```

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
madoar: I'm not sure.
I think it would be useful to be able to fetch multiple different repositories at the same time.
This is only possible if we use a different mutex for each repository location.

plata: I don't currently see how this could be implemented. Apart from that, I'm not really sure it's worth the effort because it's nothing that will normally happen/have much of an effect.

madoar: It is ok for now, it is just possible that we need to revise it later on if fetching multiple repositories takes to long.

PR #1506, opened 2018-12-01
Run AppConfigurationInitialisationTest in a temporary application.user.root

### LINKED ISSUE #1402: Folder '...' is no git-repository

plata:
fixes #1402 (test creates an empty git folder)

### DIFF
--- a/phoenicis-javafx/src/test/java/org/phoenicis/javafx/AppConfigurationInitialisationTest.java
+++ b/phoenicis-javafx/src/test/java/org/phoenicis/javafx/AppConfigurationInitialisationTest.java
@@ -1,12 +1,12 @@
-/**
- *
- */
 package org.phoenicis.javafx;
 
 import org.junit.Test;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
 import org.testfx.api.FxToolkit;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.concurrent.TimeoutException;
 
 /**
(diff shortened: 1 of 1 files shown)


phoenicis-javafx/src/test/java/org/phoenicis/javafx/AppConfigurationInitialisationTest.java
```diff
@@ -1,26 +1,37 @@
...
  */
 public class AppConfigurationInitialisationTest {
 
+    /**
+     * run the test in a temporary application root environment
+     */
+    private static void setupApplicationRoot() throws IOException {
+        Path tempDir = Files.createTempDirectory(AppConfigurationInitialisationTest.class.getName());
+        tempDir.toFile().deleteOnExit();
+        System.setProperty("application.user.root", tempDir.toString());
```
madoar: This sounds like something we want to do in a `@BeforeClass` annotated method.

phoenicis-javafx/src/test/java/org/phoenicis/javafx/AppConfigurationInitialisationTest.java
```diff
@@ -29,10 +40,17 @@ public static void setupHeadless() {
...
-        setupHeadless();
-        FxToolkit.registerPrimaryStage();
-        FxToolkit.setupFixture(() -> new AnnotationConfigApplicationContext(AppConfiguration.class));
+    public void testAppConfigurationInitialisation() {
+        try {
+            setupApplicationRoot();
+            setupHeadless();
+            FxToolkit.registerPrimaryStage();
+            FxToolkit.setupFixture(() -> new AnnotationConfigApplicationContext(AppConfiguration.class));
+        } catch (TimeoutException | IOException e) {
```
madoar: Why don't we continue to forward the exceptions to junit?

plata: Codacy was complaining that the test didn't contain `fail` or `assert`.

phoenicis-javafx/src/test/java/org/phoenicis/javafx/AppConfigurationInitialisationTest.java
plata: Why should it be different than the setup for headless?

madoar: It shouldn't.
I think we should move the headless initialization to a method annotated with `@BeforeClass` too.
I don't know why I didn't do that initially...

plata: Maybe because I wrote that method ;) (you had it in maven).

PR #1509, opened 2018-12-01
Add button to open container in file browser

### LINKED ISSUE #1441: Add button to open container in file manager
Hello, It would be great if I select container then there was button for example "Open directory", and after click it would open directory of container in file manager :)

plata:
A simple button on the container info tab.

fixes #1441

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
@@ -34,6 +34,9 @@ import org.phoenicis.repository.RepositoryManager;
 import org.phoenicis.repository.dto.ApplicationDTO;
 import org.phoenicis.repository.dto.RepositoryDTO;
 
+import java.awt.*;
+import java.io.File;
+import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
```diff
@@ -34,6 +34,9 @@
 import org.phoenicis.repository.dto.ApplicationDTO;
 import org.phoenicis.repository.dto.RepositoryDTO;
 
+import java.awt.*;
```
madoar: Do we really require `awt` for this?
Is there no way to do this with javafx?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerInformationTab.java
```diff
@@ -110,11 +111,23 @@ private void populate() {
         Button deleteButton = new Button(tr("Delete container"));
         deleteButton.setOnMouseClicked(event -> this.onDeletePrefix.accept(container));
 
-        informationPane.getChildren().addAll(informationContentPane, spacer, /* changeEngineComboBox, */ deleteButton);
+        Region buttonSpacer = new Region();
+        buttonSpacer.setPrefHeight(20);
```
madoar: This looks like something we want to move to css/less

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
plata: I searched for it. If there is a way, I couldn't find it.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerInformationTab.java
plata: Yes, but not in this pr (it's the same like the other spacer above).

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
plata: This is to open something in the default browser. I'm not sure it's a good idea. Check also the comment in the second link:

> On my computer starts the default browser and shows the PDf file in the browser. Why? (Host is Linux Mint).

madoar: Yes, you're right.
I'm fine with merging this for now.
But this solution is not really platform independent, because it runs only on standard JVMs which provide AWT, which Android for example doesn't.

As a long term solution, maybe we can implement a similar solution to our terminal opener?

qparis:
+1, we need a open() command (we can use xdg-open Linux and open on OSX)

plata:
Can you create a new issue for the `open()` command?

PR #1523, opened 2018-12-02
Refactor sidebar groups with MVC in mind

madoar:
This PR refactors the two classes `SidebarGroup` and `SidebarToggleGroup` with MVC in mind.
For the `SidebarToggleGroup` this means that the class is split in one base class and one concrete implementation for each tab.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/control/ApplicationSidebarToggleGroup.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * A toggle group component used inside the {@link ApplicationsSidebar}
+ */
+public class ApplicationSidebarToggleGroup
+        extends SidebarToggleGroupBase<CategoryDTO, ApplicationSidebarToggleGroup, ApplicationSidebarToggleGroupSkin> {
+    /**
+     * A consumer, which is called when the "all" categories button has been selected
+     */
+    private final ObjectProperty<Runnable> onAllCategorySelection;
```
plata: Can't we have this in the base class? Now it's repeated in all derived classes. Same for `onCategorySelection`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/control/ApplicationSidebarToggleGroup.java
```diff
@@ -0,0 +1,69 @@
...
+
+        this.onAllCategorySelection = new SimpleObjectProperty<>();
+        this.onCategorySelection = new SimpleObjectProperty<>();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public ApplicationSidebarToggleGroupSkin createSkin() {
```
plata: Is this used anywhere?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationsSidebar.java
```diff
@@ -141,8 +140,9 @@ private void populateFilters() {
         this.operatingSystemCheck.selectedProperty().bindBidirectional(filter.containAllOSCompatibleApplications());
         this.operatingSystemCheck.setSelected(false);
 
-        this.filterGroup = new SidebarGroup("Filters", testingCheck, requiresPatchCheck, commercialCheck,
-                operatingSystemCheck);
+        this.filterGroup = new SidebarGroup<>("Filters");
```
plata: Shouldn't this be translated?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/control/ApplicationSidebarToggleGroup.java
madoar: Normally yes, but I'm not sure if it should belong to the base class, because it is not used there.
In addition not all `SidebarToggleGroup` implementations require it.
These are the reasons why i decided to add the required callback methods to every `SidebarToggleGroup` implementation.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/control/ApplicationSidebarToggleGroup.java
madoar: Yes, by JavaFX internally.
The method is called inside the [createDefaultSkin](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Control.html#createDefaultSkin--) method of our control objects.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/apps/ApplicationsSidebar.java
madoar: You're right

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/control/ApplicationSidebarToggleGroup.java
plata: Ok.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/control/ApplicationSidebarToggleGroup.java
plata: Ok.

PR #1533, opened 2018-12-03
Fix: cannot access TerminalOpener from Javascript with Java 10+

### LINKED ISSUE #1527: How to run winetricks command inside container
Hello, I just have to run this command `winetricks settings fontsmooth=rgb` inside my container, how can I do it?

plata:
fixes #1527


madoar:
Is this the only button/functionality that doesn't work with Java 9+?

plata:
Yes because the module was not opened.

PR #1536, opened 2018-12-03
Add additional properties to SearchBox and SidebarToggleGroupBase

madoar:
This PR:
- adds an additional property for the entered search term to the `SearchBox`
- adds an additional property for the selected element to the `SidebarToggleGroupBase`
- fixes a bug in the text property listener of the search field in the `SearchBox`

This PR assumes, that empty search terms (including the initial state) are represented by an empty (i.e. "") string.
This PR should allow us to use [createObjectBinding](https://openjfx.io/javadoc/11/javafx.base/javafx/beans/binding/Bindings.html#createObjectBinding(java.util.concurrent.Callable,javafx.beans.Observable...)) to create a bindable property for our filter function.
This in turn should allow us to use the official `FilteredList` class provided by JavaFX, because changes in the properties defining the filter function can then be used to invalidate the filter function property of the `FilteredList`.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/control/SidebarToggleGroupBase.java
```diff
@@ -27,17 +31,24 @@
      */
     private final ObservableList<E> elements;
 
+    /**
+     * The selected element or null if no/all elements has been selected
```
plata: have been

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/control/SidebarToggleGroupBase.java
```diff
@@ -27,17 +31,24 @@
...
+     */
+    private final ObjectProperty<E> selectedElement;
+
     /**
      * Constructor
      *
      * @param title The title of the sidebar toggle group
      * @param elements The elements to be shown inside the sidebar toggle group
+     * @param selectedElement The selected element or null if no/all elements has been selected
```
plata: have been

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/skin/ApplicationSidebarToggleGroupSkin.java
```diff
@@ -31,7 +31,10 @@ public ApplicationSidebarToggleGroupSkin(ApplicationSidebarToggleGroup control)
         final ToggleButton allCategoryButton = createSidebarToggleButton(tr("All"));
 
         allCategoryButton.setId("allButton");
-        allCategoryButton.setOnAction(event -> getControl().getOnAllCategorySelection().run());
+        allCategoryButton.setOnAction(event -> {
+            getControl().setSelectedElement(null);
```
plata: Thinking if it would be nice to have something like `setNothingSelected`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/skin/ApplicationSidebarToggleGroupSkin.java
```diff
@@ -44,7 +47,10 @@ protected ToggleButton convertToToggleButton(CategoryDTO category) {
         final ToggleButton categoryButton = createSidebarToggleButton(category.getName());
 
         categoryButton.setId(String.format("%sButton", category.getId().toLowerCase()));
-        categoryButton.setOnAction(event -> getControl().getOnCategorySelection().accept(category));
+        categoryButton.setOnAction(event -> {
+            getControl().setSelectedElement(category);
+            getControl().getOnCategorySelection().accept(category);
```
plata: Wouldn't it make sense to combine these two calls?

madoar: I have a better idea:
Why don't we remove the callback methods from the sidebar toggle groups and the search box?
Instead we can subscribe directly to the properties for changes.
This makes it also much easier to add listeners where they are needed.

plata: My problem with this is that you would need to know the internal implementation (i.e. the properties). With the call back, this is transparent.

madoar: I'm not sure what you mean?
Don't you have the same problem with the callback methods?
If you don't know them you can't use them, it's the same with properties.

The properties are quite easy to use. You can just add your callback/listener to the property by calling `Property#addListener`

plata: You're right.

madoar:
Ok, I've taken the PR a step further and updated the tabs (including their sidebars) to use the properties  via `createObjectBinding`.
The only issue I detected is that the engines tab is currently kind of broken.
The reason for this is, that the `engineCategoryDTOS` passed to `public void populate(List<EngineCategoryDTO> engineCategoryDTOS, Map<String, Engine> newEngines)` contain an empty list of subcategories, which leads to `engineSubCategoryTabs` in [line 126](https://github.com/PhoenicisOrg/phoenicis/blob/46b2975637d33b1e01dd529ea5a20db092a1048d/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginesView.java#L126) being empty.

@plata do you know why this is the case?

plata:
Not really. Does it also happen without your changes? I've checked this a few days ago (when I merged Wine 3.0.4) and it was working.

madoar:
No, it doesn't happen without my changes.
Therefore I would like to know how the engines are loaded, because it seems like the sub-categories are loaded at a later time (see https://github.com/PhoenicisOrg/phoenicis/blob/master/phoenicis-engines/src/main/java/org/phoenicis/engines/EnginesManager.java#L98).

madoar:
I think we need to straight the engine loading part out.
The best approach in my opinion is the following workflow:
1) load the repository containing the engine information
2) load the engines
3) load the engine categories
4) if (3) doesn't contain the loading of the sub-categories.
   Ideally we do this as part of (3)
5) show/update the view

madoar:
Ok, I got the engines tab working as intended.

@plata can you please test the library tab (and the other tabs too) to see if they are working for you?
I'm currently unable to test the library tab because the installation stalls during the wine download...

plata:
I will. Can we maybe split the rework of the engines from the rest? This PR is quite big now.

plata:
I've checked it. Everything working as far as I can tell.

PR #1555, opened 2018-12-08
Improve the javafx observable list implementations

madoar:
This PR is part of #1536


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/lists/InitialisationChange.java
```diff
@@ -0,0 +1,53 @@
+package org.phoenicis.javafx.views.common.lists;
+
+import javafx.collections.ListChangeListener;
+import javafx.collections.ObservableList;
+
+import java.util.Collections;
+import java.util.List;
+
+public class InitialisationChange<E> extends ListChangeListener.Change<E> {
```
plata: Javadoc: what does this class do?

PR #1556, opened 2018-12-08
Fix fetching process of the engines before the engines tab is populated

madoar:
Currently the loading process of the engines for the engines tab is strange.
The engine categories are loaded without their subcategories.
This in turn should prevent them from showing up in the sidebar of the engines tab, because they are then assumed to be empty.

This PR solves this problem by additionally fetching all subcategories of the fetched engine categories, before the engines tab is populated.

This PR is part of #1536

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
@@ -56,7 +56,7 @@ public class EnginesController {
     private Map<String, List<EngineSubCategoryDTO>> versionsCache = new HashMap<>();
 
     public EnginesController(EnginesView enginesView, RepositoryManager repositoryManager,
-                             EnginesManager enginesManager, ThemeManager themeManager) {
+            EnginesManager enginesManager, ThemeManager themeManager) {
         super();
 
         this.enginesView = enginesView;
@@ -165,12 +165,12 @@ public class EnginesController {
      * Fetches all engine subcategories that belong to a given list of engine categories
      *
      * @param engineCategories The engine categories
-     * @param result           The temporary transport variable
-     * @param callback         A callback method, which is called after all engine subcategories have been fetched
+     * @param result The temporary transport variable
+     * @param callback A callback method, which is called after all engine subcategories have been fetched
      */
     private void fetchEngineSubcategories(Queue<EngineCategoryDTO> engineCategories,
-                                          Map<EngineCategoryDTO, List<EngineSubCategoryDTO>> result,
-                                          Consumer<Map<EngineCategoryDTO, List<EngineSubCategoryDTO>>> callback) {
+            Map<EngineCategoryDTO, List<EngineSubCategoryDTO>> result,
+            Consumer<Map<EngineCategoryDTO, List<EngineSubCategoryDTO>>> callback) {
         final Queue<EngineCategoryDTO> queue = new ArrayDeque<>(engineCategories);
 
         if (queue.isEmpty()) {
@@ -185,7 +185,7 @@ public class EnginesController {
                     versions -> {
                         // recursively process the remaining engine categories
                         fetchEngineSubcategories(queue,
-                                ImmutableMap.<EngineCategoryDTO, List<EngineSubCategoryDTO>>builder()
+                                ImmutableMap.<EngineCategoryDTO, List<EngineSubCategoryDTO>> builder()
                                         .putAll(result)
                                         .put(engineCategory, versions)
                                         .build(),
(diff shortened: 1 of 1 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
```diff
@@ -39,78 +40,85 @@
...
+
         this.enginesView = enginesView;
-        this.repositoryManager = repositoryManager;
         this.enginesManager = enginesManager;
         this.themeManager = themeManager;
 
-        this.repositoryManager.addCallbacks(repositoryDTO -> this.enginesManager.fetchAvailableEngines(repositoryDTO,
-                engines -> this.populateView(repositoryDTO, engines),
-                e -> Platform.runLater(() -> enginesView.showFailure(tr("Loading engines failed."), Optional.of(e)))),
+        repositoryManager.addCallbacks(
```
plata: Why did your remove `this` in several places?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
```diff
@@ -121,26 +129,84 @@ public EnginesView getView() {
...
 
     /**
-     * forces an update of the view
+     * Fetches all engine subcategories that belong to a given list of engine categories
+     *
+     * @param engineCategories The engine categories
+     * @param result The temporary transport variable
+     * @param callback A callback method, which is called after all engine subcategories have been fetched
+     */
+    private void fetchEngineSubcategories(Queue<EngineCategoryDTO> engineCategories,
```
plata: Why `Queue`? What's the advantage to use recursion here?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
madoar: I didn't notice this myself.

The changes reflect my current programming style.
I try to use `this` only for assignments or it's otherwise required, for example in inner classes.
Otherwise I depend on my IDE to show me whether a variable is defined on an object scope or a method scope.

This approach has the benefit in my eyes that it shortens the code.
In addition it makes it easier to see the important object scoped assignments via `this`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
madoar: About the `Queue`:
A queue seems like the best data structure for this use case to me, because it allows for the retrieval and removal of the first object at the same time.
Ok, you could now argue that this is also possible with other collection types like `ArrayList` through the use of `remove()`, but in my opinion `remove` has a different semantic meaning.

About the use of `recursion`:
Recursion seemed to me like the only solution to implement this functionality without rewriting `EnginesManager`.
The issue about all our manager classes (e.g. `EnginesManager` and `RepositoryManager`) is that they all work with callbacks.
This makes it quite hard to concatenate multiple method calls without calling the "next" method inside the callback of the "current" method call.
This cries for a recursive solution to me.
Still I think it would be better to provide a "flat" way to work with our manager classes, i.e. a way which doesn't require callback methods.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
plata: I understand. However, the Phoenicis code mainly uses `this` whenever class members are accessed. I think we should keep this style until decided otherwise (especially in this case because the diff shows changes which are not really changes).

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
plata: Ok. Yes, we should think about the callbacks again.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
madoar: These are quite a lot of unnecessary `this`.
I think it would be better to change the code specification.
I agree that it is a bad idea to simply change the use of `this` in methods/places you otherwise don't modify.
Therefore I propose to only change it when a method is rewritten.

What do you think?

plata: We should discuss this topic in a separate issue and leave the code like it was for this PR.

madoar: My problem is that I've basically rewritten the whole constructor, `populateView` and added a new method `fetchEngineSubcategories` so it's hard for me revert the code to a state it has never been in.

plata: The constructor is very much the same like before. I agree for the methods.

PR #1569, opened 2018-12-13
Implement OOP Verbs

plata:
fixes https://github.com/PhoenicisOrg/scripts/issues/534

### DIFF
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
@@ -1,0 +1,13 @@
+package org.phoenicis.engines;
+
+/**
+ * interface which must be implemented by all Verbs in Javascript
+ */
+public interface Verb {
+    /**
+     * installs the Verb in the given container
+     * @param container directory name (not the complete path!) of the container where the tool shall be run
+     * @param version version of the Verb
+     */
+    void install(String container, String version);
+}
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
@@ -7,7 +7,6 @@ public interface Verb {
     /**
      * installs the Verb in the given container
      * @param container directory name (not the complete path!) of the container where the Verb shall be installed
-     * @param version version of the Verb
      */
-    void install(String container, String version);
+    void install(String container);
 }
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
@@ -6,7 +6,7 @@ package org.phoenicis.engines;
 public interface Verb {
     /**
      * installs the Verb in the given container
-     * @param container directory name (not the complete path!) of the container where the tool shall be run
+     * @param container directory name (not the complete path!) of the container where the Verb shall be installed
      * @param version version of the Verb
      */
     void install(String container, String version);
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
@@ -61,7 +61,7 @@ public class VerbsManager {
                 "include([\"engines\", \"" + engineId + "\", \"verbs\", \"" + verbId + "\"]);",
                 ignored -> interactiveScriptSession.eval("new Verb()", output -> {
                     final Verb verbObject = (Verb) output;
-                    verbObject.install(container, "");
+                    verbObject.install(container);
                     doneCallback.run();
                 }, errorCallback), errorCallback);
     }
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (C) 2015-2017 PÂRIS Quentin
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.engines;
+
+import org.phoenicis.repository.dto.ApplicationDTO;
+import org.phoenicis.repository.dto.CategoryDTO;
+import org.phoenicis.repository.dto.RepositoryDTO;
+import org.phoenicis.repository.dto.TypeDTO;
+import org.phoenicis.scripts.interpreter.InteractiveScriptSession;
+import org.phoenicis.scripts.interpreter.ScriptInterpreter;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+/**
+ * manages the Verbs
+ */
+public class VerbsManager {
+    private final ScriptInterpreter scriptInterpreter;
+
+    /**
+     * constructor
+     * @param scriptInterpreter
+     */
+    public VerbsManager(ScriptInterpreter scriptInterpreter) {
+        this.scriptInterpreter = scriptInterpreter;
+    }
+
+    /**
+     * installs a Verb in a given container
+     * @param engineId ID of the engine which provides the Verb (e.g. "Wine")
+     * @param container name of the container
+     * @param verbId ID of the Verb
+     * @param doneCallback callback executed after the script ran
+     * @param errorCallback callback executed in case of an error
+     */
+    public void installVerb(String engineId, String container, String verbId, Runnable doneCallback,
+            Consumer<Exception> errorCallback) {
+        final InteractiveScriptSession interactiveScriptSession = scriptInterpreter.createInteractiveSession();
+
+        interactiveScriptSession.eval(
+                "include([\"engines\", \"" + engineId + "\", \"verbs\", \"" + verbId + "\"]);",
+                ignored -> interactiveScriptSession.eval("new Verb()", output -> {
+                    final Verb verbObject = (Verb) output;
+                    verbObject.install(container, "");
+                    doneCallback.run();
+                }, errorCallback), errorCallback);
+    }
+
+    /**
+     * fetches the available Verbs
+     * @param repositoryDTO
+     * @param callback
+     */
+    public void fetchAvailableVerbs(RepositoryDTO repositoryDTO, Consumer<Map<String, ApplicationDTO>> callback) {
+        Map<String, ApplicationDTO> verbs = new HashMap<>();
+        // get engine CategoryDTOs
+        List<CategoryDTO> categoryDTOS = new ArrayList<>();
+        for (TypeDTO typeDTO : repositoryDTO.getTypes()) {
+            if (typeDTO.getId().equals("engines")) {
+                categoryDTOS = typeDTO.getCategories();
+            }
+        }
+        for (CategoryDTO engine : categoryDTOS) {
+            for (ApplicationDTO applicationDTO : engine.getApplications()) {
+                if (applicationDTO.getId().equals("verbs")) {
+                    verbs.put(engine.getId(), applicationDTO);
+                }
+            }
+        }
+        callback.accept(verbs);
+    }
+}
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
@@ -165,11 +165,11 @@ public class ContainersController {
 
     private void updateVerbs(RepositoryDTO repositoryDTO) {
         this.verbsManager.fetchAvailableVerbs(repositoryDTO,
-                verbs -> Platform.runLater(() -> this.verbs = verbs));
+                verbs -> this.verbs = verbs);
     }
 
     private void updateEngineTools(RepositoryDTO repositoryDTO) {
         this.engineToolsManager.fetchAvailableEngineTools(repositoryDTO,
-                engineTools -> Platform.runLater(() -> this.engineTools = engineTools));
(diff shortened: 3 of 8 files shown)


phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param repositoryDTO
+     * @param callback
+     */
+    public void fetchAvailableVerbs(RepositoryDTO repositoryDTO, Consumer<Map<String, ApplicationDTO>> callback) {
+        Map<String, ApplicationDTO> verbs = new HashMap<>();
+        // get engine CategoryDTOs
+        List<CategoryDTO> categoryDTOS = new ArrayList<>();
+        for (TypeDTO typeDTO : repositoryDTO.getTypes()) {
+            if (typeDTO.getId().equals("engines")) {
+                categoryDTOS = typeDTO.getCategories();
```
madoar: Is there always only a single `TypeDTO` containing the engines?

phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+        for (CategoryDTO engine : categoryDTOS) {
+            for (ApplicationDTO applicationDTO : engine.getApplications()) {
+                if (applicationDTO.getId().equals("verbs")) {
+                    verbs.put(engine.getId(), applicationDTO);
```
madoar: Is there always only a single `ApplicationDTO` for every `CategoryDTO`, which contains verbs?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
```diff
@@ -0,0 +1,94 @@
...
+                        e -> Platform.runLater(() -> new ErrorMessage("Error", e).show()));
+            });
+            this.lockableElements.add(verbButton);
+            verbsContentPane.getChildren().add(verbButton);
+        }
+
+        verbsPane.getChildren().add(verbsContentPane);
+
+        final ScrollPane scrollPane = new ScrollPane();
+        scrollPane.setHbarPolicy(ScrollPane.ScrollBarPolicy.NEVER); // the TilePane adjusts the number of columns
```
madoar: This line should go to the css files

phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * fetches the available Verbs
+     * @param repositoryDTO
+     * @param callback
+     */
+    public void fetchAvailableVerbs(RepositoryDTO repositoryDTO, Consumer<Map<String, ApplicationDTO>> callback) {
+        Map<String, ApplicationDTO> verbs = new HashMap<>();
+        // get engine CategoryDTOs
+        List<CategoryDTO> categoryDTOS = new ArrayList<>();
```
madoar: This looks like something to be done using streams.

phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
```diff
@@ -0,0 +1,13 @@
+package org.phoenicis.engines;
+
+/**
+ * interface which must be implemented by all Verbs in Javascript
+ */
+public interface Verb {
```
madoar: What about a unique name?
How does Phoenicis remember, whether a verb has been installed previously or not?

phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
plata: That's just copy paste from the engine tools.

phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
plata: That's just copy paste from the engine tools.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
plata: That's just copy paste from the engine tools.

phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
plata: That's just copy paste from the engine tools.

phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
plata: It won't. Check @qparis comment in https://github.com/PhoenicisOrg/scripts/issues/534.

madoar: Yes, but I think Phoenicis should remember which verbs were installed on a container.
Otherwise, the user can't track what he did to his container.

phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
madoar: Which makes it even more important to find an answer to these questions, now that they can affect two classes...

phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
plata: So some kind of database? I guess we can consider that. However, we should do that later otherwise this PR will blow up too much.

phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
plata: Ok. In this case I would say: Yes, all engines have to be in the "engines" type.

phoenicis-engines/src/main/java/org/phoenicis/engines/Verb.java
madoar: Yes, but I think in our case a simple json file per container, containing a list of verb ids/names would be enough.
I agree we can do this in a later PR.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
qparis: Do you think that some of the code would be factored?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
```diff
@@ -151,6 +163,11 @@ private void updateEngineSettings(RepositoryDTO repositoryDTO) {
                         .runLater(() -> new ErrorMessage(tr("Loading engine tools failed."), e, this.containersView)));
     }
 
+    private void updateVerbs(RepositoryDTO repositoryDTO) {
+        this.verbsManager.fetchAvailableVerbs(repositoryDTO,
+                verbs -> Platform.runLater(() -> this.verbs = verbs));
```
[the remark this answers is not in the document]
qparis: Yep, JavaFX class should *never* be used outside Phoenicis-javafx

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
plata: Any suggestions/what would you expect?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+
+import static org.phoenicis.configuration.localisation.Localisation.tr;
+
+public class ContainerVerbsTab extends Tab {
+    private static final String CONFIGURATION_PANE_CSS_CLASS = "containerConfigurationPane";
+    private static final String TITLE_CSS_CLASS = "title";
+
+    private final ContainerDTO container;
+    private VerbsManager verbsManager;
```
madoar: I believe the `VerbsManager` is final too, right?

phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public void installVerb(String engineId, String container, String verbId, Runnable doneCallback,
+            Consumer<Exception> errorCallback) {
+        final InteractiveScriptSession interactiveScriptSession = scriptInterpreter.createInteractiveSession();
+
+        interactiveScriptSession.eval(
+                "include([\"engines\", \"" + engineId + "\", \"verbs\", \"" + verbId + "\"]);",
+                ignored -> interactiveScriptSession.eval("new Verb()", output -> {
+                    final Verb verbObject = (Verb) output;
+                    verbObject.install(container, "");
```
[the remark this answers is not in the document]
plata: I've updated the implementation.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
```diff
@@ -0,0 +1,94 @@
...
+import org.phoenicis.repository.dto.ScriptDTO;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.phoenicis.configuration.localisation.Localisation.tr;
+
+public class ContainerVerbsTab extends Tab {
+    private static final String CONFIGURATION_PANE_CSS_CLASS = "containerConfigurationPane";
+    private static final String TITLE_CSS_CLASS = "title";
```
madoar: Can we directly use the css classes in the files please?
I'm not a huge fan of global variables for css classes

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
```diff
@@ -0,0 +1,94 @@
...
+            this.lockableElements.add(verbButton);
+            verbsContentPane.getChildren().add(verbButton);
+        }
+
+        verbsPane.getChildren().add(verbsContentPane);
+
+        final ScrollPane scrollPane = new ScrollPane();
+        scrollPane.setHbarPolicy(ScrollPane.ScrollBarPolicy.NEVER); // the TilePane adjusts the number of columns
+                                                                    // already
+        verbsContentPane.prefWidthProperty().bind(scrollPane.widthProperty());
```
madoar: Does it work with `verbsContentPane.setPrefWidth(0)` or `verbsContentPane.setPrefWidth(-1)` too?
As far as I know this means, that the container should use the width of its parent.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
```diff
@@ -0,0 +1,94 @@
...
+        this.verbsManager = verbsManager;
+
+        this.setClosable(false);
+
+        this.populate(verbs);
+    }
+
+    private void populate(ApplicationDTO verbs) {
+        final VBox verbsPane = new VBox();
+        final Text title = new TextWithStyle(tr("Verbs"), TITLE_CSS_CLASS);
```
madoar: It's a goal of mine to remove the class `TextWithStyle` can you please directly create a `Text` object?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
```diff
@@ -0,0 +1,94 @@
...
+import org.phoenicis.javafx.views.common.TextWithStyle;
+import org.phoenicis.repository.dto.ApplicationDTO;
+import org.phoenicis.repository.dto.ScriptDTO;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.phoenicis.configuration.localisation.Localisation.tr;
+
+public class ContainerVerbsTab extends Tab {
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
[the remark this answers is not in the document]
madoar: I agree with @plata, we need to sometimes call `Platform.runLater` inside a callback passed to a manager class like `ApplicationsManager`, `EnginesManager` or `VerbsManager`, because of how the callbacks work.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
madoar: @qparis I'm not sure what you mean?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
plata: I'm not sure and I don't really want to change this now. If you want to change it, we should handle this in a separate issue and also fix the same code for engine tools at the same time.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
plata: We should handle this in a separate issue and also fix the same code for engine tools at the same time.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
plata: Done.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
plata: Yes.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/containers/ContainerVerbsTab.java
plata: Done.

plata:
@madoar please check also the PR in scripts (they must be merged at the same time).

PR #1580, opened 2018-12-15
Shortcut creation: use fallback miniature/icon if miniature/icon cannot be accessed

### LINKED ISSUE #1534: Exception when trying to edit a shortcut
It fails in `ShortcutManager` for `shortcutDTO.getIcon()` because the default (`phoenicis.png` inside the jar) cannot be used as a file.

plata:
Fixes #1534

### DIFF
--- a/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
+++ b/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
@@ -79,7 +79,13 @@ public class ShortcutManager {
 
             try {
                 if (shortcutDTO.getIcon() != null) {
-                    File file = new File(shortcutDTO.getIcon());
+                    File file;
+                    try {
+                        file = new File(shortcutDTO.getIcon());
+                    } catch (IllegalArgumentException e) {
+                        // fallback to phoenicis.png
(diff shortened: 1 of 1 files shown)


phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
```diff
@@ -79,7 +79,13 @@ public void createShortcut(ShortcutDTO shortcutDTO) {
 
             try {
                 if (shortcutDTO.getIcon() != null) {
-                    File file = new File(shortcutDTO.getIcon());
+                    File file;
```
madoar: Can we maybe move the `File` creation to a new method?
I mean both changes look equal to me with the difference of two parts, which could be specified using two input parameters.

plata: Technically that's of course possible but honestly it doesn't make much sense for me. If you really want it, I can do it.

madoar: I don't really like this kind of object creation where we begin with `File file;`.
Therefore I think a function would be a more clean solution for this.

PR #1587, opened 2018-12-18
Refactor ApplicationPanel with MVC in mind

madoar:
This PR reimplements the `ApplicationPanel` with the MVC pattern.

An interesting introduction of this PR is the way the new `ApplicationDetailsPanel` is opened and closed. This task is solved by a conditional binding, which checks whether an application has been selected in the list widget or not. If no application has been selected, i.e. the selection is `null`, the details panel is hidden, otherwise it's shown.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/application/skin/ApplicationDetailsPanelSkin.java
```diff
@@ -0,0 +1,221 @@
...
+        closeButton.getStyleClass().add("closeIcon");
+        closeButton.setOnAction(event -> getOnClose().ifPresent(Runnable::run));
+
+        final Region filler = new Region();
+        HBox.setHgrow(filler, Priority.ALWAYS);
+
+        return new HBox(titleLabel, filler, closeButton);
+    }
+
+    private void installScript(ScriptDTO script) {
```
madoar: I've moved the following from`the `AppsController` class to this class.
I'm not sure where this should go in the end, but I think it shouldn't be set from too far outside (i.e. from `AppsController`).
Ideally I prefer to specify the action, when an install button has been clicked, either in this class or in the `ApplicationsView` class.

@plata what do you think?

plata: In my opinion, these new MVC widgets shall only be the widgets so I don't think that application logic (like `installScript`) belongs here. In general, we must be a bit careful that we do not make 2 MVC in parallel (one the new widgets and the other one the existing views/controllers).

For the lists, the implementation makes sense because they are something which is only used by the application (like you would use any other JavaFX control). With the current details panel MVC, I have the feeling that it has too much application logic inside.

Not quite sure if I could express myself correctly. Does it make sense to you?

madoar: The question is, where should the application logic go?
I wouldn't separate it too far from the components where the logic is used afterwards (which is why I don't like our old controller classes, e.g. `AppsController`, etc.).
In addition the `ApplicationDetailsPanel` is not really a universal usable component, you can **only** use it for `ApplicationDTO` objects.
Therefore I think we should place the `installScript` method either inside the `ApplicationDetailsPanel` class or inside `ApplicationsView`, which is kind of the root class/component for the applications.

madoar:
This PR is ready for review now :)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/application/skin/ApplicationDetailsPanelSkin.java
plata: I agree to the part with "logic too far from components".

Like you said, the `ApplicationDetailsPanel` is really specific (not a widget but application code). Therefore it should be handled in a similar way to the `ApplicationsView`.

madoar: Ah ok.
Long term I'm planning to move the content from the classes in the `controller` package into the components. The `installScript` method is just the first part of this task.

Please be aware that this is a long term goal which could take quite a while until it has been achieved.

plata: I would like to hear @qparis opinion about this approach as well.

madoar: @qparis what do you think about the suggested changes?

qparis: I think installScript cannot stand here for these. To my comprehension, it is not really "installing a script", but it is instanciating the javascript interpreter so that it can run the script. Therefore, this could be used somewhere else (Does Phoenicis CLI still work?) and I guess should be part of another module.

madoar: Just to summarize:
Your comment is about two parts:
- the name of the method is wrong, it should be more generic
- the method is in the wrong place

The part where a script is executed is already port of another module.
It is part of the class `ScriptInterpreter`, which takes the script and executes it.
I don't think that the name is wrong, because the method is **only** used as the action listener of the `Install` button for each script (see https://github.com/PhoenicisOrg/phoenicis/pull/1587/files/10b7e81885cb62c89596d4371c345f4767d342fe#diff-bb0cc0e237c237e42af7cb1e380541f8R161).
The method is not used in other places, therefore it can be moved to the class where it is used.

qparis: What about all the logic where you are setting JS variables? Should'n be used somewhere else also? 
I think we would need to have a runInstaller(typeId, categoryId, applicationId, scriptId, scriptContent) at some places

madoar: That sounds like a good idea, but in my opinion unrelated to this PR.
Should I open a new issue?

qparis: You can, or you can do it directly as it is a small change. Up to you :)

madoar: Then let me open a new issue.
I would like to keep the addition of a `runInstaller(typeId, categoryId, applicationId, scriptId, scriptContent)` method separate.

plata: @qparis the real question in this PR is how we handle the MVC (do we delete the `controllers` package?). What's your opinion regarding that?

madoar: I should adjust my previous statement:
I'm not sure yet whether we can delete the complete controllers package, because we need to do the (lazy) initialization somewhere and I'm not sure whether we can cleanly move this to the components.

If we can't move the initialization code to the components this code will stay in some kind of controller classes, but all other codes, especially listener/controller code, which is currently set via setter methods should be moved to the component classes where they are used.

plata: Personally, I don't have enough experience with JavaFX to really provide any educated opinion regarding this topic so I'll trust whatever you two decide ;).

madoar: @qparis I've created a new issue for the `runInstaller` method: #1659

madoar:
@qparis has marked the conversation as closed, I guess this means he approves the changes?

qparis:
Yes sorry, I did this on my phone and the message was not sent

plata:
Like I said: I can't quite imagine yet how this will look like if we follow this approach for all our GUI classes (especially if we need the `Controller` classes as well) but I have not enough experience with JavaFX to really say anything so I will trust you and @qparis.

madoar:
As mentioned above @qparis approves my changes.

About moving the code from the `Controller` classes into the GUI components:
I don't think we should discuss this topic inside this PR in detail.
One reason for this is, that this is not the main topic of this PR (which is mainly about refactoring the `ApplicationsPanel`).
In addition I don't think that we can decide this now for all `Controller` classes, and all methods they contain, because moving the controller code to the GUI components should happen in an iterative way across multiple pull requests.
This in turn allows us to discuss this for each case separately.

plata:
I understand. I just wouldn't like to end up in a situation where some methods are in a `Controller` class and some not.

@qparis should then also formally approve this.

plata:
Ok.

PR #1596, opened 2018-12-18
Set version to 5.0-alpha-2

plata:


.travis.yml
```diff
@@ -3,7 +3,8 @@ language: java
 # avoid double builds on pull requests
 branches:
   only:
-    master
+    - master
+    - release/5.0-alpha-2
```
qparis: I think we should use a tag, not a branch

madoar: Branches have the benefit that we're able to push bugfixes later on.
Not sure if we require this, especially in the alpha.

plata: When we released alpha, it was very good to have a branch because it enabled us to continue working on master without breaking the release (we had to fix some bugs on the release branch before we could really publish it).
Honestly we had a discussion about this topic at that time already and I don't think we should start it again (at least not now).

qparis: A release is a release, it should not be modifed (or we need to create another release), so it is a tag. 

Whether we decide to maintain several release streams (i.e. branch of tags) is another story. I'm not sure we need that honestly

plata: Sure, it will be a tag. Just not on master but on the branch. Might be that we never need to change the branch again. But in case we have to, we can do it. It's just an opportunity which doesn't hurt us in any way.

qparis: Yep, ok.

Let’s just use different code names for tags and branch’s and we are good :-)

plata: Fine with that.

plata:
Will merge this now. Codacy doesn't seem to finish and there's nothing it could find anyways.

PR #1605, opened 2018-12-20
Remove some unnecessary JavaFX components

madoar:
This PR removes the following JavaFX classes:
- `WaitPanel`
- `SidebarSpacer`
- `SidebarScrollPane`
- `SidebarCheckBox`
- `SidebarButton`
- `MessagePanel`


phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/ui/MainWindowView.java
```diff
@@ -66,15 +67,21 @@ public MainWindowView(String text, ThemeManager themeManager) {
...
+        this.waitPanel = createWaitPanel();
 
         this.content.setCenter(waitPanel);
 
         this.setContent(content);
     }
 
-    private void populateWaitPanel() {
-        this.waitPanel = new WaitPanel();
+    private HBox createWaitPanel() {
```
plata: With this implementation, the wait panel can only be used in `MainWindowView`. Not sure if that's important or not.

madoar: The wait panel was only used in the `MainWindowView`.
Otherwise it would be worth a thought to add a separate control and skin class for it.

plata: We can always do that later if we need to.

PR #1607, opened 2018-12-21
Refactor confirm message

madoar:
This PR replaces the `ConfirmMessage` class with a `ConfirmDialog` class, which can be configured via a builder.


plata:
Does it still look the same? Otherwise, can you provide a screenshot?

madoar:
As far as I tested it looked the same.
The `ConfirmDialog`s are just easier and safer to use than the `ConfirmMessage`s in my opinion.

plata:
To be honest, I don't really see the advantage yet. Can you give an example in the pr description?

madoar:
In my opinion the following old initialization is quite fuzzy:
```java
ConfirmMessage confirmMessage = new ConfirmMessage(
    tr("Delete {0} container", containerToDelete.getName()),
    tr("Are you sure you want to delete the {0} container?", containerToDelete.getName()),
    this.containersView.getContent().getScene().getWindow());
...
```
One open questions I ask myself when I see this statement is: What does each string stand for?
In addition I think the usage of the `ask(yes, no)` method is quite unclear, because of the same reason.

The builder pattern in turn makes all the input explicit, you can see through the `withX(X x)` method what exactly you specify in the dialog.
It is also much clearer to not specify things, because you don't need to pass `null` parameters.

madoar:
Another issue I have with the old `ConfirmMessage` class is, that it opens the dialog in its constructor, this in my opinion is not transparent to other developers using the `ConfirmMessage` class.

PR #1608, opened 2018-12-21
Restructure the list widget package structure

madoar:
This PR moves the classes for the list widgets to more fitting subpackages.
All remaining classes in package `org.phoenicis.javafx.views.common.widgets.lists` will be removed or reimplemented in the future.

In addition the class `ListWidgetEntry` has been renamed to `ListWidgetElement` and the class `AdditionalListWidgetInformation` has been renamed to `ListWidgetAdditionalInformation`.


madoar:
Ideas how to better structure our JavaFX classes are always welcome :)

Issue #1512, opened 2018-12-02
Lazy load repository


plata:
The repository (and thus the apps and engines) shall only be loaded if they are required.

plata:
@madoar any ideas how this should work together with the repository update callback?

qparis:
I'm not sure than they shall be lazy loaded. However, I think that the loading should be non-blocking as possible. (i.e., use an old / local version as long as the loading process is not finished)

plata:
That won't work. Especially if you have a slower PC, the background loading will block the rest of the application (e.g. because the git cloning is writing to the HDD). I can observe this behavior currently (which is why I did the preloading in the splash).

qparis:
This seems weird. I mean, in general you can use your PC while you are cloning a git repository inside another process, can't you?

plata:
Yes. I'm not saying it doesn't work. It's just really slow. This leads to effects like library doesn't show anything for 1-2 minutes.

qparis:
Is the cloning stuff running in Platform.runLater by any chance?

plata:
I don't think so. But I'm not sure. Would have to check it in the code.

madoar:
I will need to take a look at the repository loading in more detail before I can give a suggestion/solution.

plata:
@madoar that would be great!

In general, I see two options:
1. Preload everything such that the application is completely up and running.
2. Lazy load everything which is not needed immediately.

I prefer the 2. because:
- faster startup
- smaller memory footprint
- like @qparis said, most users only want to run a shortcut from their library (I agree with that)

plata:
I just checked this again and it seems that we cannot do this: The translations for the categories (also in the library) are stored in the repository. The best I can do is create the widgets for apps, containers and engines only if required.

madoar:
@plata should this issue be closed?

plata:
Yes.

PR #1610, opened 2018-12-21
Lazy load apps, containers and engines

plata:


plata:
The repository is still loaded initially (see #1512). This PR builds the widgets for apps, container and engines when the respective tab is opened for the first time. It would be nice to avoid triggering a repository change but I'm not sure if that's possible.

madoar:
I don't really like to do this in the `setOnSelectionChanged` callback methods.
Because this doesn't suggest a loading of the entities during the first opening but during a selection change which happens afterwards.
This is especially problematic with the sidebar toggle button groups, which require the lazy loaded entities but are, strictly speaking, loaded before the selection has changed.

plata:
I don't think there's an alternative. As far as I'm aware there's no dedicated method for the first opening of a tab.

madoar:
As far as I know the skin of the tab is created during its first opening.
This means, that all components shown inside the tab should also be created upon the opening of the tab.
With this assumption, we could maybe achieve this by loading the entities in the constructor of the view skins.

madoar:
It should be quite easy, just add a logging output to a skin constructor.
Depending on when the output is shown it happens either:
- at application startup
- during the first tab opening
- or every time the tab has been opened

plata:
Ok, I will try.

plata:
I put a break point into the `ApplicationView` constructor and it got called in `ViewsConfiguration#viewApps` (obviously). I'm not sure if this is really what you wanted me to test...

madoar:
Sorry I should have clarified it some more.
I'm talking about our MVC approach.
JavaFX calls the `createDefaultSkin` method of the controls only when the control is first shown (as far as I know).
So when we implement our view classes according to the MVC approach we can make use of this assumption, by loading the entities in our implemented skin classes of the views.

So to test if my assumption is correct we need to add a debug breakpoint or a test output to one of our already existing skin classes. Ideally inside a skin class which is only shown on one tab.
These classes are the sidebar toggle group classes, for example the `ApplicationSidebarToggleGroupSkin` class.

plata:
Just tested: The skin constructor is called in `MainWindow` (`this.show`) before the GUI even opens.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/library/LibraryController.java
```diff
@@ -167,4 +191,30 @@ public void updateLibrary() {
     public LibraryView getView() {
         return libraryView;
     }
+
+    private void setDefaultCategoryIcons(List<CategoryDTO> categoryDTOS) {
```
madoar: This method looks the same as the one in `AppsController` to me.
Maybe we should move it to a common class, so that we implement it only once?
A good location for the method would be the `ThemeManager` class in my opinion.

plata: Yes, I know. I wanted to handle this later because I'm not quite sure yet how to do it. For example, it might make sense to store the category icons which are required for the shortcuts separately such that you do not need the repository.

madoar: What do you mean with repository?
Do you mean that you don't need an online connection to create the css or something else?

plata: I'll open an issue. Let's discuss it there.

madoar: Ok

PR #1611, opened 2018-12-21
Fix usage of "this" in "...View" classes

plata:


madoar:
I think this PR is a little bit too early.
My next two big projects are refactoring the view and sidebar classes with MVC.

plata:
Then let's just merge this before.

PR #1620, opened 2018-12-22
Unify setting of default category icons

### LINKED ISSUE #1617: Avoid duplicate code for setDefaultCategoryIcons
The default category icons (from the repository) are required in different places, e.g. library and apps.

Simple solution:
Move method to `ThemeManager` and call it during initial load.

A different approach could be:
- If a shortcut is created, copy the category icon to the shortcuts directory.
- Load category icons for shortcuts from shortcut directory (not repository).

Questions:
- Which icon shall be used if different repositories have the same category?

plata:
fixes #1617

I did it in `MainController` instead of `ThemeManager` because I didn't want the repository dependency in the `ThemeManager`.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
@@ -30,13 +30,25 @@ import org.phoenicis.javafx.controller.settings.SettingsController;
 import org.phoenicis.javafx.settings.JavaFxSettingsManager;
 import org.phoenicis.javafx.views.common.ThemeManager;
 import org.phoenicis.javafx.views.mainwindow.ui.MainWindow;
+import org.phoenicis.repository.RepositoryManager;
+import org.phoenicis.repository.dto.CategoryDTO;
+import org.phoenicis.repository.dto.RepositoryDTO;
+import org.slf4j.LoggerFactory;
 
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
 import java.util.Optional;
 
 import static org.phoenicis.configuration.localisation.Localisation.tr;
 
 public class MainController {
+    private final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(MainController.class);
     private final MainWindow mainWindow;
+    private final ThemeManager themeManager;
     private final JavaFxSettingsManager javaFxSettingsManager;
 
     private String applicationName;
@@ -48,6 +60,7 @@ public class MainController {
             ContainersController containersController,
             InstallationsController installationsController,
             SettingsController settingsController,
+            RepositoryManager repositoryManager,
             ThemeManager themeManager,
             JavaFxSettingsManager javaFxSettingsManager) {
         super();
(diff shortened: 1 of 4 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
```diff
@@ -30,13 +30,25 @@
...
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
 import java.util.Optional;
 
 import static org.phoenicis.configuration.localisation.Localisation.tr;
 
 public class MainController {
+    private final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(MainController.class);
```
madoar: please import the `Logger` class in the header.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
```diff
@@ -64,26 +77,28 @@ public MainController(String applicationName,
...
                 javaFxSettingsManager);
 
+        this.themeManager = themeManager;
         this.javaFxSettingsManager = javaFxSettingsManager;
 
-        installationsController.setOnInstallationAdded(mainWindow::showInstallations);
-
-        appsController.setOnAppLoaded(() -> {
-            containersController.loadContainers();
+        repositoryManager.addCallbacks(this::setDefaultCategoryIcons, e -> {
```
madoar: I think we should add an `addCallbacks` method to the `RepositoryManager` class which doesn't require an error callback

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
```diff
@@ -64,26 +77,28 @@ public MainController(String applicationName,
...
     }
 
     public void show() {
-        mainWindow.show();
+        this.mainWindow.show();
     }
 
     public void setOnClose(Runnable onClose) {
         this.mainWindow.setOnCloseRequest(event -> {
             Alert alert = new Alert(Alert.AlertType.CONFIRMATION);
```
madoar: This looks like a use case for `ConfirmDialog`

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
```diff
@@ -97,4 +112,33 @@ public void setOnClose(Runnable onClose) {
             }
         });
     }
+
+    private void setDefaultCategoryIcons(RepositoryDTO repositoryDTO) {
```
madoar: JavaDoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
madoar: You also need to change the line to:
```java
private final Logger LOGGER = LoggerFactory.getLogger(MainController.class);
```

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
madoar: Can you open a new issue for this?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
madoar: Can you open a new issue for this?

plata: #1623

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
plata: Makes sense. Copy+paste...

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
plata: #1624

PR #1625, opened 2018-12-23
Fix list widget selection bug

madoar:
This PR fixes a bug where a list widget (a) doesn't show a selection, if the selection was has been performed in another list widget (b) before (a) has been opened once.


plata:
To be honest, I didn't quite understand what the issues were in end (in the pr description).

plata:
Thanks for the explanation.

PR #1635, opened 2018-12-25
Java 8 removal and standalone package creation

qparis:
The main idea here is to get rid of Java 8, as it is really hard to maintain and package a modular and a non-modular code base at the same time. 

I propose to get rid of Maven profile and target our builds to Java 10+ (independently of the Java version used to build Phoenicis) 

**Remains to be done**
Currently, I'm not able to create a OSX native package without Java 8 (However, the good news is the app works great on OSX manually). This is for two reasons: 
* Maven jfx:native plugin is deprecated and does not work 
* jpackager (http://download2.gluonhq.com/jpackager/11/jdk.packager-osx.zip to be downloaded manually because JavaFX is no longer part of JDK) relies on jlink to make package as small as possible. The problem is that a large part of our dependencies are not modularized, so we get an error:
` automatic module cannot be used with jlink: org.apache.commons.codec from `

I believe that jlink is really promising and we should pursue this investigation: It would allow us to generate small and standalone package for all operating system that do not require JDK. 

@plata @madoar, I may need your help here


plata:
The main reason I left Java 8 when implementing support for Java 10/11 is that it is the default JDK for Ubuntu 16.04. I assume that this is still used a lot so many people might have issues to run Phoenicis without Java 8 support.

plata:
I'm not sure. Can the jlink result be used to build a .deb?

qparis:
Yes, jlink creates a lightweight JRE. Then jpackager can finish the job

plata:
If we want to do this, we should have a different issue to implement jlink support first.

qparis:
Supporting Jlink is the only way to support OSX from Java 9+ if we keep modules (AFAIK). I think it is a priority

qparis:
Well this is not a particular OSX issue. I think we rely too much on flatpak and we should not (or we are going to lose some people). 

Phoenicis binaries should work on every modern distribution (I consider flatpak to be a kind of distribution).
Therefore, we need to rationalize drastically the way the build is done, or we are going to have hundred of cases to debug when people have problems. 

As far as I am concerned, it took me two hours to make the package for the release of the alpha 2 and at the end, flatpak did not work. 

What I propose is pretty simple (in theory): 
 * We make jlink work 
 * We use jpackager http://download2.gluonhq.com/jpackager/11/jdk.packager-osx.zip and http://download2.gluonhq.com/jpackager/11/jdk.packager-linux.zip
 * We use jpackager to create a standalone .deb, .rpm, .dmg and some flatpak / zipfiles. 

This method will need some work, but at least, it will be really simple to release any version and we are not going to rely on the what kind of Java version is installed on the sytem.

plata:
I'm completely fine with that approach. You don't want to do this for alpha-2 already, do you?

qparis:
> I'm completely fine with that approach. You don't want to do this for alpha-2 already, do you?

Not necessarely if we find a way to release alpha 2 properly and reliably :)

plata:
I think it is valid to simply state that alpha-2 has some requirements regarding the system Java version (e.g. the .deb is made for Ubuntu 18.04 and the .dmg needs Java 8). It's just the same as for the first alpha. I would like to have a JDK 10/11 .zip so I can make a flatpak manifest for flathub. If you have issues building the flatpak, well it's not there. I don't think that's a show stopper.

qparis:
Well the package are here: https://repository.playonlinux.com/PlayOnLinux/5.0.0-alpha2/

plata:
If you add the .dmg this should be enough. Maybe we could have 2 .zip (Java 8+10/11) but otherwise...

qparis:
> If you add the .dmg this should be enough. Maybe we could have 2 .zip (Java 8+10/11) but otherwise...

The dmg is not fully operational.... We need to bundle unix tools (cabextract, etc...)

plata:
Ok. Then we will have it for the next release.

plata:
Yep, just tried jlink as well and immediately got errors because of automatic modules. I can't remember why I implemented the modules originally. Might be something regarding OpenJFX. Or only during Java 10/11 migration.

plata:
Spring seems to go rather toward GraalVM native image than jlink (see https://github.com/spring-projects/spring-boot/issues/13799).

qparis:
> Spring seems to go rather towards GraalVM native image than jlink (see [spring-projects/spring-boot#13799](https://github.com/spring-projects/spring-boot/issues/13799)).

Seems to be more promising indeed. However, still no solution to run JavaFX with GraalVM

plata:
Maybe we just have to wait until there is more progress. I don't see any solution currently.

qparis:
> Maybe we just have to wait until there is more progress. I don't see any solution currently.

Or we need to drastically reduce the dependencies of phoenicis maybe.

plata:
Towards spring?

qparis:
Maybe yes. Other modules are causing problems: 
- bouncycastle
- common.langs
- loggers

plata:
Uff. I'm not sure. There's really no other way to get it running for OSX? I mean supporting different Java versions is not optimal but I think it works for Linux.

qparis:
It is just a pain we are going to have troubles when we will have to debug users. 

What I don't like here is that we are losing the **simple** aspect of Java by having multiple profiles. And things are just going to get worst with all GraalVM cases

plata:
So what's the alternative? Bundle the JRE?

qparis:
500mo without jlink

qparis:
What can also be done is to recreate module-info to automatic modules thanks to jdeps. I have a proof of concept that works for the main part of our dependencies. Unfortunately, some of them calls some classes that we do not use and we do not bundle.

So I created another tool that creates a "Dummy" version of theses classes. But it becomes weird and it does not work for commons.lang that use a package name with the keyword "enum"

qparis:
Ok, I think I can create a package if I remove all the module-info

madoar:
I'm currently still using Java 8 on my Desktop PC.

About the modules:
We're using modules because of OpenJFX 11.
Without modules we didn't get it to working by providing openjfx only via maven.
In addition I believe that modules are a part of the future of Java, therefore I think we shouldn't remove them.

About jpackager and jlink:
As far as I know jpackager will use jlink internally to create an image.
Therefore I don't think we can get jpackager working without getting jlink working.

About javapackage vs. jpackage
About which of the both tools are we talking?
`javapackager` is the JavaFX tool, which has been provided with JavaFX previously.
`jpackage` is a new tool, which is still under development and is likely to be included in [OpenJDK 13](https://bugs.openjdk.java.net/browse/JDK-8212780), see also [here](http://openjdk.java.net/jeps/343).

qparis:
> I'm currently still using Java 8 on my Desktop PC.
> 
> About the modules:
> We're using modules because of OpenJFX 11.
> Without modules we didn't get it to working by providing openjfx only via maven.
> In addition I believe that modules are a part of the future of Java, therefore I think we shouldn't remove them.

None of our dependencies use modules so far, so it is really a pain..

> About jpackager and jlink:
> As far as I know jpackager will use jlink internally to create an image.
> Therefore I don't think we can get jpackager working without getting jlink working.

I think I can, but only if we don't use module. (In fact I have successfully created a .dmg, I just have to fix some javax dependencies problem) 

> 
> About javapackage vs. jpackage
> About which of the two tools are we talking?
> `javapackager` is the JavaFX tool, which has been provided with JavaFX previously.
> `jpackage` is a new tool, which is still under development and is likely to be included in [OpenJDK 13](https://bugs.openjdk.java.net/browse/JDK-8212780), see also [here](http://openjdk.java.net/jeps/343).

Reading the doc, it looks like the new tool

qparis:
Yeah! I got POM running with Java 11 and jpackager 

Merge request upcoming

plata:
We can make a profile for the modules (they are not used for Java 8 either). I also wouldn't like to loose them completely.
Curious how you got OpenJFX working.

qparis:
> We can make a profile for the modules (they are not used for Java 8 either). I also wouldn't like to loose them completely.

It causes issue when I use them with IntellIJ. How do you execute POL locally? 

> Curious how you got OpenJFX working.

Few minutes and the PR is ready

plata:
I use IntelliJ but with a customized JavaFX application containing `--module-path` like the phoenicis.sh. If you do not activate the profile by default, it shouldn't bother you at all.

plata:
Btw I think using jpackager and removing Java 8 support are two different topics which should not be mixed up.

qparis:
@plata @madoar 
With this current merge request: 

* Java 8 is unsupported 
* Modules are removed 
* OpenJDK 10+ is required to compile locally
* OpenJDK 11+ is required to generate package (jpackage needs major version 55+) 

So that was about the drawbacks. Now, let's talk about the advantages: 
* No longer any profile on maven files. It builds "as it" 
* We have a script phoenicis-web-package.sh that will: 
   * Generate a **lightweight and fully standalone** (no Java required) .app for macOS : 150Mo as it, 75Mo if we manage to get rid of java.sql (see #1639)
   * Generate a **lightweight and fully standalone** (no Java required) .deb for ubuntu: 76Mo as it  
   * We could generate a .tar.gz or a .zip containing a native binary 
   * We could generate a rpm 

* During developement process, the can run easily with mvn exec:java. I did not manage to run it directly through intelliJ but I'm sure it is possible to do it if we set correctly the add-modules and the module-path param

qparis:
> Btw I think using jpackager and removing Java 8 support are two different topics which should not be mixed up.

The two topics are bound: 
 - If we remove Java 8, we must provide an alternative solution for Ubuntu => jpackage 
 - If we decide to implement jpackage, it has a lot of impact on how the package are made and frankly, the ticket is hard enough to do :)

qparis:
@plata if you want, we can keep modules. But they are not needed and it will require more work to maintain these files

.travis.yml
```diff
@@ -3,7 +3,8 @@ language: java
 # avoid double builds on pull requests
 branches:
   only:
-    master
+    - master
+    - release/5.0-alpha-2
```
plata: You branched from the alpha-2 release branch but the PR is towards master. That doesn't match.

phoenicis-dist/src/scripts/phoenicis-create-package.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+## Dependencies
+# Linux:
+# - fakeroot
+#
+
+VERSION="$1"
```
plata: Can't we do this in maven?

plata:
I cannot run this. `mvn exec:java` gives
```
The parameters 'mainClass' for goal org.codehaus.mojo:exec-maven-plugin:1.6.0:java are missing or invalid
```

Regarding modules: It still uses modules otherwise you would not need this command line parameters. I thing you just have everything in the unnamed module now.

.travis.yml
qparis: I just need to remove this line, is it correct?

phoenicis-dist/src/scripts/phoenicis-create-package.sh
qparis: Calling the script from maven, maybe. Doing what it does, not easily except if we create a plugin (which would be a cool open source contribution though)

.travis.yml
plata: No, there are more places if you check the diff.

phoenicis-dist/src/scripts/phoenicis-create-package.sh
plata: Is it because we basically have 2 packaging stages (jpackager and then deb/dmg etc)?

qparis:
> I cannot run this. `mvn exec:java` gives
> 
> ```
> The parameters 'mainClass' for goal org.codehaus.mojo:exec-maven-plugin:1.6.0:java are missing or invalid
> ```
> 

Are you running it from javafx module? 

> Regarding modules: It still uses modules otherwise you would not need this command line parameters. I thing you just have everything in the unnamed module now.

Not really. It is different here because our main class is no longer called from the modulepath but from the classpath. The differences is that when we add these switch to the command lines, we just ask the JVM to add these modules to the classpath (if I understand well).

On the contrary, if phoenicis uses module, then we need to import legacy libraries as if they were modular. The JVM automatically gives them a module name based on the name of the .jar archive. This is what is not supported by jlink

.travis.yml
qparis: I've already changed pom.xml files

plata:
I was running from top level. In phoenicis-javafx I get:
```
[ERROR] Failed to execute goal on project phoenicis-javafx: Could not resolve dependencies for project org.phoenicis:phoenicis-javafx:jar:5.0-SNAPSHOT: The following artifacts could not be resolved: org.phoenicis:phoenicis-configuration:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-repository:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-scripts:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-library:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-engines:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-containers:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-settings:jar:5.0-SNAPSHOT: Could not find artifact org.phoenicis:phoenicis-configuration:jar:5.0-SNAPSHOT
```

.travis.yml
qparis: Ok done

phoenicis-dist/src/scripts/phoenicis-create-package.sh
qparis: jpackager does pretty much everything for Linux including generating a .deb archive, except that it does not install scripts in /usr/bin. So far, everything is installed in /opt. 

This is not ideal but good enough for testing the concept. If we find out that the approach are good for you, I'll improve the .deb structure so that it it matches POLv4 one

qparis: Concerning .osx, I did not find how to put the .app in a dmg but this can also be easily scripted

qparis: Finally, there are no maven plugin for jpackager afaik

qparis:
> I was running from top level. In phoenicis-javafx I get:
> 
> ```
> [ERROR] Failed to execute goal on project phoenicis-javafx: Could not resolve dependencies for project org.phoenicis:phoenicis-javafx:jar:5.0-SNAPSHOT: The following artifacts could not be resolved: org.phoenicis:phoenicis-configuration:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-repository:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-scripts:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-library:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-engines:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-containers:jar:5.0-SNAPSHOT, org.phoenicis:phoenicis-settings:jar:5.0-SNAPSHOT: Could not find artifact org.phoenicis:phoenicis-configuration:jar:5.0-SNAPSHOT
> ```

Ok, try maven install from the top; then run it from javafx. This is not ideal, but you can still run it from IntelliJ anyway

plata:
That's working. Please update the respective documentation in:
- https://github.com/PhoenicisOrg/phoenicis/blob/master/docs/_docs/Developers/intellij-idea.md
- https://github.com/PhoenicisOrg/phoenicis/blob/master/docs/_docs/Users/build.md
- https://github.com/PhoenicisOrg/phoenicis/blob/master/docs/_docs/Users/installation.md
- https://github.com/PhoenicisOrg/phoenicis/blob/master/docs/_docs/Users/packages.md

Do we even need phoenicis-dist now?

phoenicis-dist/src/scripts/phoenicis-create-package.sh
plata: I'm fine with separating the build and the packaging but there should be a general concept. For example, we now have information about the .deb in the control file and also in the script. This doesn't make a lot of sense to me.

qparis:
> That's working. Please update the respective documentation in:
> 
>     * https://github.com/PhoenicisOrg/phoenicis/blob/master/docs/_docs/Developers/intellij-idea.md
> 
>     * https://github.com/PhoenicisOrg/phoenicis/blob/master/docs/_docs/Users/build.md
> 
>     * https://github.com/PhoenicisOrg/phoenicis/blob/master/docs/_docs/Users/installation.md
> 
>     * https://github.com/PhoenicisOrg/phoenicis/blob/master/docs/_docs/Users/packages.md
> 
> 
> Do we even need phoenicis-dist now?

Good! Have you tried the packaging script?

qparis:
Yes, I’ll do it

phoenicis-dist/src/deb/control/control
```diff
@@ -1,5 +1,5 @@
 Package: phoenicis-playonlinux
-Version: 5.0
+Version: 5.0-alpha-2
```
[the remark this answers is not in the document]
qparis: This file will be removed anyway

plata:
Script doesn't work for me:
```
Module javafx.base does not exist.
Module javafx.graphics does not exist.
Module javafx.controls does not exist.
Module javafx.media does not exist.
Bundler DEB Installer skipped because of a configuration problem: Can not find dpkg-deb.  
Advice to fix: Please install required packages.
```

Info: I'm on Manjaro so building a deb doesn't make much sense.

qparis:
Yes, so this is normal that you could not build the deb without having dpkg-deb.
However, if you look at the packages directory inside target, you should see a standalone phoenicis

plata:
I have two folders: "Phoenicis PlayOnLinux" and "PhoenicisPlayOnLinux" is that intended?

qparis:
That’s a jpackage bug. I think we are going to fix this by changing the name

plata:
How do we handle flatpak now? I need an archive containing the jar and /lib. Even better also the other files (icons, .desktop etc). Can we maybe use maven archiver to do this?

I think we should remove jdeb. Maintaining two different ways to build a .deb is not a good idea.

I still would like to have a profile to build proper modules (doesn't have to be active by default).

qparis:
Dev is already removed from Maven

For the modules, can we re-add them in another merge request? Because there are a lot of things to check (That it does not break IntelliJ)

Concerning flatpak, it should not have changed at all

plata:
I don't think that the modules will make any issue for IntelliJ. Simply use
```
<excludes>
    <exclude>module-info.java</exclude>
</excludes>
```
We can add the profile later on.

plata:
For the flatpak: it doesn't work if you try to build it according to https://phoenicisorg.github.io/phoenicis/Users/packages/.

qparis:
> I don't think that the modules will make any issue for IntelliJ. Simply use
> 
> ```
> <excludes>
>     <exclude>module-info.java</exclude>
> </excludes>
> ```
> 
> We can add the profile later on.

This causes problems on my setup:
capture d ecran 2018-12-27 a 12 01 56

plata:
Ah now I understand. It's because IntelliJ doesn't really use the Maven build internally. Ok then we leave it.

qparis:
There are obviously some problem with Flatpak. I'm wondering if we should not directly bundle the jpackage generated tree

qparis:
@plata If you are ok with that, we can merge this base and while I can work on improving the .deb package and .dmg, maybe you can help concerning the flatpak. 
I think we can target alpha-2

plata:
I don't think that's a good idea. It creates a native bundle so to do that properly, you would have to run jpackager during the flatpak build.
Ok, we can probably do that. I think it's sufficient to merge master into 5.0-alpha-2 after merging this. Can you do that?

i18n/keys.pot
```diff
@@ -17,229 +17,163 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 
-#: phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java:66
```
plata: Can you remove this file from the commit? I'm not sure if it will conflict with Crowdin after the merge (and it's not related).

phoenicis-dist/pom.xml
```diff
@@ -57,7 +57,7 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
+            <!-- plugin>
```
plata: We should remove this.

phoenicis-dist/src/deb/control/control
```diff
@@ -1,5 +1,5 @@
 Package: phoenicis-playonlinux
-Version: 5.0
+Version: 5.0-SNAPSHOT
```
plata: We should remove the deb folder. Not a good idea to have two separate ways to build a deb.

qparis:
Yep

plata:
For flatpak: It works if you change the command in the manifest to 
```
"commands": [
    "/app/jre/bin/java --module-path /app/phoenicis/lib/ --add-modules=ALL-MODULE-PATH org.phoenicis.javafx.JavaFXApplication \"$@\""
]
```

plata:
What about the `phoenicis.sh` and `phoenicis-java8.sh`? Also the xmls in `assembly`?

qparis:
I’m going to remove these scripts

plata:
If you do that and fix the line for flatpak, I think I can approve the PR.

phoenicis-dist/pom.xml
qparis: Fixed

i18n/keys.pot
qparis: Done

qparis:
> If you do that and fix the line for flatpak, I think I can approve the PR.

Done

plata:
What about the assembly xmls?

qparis:
I don't remember what there roles are

plata:
They are used by the maven assembly plugin to build the zip. As you removed the plugin, you should be able to delete the xmls as well.

qparis:
Done. In the next merge request, I'll try to create a single bash script that does everything

plata:
Just give me 5mins after this is merged. I have fixed the flatpak build.

qparis:
Ok. Do you confirm that the standalone app works for you?

plata:
Yes, it was working.

plata:
You can merge.

madoar:
I think it was an overhasty decision to merge this PR so early.
I believe we should have given this PR at least a few more days to talk about the advantages and disadvantages of the suggested changes, especially because the changes target the build behavior and the installation behavior of our application.

The suggested changes affect nearly all people who are associated with Phoenicis:
- the developers: because they now need to use Java 11 to work with and compile Phoenicis (to this group I belong)
- the users: because they now are forced to download a larger installer and to install an eventual additional JRE, which can largely increase the installation size.

In addition this PR targets changes to the module system of Phoenicis, which again are huge changes that target the future development of the application.
I would have liked it, if we put more effort into making the suggested build process work without removing the module system support.

I believe that these changes should be incubated a bit longer before we decide whether to merge them or not.

madoar:
Another huge point is, that we're using an unofficial build of `jpackage`, which is still under development, which could be quite buggy with strange side effects.

PR #1637, opened 2018-12-26
First steps

plata:
Note that the PR is not towards master but a feature branch so that we can work together and share our progress whenever someone figures anything out.

Next topic that must be tackled is the "include" function.

### DIFF
--- a/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutRunner.java
+++ b/phoenicis-library/src/main/java/org/phoenicis/library/ShortcutRunner.java
@@ -21,7 +21,11 @@ package org.phoenicis.library;
 import org.phoenicis.library.dto.ShortcutDTO;
 import org.phoenicis.scripts.interpreter.InteractiveScriptSession;
 import org.phoenicis.scripts.interpreter.ScriptInterpreter;
-import jdk.nashorn.api.scripting.ScriptObjectMirror;
+
+import javax.script.Invocable;
+import javax.script.ScriptEngine;
+import javax.script.ScriptEngineManager;
+import javax.script.ScriptException;
 
 import java.util.List;
 import java.util.function.Consumer;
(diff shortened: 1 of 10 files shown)


qparis:
I think we need to fix #1635 first

plata:
Why?

qparis:
> Why?

Because I suspect that the way you'll implement graalVM will be way different from Java 8 to Java 11, and it seems that we are going to complexify the deployement process again

qparis:
Maybe we should use a new branch for these two topics.

plata:
Probably polyglot but scripting might be good for transition.

madoar:
@qparis I wouldn't understand this PR as a try to change to GraalVM.
I understand it as a try to replace Nashorn with GraalJS.
This decision is independent of the decision to change to GraalVM.

@plata this is fine by me.
I think the biggest changes will be the move from `ScriptObjectMirror` (Nashorn) to `Value` (GraalJS).

plata:
Yes, this `ScriptObjectMirror` is giving me headaches. Therefore I wanted to share this first steps so you guys can also have a look.

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutRunner.java
```diff
@@ -42,11 +46,20 @@ public void run(String shortcutName, List<String> arguments, Consumer<Exception>
...
 
+        ScriptEngineManager m = new ScriptEngineManager();
+        ScriptEngine engine = m.getEngineByName("graal.js");
+        Invocable inv = (Invocable) engine;
         interactiveScriptSession.eval("include([\"engines\", \"wine\", \"shortcuts\", \"reader\"]);",
                 ignored -> interactiveScriptSession.eval("new ShortcutReader()", output -> {
-                    final ScriptObjectMirror shortcutReader = (ScriptObjectMirror) output;
-                    shortcutReader.callMember("of", shortcutDTO);
-                    shortcutReader.callMember("run", arguments);
+                    final Object shortcutReader = (Object) output;
```
madoar: My guess is, that when you're using `graal.js` you'll receive a `Value` object here.
Have you checked?
If this is the case you can use the more comfortable methods of the `Value` class, see: https://www.graalvm.org/sdk/javadoc/org/graalvm/polyglot/Value.html

madoar: What issues did you have?
For the include:
Is it possible to replace `ScriptObjectMirror` with `Value`?
The `Value` class has a method `as(Class<T>)` which returns an object of type `T` can we use this?

plata: I have to try. Actually this is exactly why I want to have the branch on upstream: that you can try yourself.

plata:
I will merge this now. I don't really have an idea currently how to replace `ScriptObjectMirror`. So if anyone of you can make any progress with Graal, please open you PR vs that branch (even if it doesn't fully work yet).

PR #1646, opened 2018-12-27
Build zip for flatpak

plata:


qparis:
Why is there still a JDK 8 on the build?

PR #1649, opened 2018-12-27
[Backup] Module infos

qparis:


ImperatorS79:
You should add 
requires org.phoenicis.lnk; in tools
requires org.phoenicis.configuration; in multithreading
requires javafx.collections; in javafx

or it will not compile.

plata:
As we specify the modules on the command line now, do you think it would make sense to go for Java modules again?

qparis:
Not as long as our dependencies do not support modules IMHO

plata:
Ok, so probably not before Spring 6.

PR #1651, opened 2018-12-27
Travis: fix .deb build

plata:


qparis:
Seems better indeed.

plata:
In the end, though, if the build doesn't fail if it doesn't work, the check doesn't make much sense.

qparis:
The script will later be improved

PR #1654, opened 2018-12-27
Fixed deb package

qparis:
This MR does two things:
* It provides nashorn engine to the packages. Without this, packaged version of phoenicis cannot run any script
* It removes extra dependencies so that we ensure that Phoenicis will install on debian


plata:
When this is merged, just cherry pick the commit on master to the release branch with
```
git cherry-pick <checksum>
```

PR #1668, opened 2018-12-29
Update flatpak

plata:
- use name `org.phoenicis.playonlinux`
- update appdata.xml according to flathub requirements

### DIFF
--- a/docs/_docs/Users/packages.md
+++ b/docs/_docs/Users/packages.md
@@ -18,10 +18,10 @@ Required extensions:
 
 ```
 cd phoenicis-dist/src/flatpak/
-flatpak-builder build-dir org.phoenicis.javafx.json --force-clean --user --install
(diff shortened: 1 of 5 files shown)


plata:
@qparis shall I merge this as is or do you want to upload the screenshot on our server?

qparis:
What do you think is best?

plata:
I don't know. It might be perceived as cleaner on our server.

plata:
But if it's too much work, I can just use the GitHub image. I don't want to distract you from the deb.

qparis:
We need to think someday about how you can have access to the servers, or at least some of them

plata:
We can think about that. So I use the GitHub image for now?

qparis:
If you can change it easily later, yes

qparis:
I have some blocker bugs I need to fix on macOS

plata:
Don't worry. Shouldn't be a problem to update it later.

PR #1669, opened 2018-12-29
Refactor the MappedList class

madoar:
This PR refactors the `MappedList` class and makes the `mapper` function changeable at runtime.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/collections/MappedList.java
```diff
@@ -94,18 +195,37 @@ protected void update(Change<? extends F> c) {
...
+        return mapper.get();
+    }
+
+    public ObjectProperty<Function<? super F, ? extends E>> mapperProperty() {
+        return mapper;
+    }
+
+    public void setMapper(Function<? super F, ? extends E> mapper) {
+        this.mapper.set(mapper);
+    }
```
plata: add newline at EOF

madoar: Done

PR #1676, opened 2018-12-30
Update translations with Travis and not during build

### LINKED ISSUE #1675: Do not update translation file during build
Use Travis to update the translations (the generated .pot is different on every computer/build). See scripts repo for how this can be done.

plata:
fixes #1675 

Note: I couldn't really test if the Travis part works.


.travis.yml
```diff
@@ -79,3 +88,12 @@ jobs:
       script:
         - bundle exec jekyll build
         - bundle exec htmlproofer ./_site --only-4xx --check-favicon --check-html --assume-extension
+
+env:
+  global:
+  # to create secure for travis-ci.com
```
madoar: I don't really understand the comments.
Why repository should be cloned? (ours or a specific one from travis.org)
Where do we find the application `travis`?

i18n/push_translations.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+git config --global user.email "travis@travis-ci.org"
+git config --global user.name "Travis CI"
+
+head_ref=$(git rev-parse HEAD)
```
madoar: Some comments explaining the `if` cases would be nice.

.travis.yml
plata: Just the same comments as for scripts. Our repository must be cloned. https://github.com/travis-ci/travis.rb

i18n/push_translations.sh
plata: I couldn't really add more than in the `echo` below (I only copied this from some tutorial).

PR #1677, opened 2018-12-30
Few exception fixes

qparis:


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/common/widgets/icons/skin/IconsListWidgetSkin.java
```diff
@@ -98,9 +98,11 @@ private void updateOldSelection(ListWidgetSelection<E> oldSelection) {
         Optional.ofNullable(oldSelection).map(ListWidgetSelection::getSelection).ifPresent(selection -> {
             final int oldValueIndex = getControl().getElements().indexOf(selection);
 
-            IconsListElement<E> oldElement = mappedElements.get(oldValueIndex);
+            if (mappedElements.size() > oldValueIndex && oldValueIndex != -1) {
```
madoar: Why do we need this?
It shouldn't be possible for this case not to happen.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
```diff
@@ -146,7 +146,7 @@ public EnginesController(EnginesView enginesView, RepositoryManager repositoryMa
                 this.repositoryManager.addCallbacks(
                         repositoryDTO -> this.enginesManager.fetchAvailableEngines(
                                 repositoryDTO,
-                                engines -> this.populateView(repositoryDTO, engines),
+                                engines -> Platform.runLater(() -> this.populateView(repositoryDTO, engines)),
```
madoar: Why is this necessary?
I mean `populateView` uses `Platform.runLater` internally therefore it shouldn't be required here

qparis: Hum, not all the content of populateView. showWait() for example

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/common/widgets/icons/skin/IconsListWidgetSkin.java
qparis: In some cases, it happens... I don't know why

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
madoar: Yes, but `showWait` shouldn't require `Platform.runLater`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/common/widgets/icons/skin/IconsListWidgetSkin.java
madoar: I think it is better to locate why this happens.
Like I mentioned the "contract" for `selection` is that it should be always contained inside `getControl().getElements()`.

plata:
The diff is strange.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
madoar: Hmmm, interesting.
Then I would suggest to add a `Platform.runLater` around `showWait()`.
This in turn would make it unnecessary to add `Platform.runLater` around `populateView` and `forceViewUpdate`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/common/widgets/icons/skin/IconsListWidgetSkin.java
qparis: The bugs appear randomly so it's hard to reproduce. It think it is because oldValueIndex is sometimes equals to -1

madoar: The question in my opinion is, why is this the case?
I mean it shouldn't be possible.

qparis: What if there are no selections?

madoar: Then the `.ifPresent(selection -> {...})` in line 98 won't be called

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/engines/EnginesController.java
qparis: Done

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/common/widgets/icons/skin/IconsListWidgetSkin.java
qparis: Probably a race condition

madoar: I don't understand how.
This code should be executed by the JavaFX GUI thread.

qparis: User interaction maybe

madoar: Again I don't understand how.
All user interaction should be processed on the, according to my knowledge, single JavaFX GUI thread.
This in turn should make it impossible for race conditions to occur.

qparis: capture d ecran 2018-12-30 a 21 04 53

qparis: capture d ecran 2018-12-30 a 21 06 15

qparis: In fact, it happens when you select another category

madoar: Did you set any filter settings?
Which applications did the list show?
Do you know the steps to reproduce this?

qparis: It happens when you select another category

madoar: Ok, the issue is, that a change to the `ListWidgetElement` list doesn't set the selection to null.
The problem here is, that setting the selection to null will still lead to the exception because the `ChangeListener` is triggered.

The only solutions seems to be to add an `if (mappedElements.size() > oldValueIndex && oldValueIndex != -1) {` to the `updateOldSelection()` method like you suggested.
Except that it needs to be added to all three list widget skins.

Still I'm unsure if there isn't a more elegant solution for the problem.

qparis: I see. Where are the others?

madoar: - `DetailsListWidgetSkin`
- `CompactListWidgetSkin`
- `IconsListWidgetSkin`

madoar: Maybe also add a comment describing why the `if` is needed, otherwise it may be accidentally removed later on

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/common/widgets/details/skin/DetailsListWidgetSkin.java
```diff
@@ -100,7 +100,10 @@ private void updateOldSelection(ListView<DetailsListElement<E>> container, ListW
         Optional.ofNullable(oldSelection).map(ListWidgetSelection::getSelection).ifPresent(selection -> {
             final int oldValueIndex = getControl().getElements().indexOf(selection);
 
-            container.getSelectionModel().clearSelection(oldValueIndex);
+            // In some corner cases (category / filter changes, the current selection can disappear)
+            if (oldValueIndex != -1) {
```
madoar: I think all three skins require both `mappedElements.size() > oldValueIndex && oldValueIndex != -1`.
I think `container.getSelectionModel().clearSelection(oldValueIndex)` will throw a `NullPointerException` if the index is larger than the number of shown elements.

qparis: In fact I'm not totally sure. The code is the the same

madoar: Which code is the same?

qparis: is not*

madoar: I'm not sure I understand.
The main difference is that the `DetailsListWidget` and the `CompactListWidget` delegate the deselection to the internal `SelectionModel`, while the `IconsListWidget` provides its own logic for the selection and deselection of items.
Both approaches require a valid `oldValueIndex` which is in the interval `[0, mappedElements.size()]` to work correctly.

PR #1684, opened 2018-12-31
Add javafx web to package

qparis:


madoar:
`javafx.base` is not needed?

plata:
@qparis can be merged.

PR #1691, opened 2019-01-01
Fix: Travis does not detect packaging issues

plata:
A failure in `after_script` doesn't fail the build.

> If you rely on the fact that a failure in the after_script phase fails the entire test, you should move such commands to the script phase.

https://blog.travis-ci.com/after_script_behavior_changes/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,8 @@ jobs:
         - sudo add-apt-repository ppa:alexlarsson/flatpak -y
         - sudo apt-get update -q
         - sudo apt-get install -y flatpak flatpak-builder
-      after_script:
+      script:
+        - mvn clean package -DskipTests # no need to run tests again
         - flatpak --user remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -67,14 +69,16 @@ jobs:
...
+        - mvn clean package -DskipTests # no need to run tests again
         - cd phoenicis-dist/src/scripts
         - bash phoenicis-create-package.sh
     - stage: Packages
       name: dmg
       os: osx
       osx_image: xcode10.1
-      after_script:
+      script:
+        - mvn clean package # run tests for OSX (doesn't run in 'build + test')
```
madoar: Why do you not sue `-DskipTests` here?

plata: Read the comment.

madoar: Why don't we run it in `build + test`?
Does it throw some kind of error?

plata: Yes, logically it makes sense but it would run the Maven build twice. Maybe open a PR with this then we can decide there.

madoar: Yes, the same as we do for our linux builds :)

plata: yep (it was like this before for OSX as well)

PR #1692, opened 2019-01-01
Update release process documentation

plata:


plata:
@qparis please check if this matches what has to be done.

qparis:
Really harder for these reasons:
* It needs hdiutil tool available only on macOS (so impossible to run on our own servers) 
* .app must be signed before the dmg is built (I think this is the hardest part) 
* compiled binaries need to be added inside the package.

plata:
Do you want to detail the .dmg build? Feel free to edit the .md directly.

qparis:
I can but the _compiled binaries need to be added inside the package._ is really artisanal :-D

plata:
Maybe we should better have a separate documentation page for that. It will probably overload this page.

PR #1697, opened 2019-01-01
File Utilities : Temporary directory creation

qparis:


qparis:
This method has no parameters, so we have nothing else to do

qparis:
@plata Same here, can we merge?

Issue #1694, opened 2019-01-01
Feedback: Build is failing


Kreyren:
https://pastebin.com/34KPmere - Release is failing
https://pastebin.com/HLN5qeW6 - Master is failing

Using `dev-java/maven-bin-3.6.0`

https://paste.pound-python.org/show/JCL0XAiAje79S3Tp9gjT/ - Release is failing 

Using `dev-java/maven-bin-3.3.3-r1`

*Give up*

Recommends adding expected maven version in readme.

plata:
I'm using Maven 3.6.0 so that should be fine. Which JDK version do you use (you need at least 10)?

Kreyren:
@plata `dev-java/maven-bin-3.6.0` fails on release.

https://paste.pound-python.org/show/xmoPvP7mq7urfqSxKWab/

Kreyren:
Noticed #1695 redownloading source

Kreyren:
Redownloaded master, fails on `dev-java/maven-bin3.6.0`

https://paste.pound-python.org/show/pgSaH7Mgg6gqMsbb15WI/

qparis:
does the directory i18n exists?

Kreyren:
@qparis it does https://paste.pound-python.org/show/RwT7Ac41VmBa6fnK65yY/

madoar:
Can you please run `mvn clean package -X`?
This will add debug output

Kreyren:
kreyren@dreamon ~/phoenicis $ wgetpaste -c "mvn clean packge -X"
https://paste.pound-python.org/show/UYW5tPxL5NzUDbZj3qLj/

Kreyren:
List of installed packages: https://paste.pound-python.org/show/UEj3tdPWR6Ie3jLrb2xn/

qparis:
You mispelled package

madoar:
You have a typing error in the command, it must be `package` not `packge`

Kreyren:
sorry new debug: https://paste.pound-python.org/show/buf9WgVQUUahNJJ813Zc/

redownloading master

Kreyren:
debug on redownloaded https://paste.pound-python.org/show/lvkyxgQRZ3uphPEeAfEw/

qparis:
Do you have the permission to write on keys.pot?

Kreyren:
@qparis permission to write where? Is invoking as root sane?

madoar:
[...]

PR #1708, opened 2019-01-02
Update README.md

Kreyren:
Fixes: https://github.com/PhoenicisOrg/phoenicis/issues/1694#issuecomment-450971639
Added dependencies for linux
Updated build instructions for Linux
Signed-off-by: Jacob Hrbek <werifgx@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,15 +16,19 @@ Supported engines:
 
 ## Build and Run
 ### Linux
+Dependencies:
+- Java 10+ (OpenJDK recommended)
+- Maven
+
 Build
 ```
 mvn clean package
 ```
 Run
 ```
-cd phoenicis-dist/target
-unzip phoenicis-dist.zip
-./phoenicis-dist/phoenicis.sh
+mvn install
+cd phoenicis-javafx/
+mvn exec:java
 ```
 
 ### OSX


README.md
```diff
@@ -16,15 +16,19 @@ Supported engines:
 
 ## Build and Run
 ### Linux
+Dependencies:
```
plata: I think do not need the Linux/OSX separation anymore, you can simply remove the OSX part. Remove "(OpenJDK recommended)".

plata:
@qparis is it correct that `exec:java` will also work on OSX?

README.md
Kreyren: (java noob) : Seems that OS X is using different commands and it's usually a good idea to separate OSX and Linux since they are very different i believe

Kreyren: Ah i see what you mean readme points to https://phoenicisorg.github.io/phoenicis/ now.. aldo i'm unable to find dependencies on it?

qparis:
> @qparis is it correct that `exec:java` will also work on OSX?

Yep

README.md
plata: I will do the update.

PR #1740, opened 2019-01-06
Replace the `AdhocList` and `ExpandedList` classes with `MappedList` and `ConcatenatedList`

madoar:
- add ? extends to the inner generics of `ConcatenatedList`
- change all usages of `AdhocList` to `ConcatenatedList`
- change all usages of `ExpandedList` to a combination of `MappedList` and `ConcatenatedList`
- remove the now unused `AdhocList` and `AdhocListTest`
- remove the now unused `ExpandedList` and `ExpandedListTest`


madoar:
@qparis, @plata please be sure to give this PR in-depth test, because I haven't used the `ConcatenatedList` before. It would be sad if its usage breaks something.

plata:
What can we test? Sidebar buttons?

madoar:
Basically everything that has been changed by me:
- all sidebars (also in combination with filters)
- the content of the shown list widgets

PR #1759, opened 2019-01-10
Flatpak documentation: update GL32.nvidia version

plata:


docs/_docs/Developers/packages.md
```diff
@@ -14,7 +14,7 @@ The built packages are located in the target/packages folder:
 ### Flatpak
 Required extensions:
 - org.freedesktop.Platform.Compat.i386
-- org.freedesktop.Platform.GL32.nvidia-410-73
+- org.freedesktop.Platform.GL32.nvidia-415-25
```
qparis: What happens for people that have an AMD card?

plata: I don't know. I don't own one so I cannot try.

PR #1765, opened 2019-01-13
Fix bug #1764:  Downloader does not follow 302 and 301 code when a locale file destination is explicitly set

qparis:
@ImperatorS79 Should fix your problem with dxvk


ImperatorS79:
Working for me.

plata:
Add "fixes #1764" to the PR description so that the bug gets closed.

qparis:
Theoritically yes if MockHTTPServer allows it

PR #1772, opened 2019-01-14
Move JavaFX collections to new repository

madoar:
This PR externalizes our JavaFX collection classes to https://github.com/PhoenicisOrg/javafx-collections

This PR is still work in progress!


madoar:
When I try to run Phoenicis built using this PR I get the following error:

```
Error occurred during initialization of boot layer
java.lang.module.FindException: Unable to derive module descriptor for phoenicis-dist/target/lib/javafx-collections-09417a8e6f.jar
Caused by: java.lang.IllegalArgumentException: javafx.collections.09417a8e6f: Invalid module name: '09417a8e6f' is not a Java identifier
```

@qparis, @plata any ideas?

qparis:
> When I try to run Phoenicis from IntelliJ (built using this PR) I get the following error:
> 
> ```
> Error occurred during initialization of boot layer
> java.lang.module.FindException: Unable to derive module descriptor for phoenicis-dist/target/lib/javafx-collections-09417a8e6f.jar
> Caused by: java.lang.IllegalArgumentException: javafx.collections.09417a8e6f: Invalid module name: '09417a8e6f' is not a Java identifier
> ```
> 
> @qparis, @plata any ideas?

Yes. When using automatic modules, the name of the .jar is used to determine the version. 
Can you add a module-info file to javafx-collections?

madoar:
@qparis thanks! It seems to be working now :)
Can you test this so that we ensure that it works for you too.

Before merging this I want to specify a better version number/name for our new `javafx-collections` dependency. In addition the `README` file of `javafx-collections` needs some improvement.

qparis:
I will test it. 

I'm not sure about the name of the module: "javafx-collections"

madoar:
I'm open for suggestions. I just think that we shouldn't add `phoenicis` in the artifact name, because this limits the field of application for the library in my opinion.

qparis:
It is just the javafx is reserved for ... javafx :)

madoar:
I don't think this is a problem. I mean the official javafx libraries can be identified by their group id (`org.openjfx`), which we don't use. In addition the `javafx-collections` library contains what the name suggests, additional collections for JavaFX, which makes it the most fitting name in my opinion.

qparis:
I mean, if they create a module javafx-collections, it will collide with our

madoar:
Any suggestions for a better name?
I thought about `additional-javafx-collections`, but I think `additional` is too long and clunky

madoar:
We also shouldn't forget, that jitpack uses the repository name as the `artifactId`. I'm not sure whether we want to rename the repository.

qparis:
Why not phoenicis.javafx.collections? 
Phoenicis is an organization, not a product

madoar:
My issue with prepending `phoenicis` to the artifact id is, that I associate `phoenicis` with the application and its ecosystem. More specific the problem is, that `javafx-collections` is not really a part of this ecosystem, it only contains some classes which are used by phoenicis but are not necessarily part of phoenicis.

It is hard for me to explain...

qparis:
Ok so let’s keep your suggestion for the moment. We can change it later if required anyway

madoar:
Ok, this PR is now ready to be merged from my side :)

PR #1774, opened 2019-01-15
Refactor the `AboutPanel` with MVC in mind

madoar:
This PR refactors the `AboutPanel` with MVC in mind.
An important change in this PR is that now a `Hyperlink` object is used independent whether the revision is unknown or not. Instead the hyperlink is disabled if the revision is unknown.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/setting/skin/AboutPanelSkin.java
```diff
@@ -0,0 +1,147 @@
...
+        final BooleanBinding disableProperty = Bindings.when(Bindings
+                .and(Bindings.isNotNull(revisionText), Bindings.notEqual(revisionText, "unknown")))
+                .then(false).otherwise(true);
+
+        final Hyperlink gitRevisionHyperlink = new Hyperlink();
+
+        gitRevisionHyperlink.textProperty().bind(revisionText);
+
+        // if the git revision equals "unknown" disable the hyperlink
+        gitRevisionHyperlink.disableProperty().bind(disableProperty);
```
plata: Does it make sense to handle this with bindings? The information presented on the about panel is static.

madoar: I see no disadvantages of using bindings in this case. In addition we don't know whether we will need to change the shown version sometime in the future at runtime.
Maybe we'll add an integrated updater later on

plata: Ok.

PR #1775, opened 2019-01-15
.exe self extractible archive support

### LINKED ISSUE #1600: Extractor is not able to extract .EXE
### Expected behavior
- .EXE are extracted 

### Actual behavior
- .EXE are not extracted 

### Steps to reproduce
- Download https://web.archive.org/web/20130218065827if_/http://download.microsoft.com/download/winntwks40/paint/1/nt4/en-us/paintnt.exe 
- Extract with Extractor
- Exception is thrown 

### System information
N/A

### LINKED ISSUE #1643: .exe icon extraction support
We need a component that can extract the .icon from a .exe file. This would be used in #1641 but it could also help generating miniatures in some cases

qparis:
Fixes #1600


phoenicis-tools/src/main/java/org/phoenicis/tools/stream/CursorFinderInputStream.java
```diff
@@ -0,0 +1,60 @@
...
+ * Example:
+ * <p>
+ * Let's suppose we have an input stream that is sending 0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06...
+ * Then, if the cursor is [0x02, 0x03], the new input stream will send 0x02, 0x03, 0x04, 0x05, 0x06...
+ */
+public class CursorFinderInputStream extends InputStream {
+    private final byte[] cursor;
+    private final InputStream inputStream;
+    private int cursorPosition = 0;
+    private int readPosition = 0; // Once the cursor is found, we need to send it through the input stream.
```
madoar: The comment should be prepended to the variable definition:

```java
/**
 * Once the cursor is found, we need to send it through the input stream
 */
private int readPosition = 0;
```

phoenicis-tools/src/main/java/org/phoenicis/tools/stream/CursorFinderInputStream.java
```diff
@@ -0,0 +1,60 @@
+package org.phoenicis.tools.stream;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This input stream will takes an input stream, a cursor (a byte-array) and find it
```
madoar: take not takes

phoenicis-tools/src/main/java/org/phoenicis/tools/stream/CursorFinderInputStream.java
```diff
@@ -0,0 +1,60 @@
...
+ * before creating a sub input stream that skip all data before the cursor.
```
madoar: skips not skip

phoenicis-tools/src/main/java/org/phoenicis/tools/stream/CursorFinderInputStream.java
qparis: Done

phoenicis-tools/src/main/java/org/phoenicis/tools/stream/CursorFinderInputStream.java
qparis: Done

phoenicis-tools/src/main/java/org/phoenicis/tools/stream/CursorFinderInputStream.java
qparis: Done

phoenicis-tools/src/main/java/org/phoenicis/tools/files/FileAnalyser.java
```diff
@@ -81,11 +80,25 @@ public String getDescription(File inputFile) {
...
             return mimeTypesMap.getContentType(inputFile);
         }
     }
+
+    private String guessMimeTypeFromDescription(MagicMatch match) {
+        if ("MS-DOS executable (EXE)".equals(match.getDescription())) {
+            return "application/x-dosexec";
+        }
+
+        return "???";
```
madoar: Why not return `null`?

qparis: To keep jmimemagic conventions

qparis:
An Travis too... @plata @madoar if you are ok, we can merge this, this should help script writing

madoar:
If we merge this as is, won't it lead to a failure of the master build on travis?

madoar:
Hmm... this is interesting in its own kind.
The error message from travis is:
```
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk11" --workspace "/home/travis/.cache/install-jdk" --feature "11" --license "BCL"" failed and exited with 8 during .
```

@plata @qparis do you have any ideas why this is the case?

qparis:
Random bug I guess. Can we re-trigger the build?

madoar:
I already restarted the build twice, it didn't help^^

qparis:
Do we really need to build with Oracle JDK?

madoar:
I don't think it's a question of needing oraclejdk. I think it is more a question of what could we lose if we remove it? Oraclejdk is one among many JDKs, if we want to support as many linux and mac users as possible, I think we should support as many JDK versions/vendors as possible. Therefore I wouldn't suggest removing it.

I believe this to be a problem with Travis. I'm not sure whether an update of the oraclejdk is the root cause or a change on travis.

qparis:
This seems to be a travis problem indeed. It seems to happen everytime Oracle update their version and until travis update their scripts. 

My question was more about the chances that a unit test breaks with OracleJDK and not OpenJDK.

madoar:
I don't think that there is a high change that the test will fail on oraclejdk, while they don't on openjdk.
But just in case this happens in the future I think we shouldn't remove oraclejdk from the tests.

PR #1778, opened 2019-01-17
Refactor the `RepositoriesPanel` with MVC in mind

madoar:
One key aspect of this PR is the replacement of the `ListView` that contains all added repositories through a `TableView`. 

This PR is still work in progress.

Open issues:
I'm unable to get the drag and drop behavior working with the `TableView`.
I'm not sure whether it is actually working and it is just my installation or it is a programming problem.
@plata @qparis can you give it a try to see if it works for you? If not do you have an idea why?


plata:
Let us know when the PR is finishes from your side (I don't think it makes too much sense if we test it while it's still changing).

madoar:
This is exactly the right moment to test the PR, because the repository changes are not persisted and not applied to the rest of the application, therefore you can't accidentally destroy your repository settings.

The issue here is that the drag and drop on the table doesn't work for some reason and I can't find out why...

plata:
Sometimes. Sometimes I get a cursor with "not allowed" sign. If I click somewhere, that disappears and I can drag again.

madoar:
This is how it looks like now:

grafik

@plata can you test it again to see if a preview image during the drag and drop process is now shown?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/settings/SettingsView.java
```diff
@@ -105,6 +111,38 @@ private void initializeSettingsItems() {
...
+
+        final RepositoriesPanel repositoriesPanel = new RepositoriesPanel(repositoryLocationLoader,
+                repositoryLocations);
+
+        // set the initial values
+        repositoriesPanel.setOnRepositoryRefresh(repositoryManager::triggerRepositoryChange);
+
+        // react on changes
+        repositoriesPanel.getRepositoryLocations()
+                .addListener((ListChangeListener.Change<? extends RepositoryLocation<? extends Repository>> change) -> {
```
madoar: This is not the cleanest way to do this.

A more solid and clean solution would be to use a `MappedList` to map the `RepositoryLocation` objects to `Repository` objects. The issue with this approach is, that the `RepositoryManager` would need to contain an `ObservableList` internally. With our current structure this is a bad idea, because the `RepositoryManager` is also used from the `cli` tool, which shouldn't depend on JavaFX.

A possible solution would be to implement two `RepositoryManager` classes: one for the javafx application and a second one for the cli application.

@qparis @plata what do you think?

plata: Don't think it's a good idea to have two `RepositoryManager` classes.

madoar:
Sadly I can't test the drag and drop functionality.
Every time I try it I get the following error message:

```
(java:4842): GLib-GObject-WARNING **: /build/glib2.0-7ZsPUq/glib2.0-2.48.2/./gobject/gsignal.c:2517: signal 'expose-event' is invalid for instance '0x7f325c86c710' of type 'GtkWindow'
```

In theory the line `dragboard.setDragView(row.snapshot(null, null));` is responsible for showing the preview image. Can you experiment a bit with it (the parameters) to see if it changes anything?

plata:
I'm not sure. Without the preview you have pretty much no control over what you are doing. I think I would prefer "up" and "down" buttons to change the order. Should be easier implementation wise and also better understandable for the users.

plata:
Mh. If you have this GTK issue, others will have it as well. If it's a general issue with the implementation of that JavaFX component, I don't think we can use it.

madoar:
I have multiple issues with JavaFX 11, which as far as I know I didn't have with JavaFX 8.
Another message I see basically whenever I start Phoenicis is:
```
(java:9636): Gdk-WARNING **: XSetErrorHandler() called with a GDK error trap pushed. Don't do that.
```

Therefore I don't think we should refrain from using a JavaFX component just because it leads to a warning/error on my system.

plata:
Ok. So what about: keep the list like it was and merge the MVC changes only. Then we can take care of the drag and drop etc. separately.

madoar:
I'm fine with that. I don't absolutely want to hold on to the drag and drop functionality. It is just the most intuitive approach I can think of to change the order between repositories. If we can think of a better, more intuitive and easier way I'm also fine with that. The most important point is that we find a solution that we can use long term.

plata:
Then let's do it like this.

madoar:
I think I will need to upgrade to a newer version of Linux Mint...
A version of this PR using `ListView` instead of `TableView` produces similar results on my system. I don't think it is worth going with `ListView`.

 see the following message quite often when I try to drag a repository with the `TableView`:
```
(java:16204): GLib-GObject-WARNING **: /build/glib2.0-7ZsPUq/glib2.0-2.48.2/./gobject/gsignal.c:2517: signal 'expose-event' is invalid for instance '0x7f97409e2710' of type 'GtkWindow'
```
Every time I see this message it results in the drag and drop process to fail.

Additionally when changing the order of the repositories to fast I get the following two errors independently:

```
Exception in thread "JavaFX Application Thread" java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:458)
	at java.base/java.util.Collections$UnmodifiableList.get(Collections.java:1308)
	at phoenicis.javafx@5.0-SNAPSHOT/org.phoenicis.javafx.controller.MainController.lambda$setDefaultCategoryIcons$3(MainController.java:126)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:428)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:427)
	at javafx.graphics/com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkDnDClipboard.pushToSystemImpl(Native Method)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkDnDClipboard.pushToSystem(GtkDnDClipboard.java:39)
	at javafx.graphics/com.sun.glass.ui.SystemClipboard.flush(SystemClipboard.java:51)
	at javafx.graphics/com.sun.glass.ui.ClipboardAssistance.flush(ClipboardAssistance.java:59)
	at javafx.graphics/com.sun.javafx.tk.quantum.QuantumClipboard.flush(QuantumClipboard.java:274)
	at javafx.graphics/com.sun.javafx.tk.quantum.QuantumToolkit.startDrag(QuantumToolkit.java:1242)
	at javafx.graphics/javafx.scene.Scene$DnDGesture.dragDetectedProcessed(Scene.java:3047)
	at javafx.graphics/javafx.scene.Scene$DnDGesture.process(Scene.java:3116)
	at javafx.graphics/javafx.scene.Scene$DnDGesture.access$8300(Scene.java:3003)
	at javafx.graphics/javafx.scene.Scene$MouseHandler.process(Scene.java:3867)
	at javafx.graphics/javafx.scene.Scene$MouseHandler.access$1200(Scene.java:3579)
	at javafx.graphics/javafx.scene.Scene.processMouseEvent(Scene.java:1849)
	at javafx.graphics/javafx.scene.Scene$ScenePeerListener.mouseEvent(Scene.java:2588)
	at javafx.graphics/com.sun.javafx.tk.quantum.GlassViewEventHandler$MouseEventNotification.run(GlassViewEventHandler.java:397)
	at javafx.graphics/com.sun.javafx.tk.quantum.GlassViewEventHandler$MouseEventNotification.run(GlassViewEventHandler.java:295)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.tk.quantum.GlassViewEventHandler.lambda$handleMouseEvent$2(GlassViewEventHandler.java:434)
	at javafx.graphics/com.sun.javafx.tk.quantum.QuantumToolkit.runWithoutRenderLock(QuantumToolkit.java:390)
	at javafx.graphics/com.sun.javafx.tk.quantum.GlassViewEventHandler.handleMouseEvent(GlassViewEventHandler.java:433)
	at javafx.graphics/com.sun.glass.ui.View.handleMouseEvent(View.java:556)
	at javafx.graphics/com.sun.glass.ui.View.notifyMouse(View.java:942)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication.lambda$runLoop$11(GtkApplication.java:277)
```

```
java.util.ConcurrentModificationException: null
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1660)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.ForEachOps$ForEachTask.compute(ForEachOps.java:290)
	at java.base/java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:746)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinTask.doInvoke(ForkJoinTask.java:408)
	at java.base/java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:736)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateParallel(ForEachOps.java:159)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateParallel(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:661)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:575)
	at phoenicis.repository@5.0-SNAPSHOT/org.phoenicis.repository.types.MultipleRepository.fetchInstallableApplications(MultipleRepository.java:54)
	at phoenicis.repository@5.0-SNAPSHOT/org.phoenicis.repository.types.CachedRepository.fetchInstallableApplications(CachedRepository.java:36)
	at phoenicis.repository@5.0-SNAPSHOT/org.phoenicis.repository.types.Repository.fetchInstallableApplications(Repository.java:49)
	at phoenicis.repository@5.0-SNAPSHOT/org.phoenicis.repository.types.BackgroundRepository.lambda$fetchInstallableApplications$0(BackgroundRepository.java:51)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
```

I think we will need to improve our repository fetching strategy.

plata:
Any idea what exactly is causing this?

madoar:
### About the system error

After searching the internet just now I found [this](https://stackoverflow.com/questions/54363577/openjfx-gdk3-bug-with-drag-n-drop-workaround). When I perform the suggested workaround the drag and drop feature with the `TableView` works for me!

### About the two Java errors

The errors are most likely caused by multiple concurrent accesses on the repositories. This can happen when two repositories are swapped, while the repositories are updated (fetched).

madoar:
Here a short git showcasing the new repositories panel:
change-repositories

The only remaining issue are the occurring exceptions after a change has been performed to the repositories. I think we should fix the concurrency issue in its own PR.

plata:
Did we have the exceptions before already or have they been introduced with this PR?

madoar:
They occurred for me with this PR, but in theory they were already possible before.
In my opinion the main problem is, that every drag and drop action is split in two subactions:
- the removal of the dragged repository from its old position in the repository list
- the addition of the dragged repository in the new position in the repository list

Because the repository refresh/reload happens asynchronous and these actions occur at nearly the same time, Phoenicis does not have enough time to reload the repositories after the removal operation, before the repository is added again. This in turn I believe leads to the previous mentioned exceptions.

madoar:
Ok now the order of the repositories can be changed without throwing exceptions.
In case the ordering is changed multiple times, the user needs to watch out and wait 2 or 3 seconds in between, otherwise the exceptions are thrown nevertheless. This interval is necessary to give the application enough time to process the change to the repository order.

plata:
Ok. I think we can go with that. Just be sure to open a new issue to really fix the remaining exception problems.

phoenicis-repository/src/main/java/org/phoenicis/repository/DefaultRepositoryManager.java
```diff
@@ -80,6 +80,40 @@ public void moveRepository(RepositoryLocation<? extends Repository> repositoryUr
...
+        repositoryLocations.forEach(repositoryLocation -> {
+            if (!this.repositoryMap.containsKey(repositoryLocation)) {
+                final Repository repository = repositoryLocation.createRepository(cacheDirectoryPath,
+                        localRepositoryFactory,
+                        classPathRepositoryFactory, fileUtilities);
+
+                this.repositoryMap.put(repositoryLocation, repository);
+            }
+
+            this.multipleRepository.addRepository(this.repositoryMap.get(repositoryLocation));
```
plata: Wouldn't this duplicate the repository in `this.multipleRepository` if it is contained in `this.repositoryMap` already (for whatever reason)?
Can we avoid this duplicate maintenance of `multipleRepository` and `repositoryMap`?

madoar: In line `91` all repositories are removed from the `multipleRepository`. This is done to ensure that:
- the ordering of the repositories inside `multipleRepository` is correct
- no duplicates are contained inside `multipleRepository`

madoar: I guess we can remove this redundancy, but in my opinion this should be part of a complete refactoring of the whole repository model.

plata: Ok.

PR #1779, opened 2019-01-18
Refactor the `UserInterfacePanel` with MVC in mind

madoar:
This PR refactors the `UserInterfacePanel` class with MVC in mind.
In addition it changes the way changes to the interface are handled (now via bindings) and "fixes" the layout:

grafik


plata:
Do all 3 bold texts use the same font? It looks like the first one is a bit different. Also, I find it strange that there is a `:` after "Erscheinungsbild" but not after the other ones. I guess it was like this before but as you're on it already, it should be fixed as well.

madoar:
Should all three labels have a `:` afterwards or none?

madoar:
@plata I've now removed the `:` from after `Theme:`

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/setting/control/UserInterfacePanel.java
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public UserInterfacePanelSkin createSkin() {
+        return new UserInterfacePanelSkin(this);
+    }
+
+    public ObservableList<Theme> getThemes() {
+        return themes;
```
plata: `this` for member variables

madoar: Done

PR #1789, opened 2019-01-20
Refactor the CreateShortcutPanel class with MVC in mind

madoar:
This PR refactors the `CreateShortcutPanel` class with MVC in mind.
In addition it adds a title to the panel and defines its own css:

grafik


madoar:
I don't like Codacy too :P

Back on topic:
I'm thinking about renaming the three details panel classes for the library tab:
- `CreateShortcutPanel` to `ShortcutCreationDetailsPanel`
- `EditShortcutPanel` to `ShortcutEditingDetailsPanel`
- `LibraryDetailsPanel` to `ShortcutDetailsPanel`

What do you think?

plata:
Makes sense.

PR #1791, opened 2019-01-21
Refactor the ContainerEngineToolsPanel with MVC in mind

madoar:
This PR refactors the `ContainerEngineToolsPanel` with MVC in mind.
In addition it replaces the used `TilePane` for the Visualization of the tool buttons with a `FlowPane`.


plata:
How does the behavior change by using `FlowPane` instead of `TilePane`?

madoar:
According to my tests it doesn't. The only change is that we have no preferred number of columns anymore.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainerEngineToolsPanel.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public ContainerEngineToolsPanelSkin createSkin() {
+        return new ContainerEngineToolsPanelSkin(this);
+    }
+
+    public ContainerDTO getContainer() {
+        return container.get();
```
plata: add `this` (also below)

PR #1797, opened 2019-01-23
Fix bug tracker url type

AsciiWolf:


plata:
Shall we set the forum for type `help`?

AsciiWolf:
@plata The "help" type should be used for "online user's reference, a software manual or help page", as noted [here](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html). ;-)

plata:
Understood. Do you think it makes sense to point help to https://phoenicisorg.github.io/phoenicis/?

AsciiWolf:
Yup, that looks reasonable. :-)

plata:
Would you mind providing a pull request for it?

PR #1798, opened 2019-01-24
Update README.md ... adding additional info ...

aiamuzz:
Would this additional bits of information be helpful ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,14 @@ cd phoenicis-javafx/
 mvn exec:java
 ```
 
+For more details on installing on various linux distro's [Install/Build](https://phoenicisorg.github.io/phoenicis/Developers/build/)
+
 For more details (e.g. regarding dependencies) consider the [documentation](https://phoenicisorg.github.io/phoenicis/).
+
+## Flatpak Install
+
+Installing from [Flathub](https://flathub.org/apps/details/org.phoenicis.playonlinux)
+Issues while running Flatpak version: https://github.com/flathub/org.phoenicis.playonlinux/issues
 
 ## Scripts
 To add or update scripts, please make pull requests to this repository: https://github.com/PhoenicisOrg/scripts


plata:
I think we misunderstood each other. I understood that you wanted to add a README with the compat dependency to https://github.com/flathub/org.phoenicis.playonlinux.

However, we can also update the README here (adding an "installation" section like you did is a good idea). Nevertheless, we should not restrict this to flatpak because there are other possibilities as well (e.g. deb). Therefore, I would suggest to only link to https://phoenicisorg.github.io/phoenicis/Users/installation/ and explain everything there.

aiamuzz:
> I think we misunderstood each other. I understood that you wanted to add a README with the compat dependency to https://github.com/flathub/org.phoenicis.playonlinux.
> 
Yeah i meant only add it as info ... at least the users will know that the compat dependency is needed ... else they will have to report the issue and somebody will have to make it known to them ... 

> However, we can also update the README here (adding an "installation" section like you did is a good idea). Nevertheless, we should not restrict this to flatpak because there are other possibilities as well (e.g. deb). Therefore, I would suggest to only link to https://phoenicisorg.github.io/phoenicis/Users/installation/ and explain everything there.


ok makes sense ... but i am kind of new to github.com ... i see github.io ... and the UI is different there ... how can i create a PR or add information to the Installation page there ?

Honestly the build and installation instructions are pretty confusing both on github.com and github.io ...

plata:
It is generated from: https://github.com/PhoenicisOrg/phoenicis/blob/master/docs/_docs/Users/installation.md

README.md
```diff
@@ -30,8 +30,15 @@ cd phoenicis-javafx/
 mvn exec:java
 ```
 
+For more details on installing on various linux distro's [Install/Build](https://phoenicisorg.github.io/phoenicis/Developers/build/)
```
AsciiWolf: Typo: Should be "distros", not "distro's". Also, "For more details on installing on various linux distros, see:" sounds better.

README.md
```diff
@@ -30,8 +30,15 @@ cd phoenicis-javafx/
...
+
 For more details (e.g. regarding dependencies) consider the [documentation](https://phoenicisorg.github.io/phoenicis/).
 
+## Flatpak Install
```
plata: We can use the official flathub badge for this purpose. See #2145.

plata:
Superseded by #2145 and #2146.

PR #1799, opened 2019-01-24
Cleanup of LibraryView and subcomponents

madoar:
This PR performs some cleanup and refactoring of the `LibraryView` and its subcomponents.
In addition it introduces a new class `SwitchBindingBuilder` which allows the creation of bindings with `switch`/`case` functionality.


madoar:
@qparis we need your expertise before we can merge this PR ;)

qparis:
I will have a look on this tonight

madoar:
@qparis I think you forgot this PR^^

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/ShortcutDetailsPanel.java
```diff
@@ -63,45 +63,50 @@ public LibraryDetailsPanel(ObjectMapper objectMapper, ObjectProperty<ShortcutDTO
...
      * {@inheritDoc}
      */
     @Override
-    public LibraryDetailsPanelSkin createSkin() {
-        return new LibraryDetailsPanelSkin(this);
+    public ShortcutDetailsPanelSkin createSkin() {
+        return new ShortcutDetailsPanelSkin(this);
     }
 
     public ObjectMapper getObjectMapper() {
```
qparis: Why do we need to wrap the object mapper inside an ObjectProperty?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/utils/SwitchBindingBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+import javafx.beans.property.ObjectProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A builder class for <code>switch</code> like {@link ObjectBinding}s
```
qparis: I think this should be put outside phoencis (can be done later). 
Also, you need to provide simple examples to see how it works exactly.
Overall, it looks clean

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/ShortcutDetailsPanel.java
madoar: We don't need to, it is just more "clean" and flexible from a JavaFX perspective.
The idea behind using `ObjectProperty` objects is to be able to change the concrete value of the property later on without requiring to recreate the whole component.
In addition this enables me to use setter methods to specify this information without abstaining from `final` fields.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/utils/SwitchBindingBuilder.java
madoar: You're right about that. I'm planning to move the whole utils package to a new repository later on. (Similar to the collections)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/ShortcutDetailsPanel.java
qparis: I understand but ObjectMapper is a bean (aka a Singleton inside Phoencis). I think we should use the convention not to make bean observables, what do you think?

madoar: That is independent from the JavaFX component perspective.
The idea is to be able to view each JavaFX component independently as a kind of sandbox.
This makes it also much easier to write tests for JavaFX components, because you don't need to setup a lot of external objects to pass to a JavaFX component.

qparis: Ok

madoar:
@plata you need to update your review, otherwise I can't merge :/

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/ShortcutDetailsPanel.java
madoar: After thinking about this some more I think I should rephrase my previous comment:

I agree with you that the `ObjectMapper` should be set only once, but for me there are multiple reasons to use an `ObjectProperty` for the `ObjectMapper` instance and other objects which are defined only once in Phoenicis:
- in JavaFX components object fields are represented by `ObjectProperty` objects, I want to adopt this pattern when possible in our components
- I want to define all fields in our components as `final` fields, but at the same time I don't want to pass parameters to a JavaFX component via a constructor. `ObjectProperty` objects allow this

PR #1810, opened 2019-01-26
Jmimemagic + LnkParser: Use external repo

qparis:
Should fix extractor problems


phoenicis-tools/src/main/java/org/phoenicis/tools/ToolsConfiguration.java
```diff
@@ -153,6 +153,6 @@ public Opener opener() {
 
     @Bean
     public LnkParser lnkParser() {
-        return new LnkParser();
+        return new LnkParser(new org.phoenicis.lnk.LnkParser());
```
madoar: Can we rename the `LnkParser` class in Phoenicis so that we don't have a naming conflict?

qparis: Any better idea of name?

madoar: `PhoenicisLnkParser`? It is basically a wrapper class for the moved `LnkParser` right? So the name is not really important as long as it retains its functional meaning.

qparis: This will be used in the script, so the name is important

madoar: We could also name it `LnkParserWrapper`, depending on whether the class name is exposed to the scripts...

madoar: Hmmm, in that case it is hard to find a different name, I agree.

PR #1815, opened 2019-01-31
Refactor the `ContainerVerbsTab` class with MVC in mind

madoar:
This PR replaces the `ContainerVerbsTab` class with the two classes `ContainerVerbsPanel` and `ContainerVerbsPanelSkin`, which need to be wrapped inside a `Tab` object.
In addition this PR replaces the `TiledPane` of the `ContainerVerbsTab` with a `GridPane`, that allows for the selection of multiple verbs to install at the same time.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainerVerbsPanel.java
```diff
@@ -0,0 +1,128 @@
...
+     * A list of all tool scripts
+     */
+    private final ObservableList<ScriptDTO> verbScripts;
+
+    /**
+     * A boolean signifying whether all verb buttons should be locked
+     */
+    private final BooleanProperty lockVerbs;
+
+    public ContainerVerbsPanel(ObjectProperty<ContainerDTO> container, ObjectProperty<VerbsManager> verbsManager,
```
plata: javadoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainerVerbsPanel.java
```diff
@@ -0,0 +1,128 @@
...
+        return verbScripts;
+    }
+
+    @Override
+    public ContainerVerbsPanelSkin createSkin() {
+        return new ContainerVerbsPanelSkin(this);
+    }
+
+    public ContainerDTO getContainer() {
+        return container.get();
```
plata: add `this` (also below)

madoar:
This is how the new verbs installation tab looks like now:

grafik

plata:
For me it's a bit strange that every verb has it's own install button additionally to the general install button.

madoar:
I believe it is a question of preferences. In case a user want to install only a single verb it is more intuitive for the user in my opinion to click on the `Install` button for the desired verb, while in case the user wants to install multiple verbs at the same time, it is more intuitive to select them first and then to install them.

madoar:
@qparis what do you think about the visual redesign of the container verbs tab?

plata:
I'm fine with the checkboxes in general. Like I said, I would remove the install buttons because I believe that they are confusing. For example: Do I have to click the checkbox to install a single verb? Probably not. But I have to do so to install multiple verbs. So the same GUI component is used differently depending on the context. That's not a good idea in general.
You can also compare this to software management tools like pamac or muon.

Maybe it would also be a good idea to show a dialog: "Are you sure you want to install these verbs: mspatcha, vcrun2013".

madoar:
Hmm, so do you suggest to only remove the install buttons or do you want to replace them with some other components?

Let's move the suggestion of adding a confirm dialog when installing verbs to a new issue.

ImperatorS79:
One could rename the button "Information" for https://github.com/PhoenicisOrg/scripts/issues/823 ?

madoar:
How/where should the description be shown after the user has pressed on an information button?
In my opinion it is better to show the description as a tooltip when the user hovers his/her mouse on top of a verb name or its checkbox.

ImperatorS79:
Why not. I am just making propositions.

madoar:
The issue with an `Information` button is, that it requires an addition component/dialog which shows the information after clicking on the button, therefore the question how where we would show the information for the verb(s). A tooltip doesn't require an additional component, but has the disadvantage that it is not as intuitive as a button solution.

Again I think we should move the topic of showing the description for the verbs to its own PR.

qparis:
I don't see how it would work if you remove the install button. Can you show me?

plata:
You select the check box and then click "install selected".

madoar:
This is how it looks now:
grafik
grafik

qparis:
I think it is much better

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainerVerbsPanel.java
```diff
@@ -0,0 +1,154 @@
...
+    public void setVerbs(ApplicationDTO verbs) {
+        this.verbs.set(verbs);
+    }
+
+    public ObservableList<ScriptDTO> getVerbScripts() {
+        return this.verbScripts;
+    }
+
+    public boolean isLockVerbs() {
+        return lockVerbs.get();
```
plata: `this.lockVerbs` (like below)

madoar: I must have overlooked that one...

PR #1821, opened 2019-02-05
Add dedicated DetailsPanel container component to replace the DetailsPanelBase class

### LINKED ISSUE #1804: Restructure `DetailsPanel` implementations similar to `Tab`
When thinking about our usage of the `DetailsPanel` implementations I see similarities to the usage of the `Tab` class from JavaFX. The `Tab` class is used as a container for other components. In addition the `Tab` class provides a set of utility fields/functions (e.g. a title and a close button/listener). This functionality is very similar to our `DetailsPanel`.

For that reason I suggest we restructure our `DetailsPanel` implementations similar to the `Tab` class.
In detail I suggest we provide a single non extendable `DetailsPanel` class, which takes a second component as its content. All our current `DetailsPanel` implementations would then be changed to normal JavaFX components and be added to a `DetailsPanel` component as its content.

madoar:
This PR closes #1804 by adding a new `DetailsPanel` and `DetailsPanelSkin` class, which are used to replace the previous used `DetailsPanelBase` base class.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/common/control/DetailsPanel.java
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public DetailsPanelSkin createSkin() {
+        return new DetailsPanelSkin(this);
+    }
+
+    public String getTitle() {
+        return title.get();
```
plata: add `this` (also below)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/utils/StringBindings.java
```diff
@@ -0,0 +1,41 @@
...
+
+import javafx.beans.binding.Bindings;
+import javafx.beans.binding.StringBinding;
+import javafx.beans.value.ObservableValue;
+
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ * An utility class containing functions to more map {@link ObservableValue} objects to {@link StringBinding} objects
```
plata: "A utility"

The "more" seems wrong.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/common/control/DetailsPanel.java
madoar: Done

phoenicis-javafx/src/main/java/org/phoenicis/javafx/utils/StringBindings.java
madoar: Done

Issue #1024, opened 2017-08-22
develop concept for generic engines


plata:
This issue collects everything related to generic engines to develop an overall concept which shall then be broken down into several issues for implementation.

### Engines
- it might be possible to extend a Java interface in JS (see #983)
- JS must be asked if a certain engine is installed (see #605)

### Containers
As the containers highly depend on the engine which is using them, the container handling has to be done in JS as well. 
- container information specific for a certain engine must be loaded (see also  #477)
- the engine version used by a specific container must be updated (see #992 and #702)
- specific settings must be handled in a generic way (see #620)

ImperatorS79:
Until this is done, you could add a simple way to update wine containers version (and distribution) (like a list of already installed wine version) an remove it when you implemented your generic engines.

plata:
Done.

Issue #992, opened 2017-08-20
Implement an update engine wizard for the containers


madoar:
Currently it is not possible to change the engine of an existing container.
Therefore we should implement a wizard for this this.
This wizard should be reachable by a button inside the container view. After clicking on this button a new tab or dialog should open which shows a list of existing engines. This list should look like the one located inside the "Engines" tab, where all already installed engines are marked. After selecting an engine inside the wizard a details view at the right side of the wizard should open, where the user finds a button through which he has the choice to update the engine of his container to the selected engine.

For more context about this see #974

plata:
consider also #702

plata:
Superseeded by #1024.

Issue #702, opened 2017-04-13
Implement generic engine changeVersion


plata:
It could be implemented similarly to engine install/delete, i.e. 
- add `ChangeEngineVersionUi` (see `EngineProgressUi`)
- implement `changeVersion` in JS
- call `changeVersion` of correct engine depending on container from Java

see #698

madoar:
This should be implemented relatively soon, because we currently only have a half functional engines list in the container view, which is for one slow during the initialization and doesn't do anything afterwards.

plata:
Superseded by #1024.

PR #1829, opened 2019-02-09
Refactor the `ContainerInformationPanel` class with MVC in mind

madoar:
This PR refactors the `ContainerInformationPanel` with MVC in mind. This includes splitting the previous `ContainerInformationPanel` class into the two classes `ContainerOverviewPanel` and `ContainerOverviewPanelSkin`. The `ContainerOverviewPanel` needs then to be wrapped inside a `Tab` object.

Screenshot:
grafik


ImperatorS79:
@madoar I will take advantage of this PR to ask you if you have any idea for #625 ? I looked at the previous code and it seems it fetched a list of available verb, but I think it should be easier to simply propose engine from already installed ones.

madoar:
The issue with #625 is not the selection of the engine, which the user wants to install. This selection can be implemented quite easily either via a drop down menu or a wizard dialog. The problem is how to install/apply the selected engine to a container? This contains both the installation of the new engine and the deinstallation of the old engine.

ImperatorS79:
Normally, wine automatically detects the version mismatch and recreates what is needed. That is what happens when I manually edit phoenicis.cfg.

madoar:
I'm not really an expert with wine. I'm not sure what needs to be done to install or uninstall wine from a container. As long as we have an `install` method and ideally an `uninstall` method for the engines it should be possible to change the engines. 

For the UI part of changing the engine of a container:
I would recommend to provide a dialog containing a `CombinedListWidget` and a details panel, like in the engines tab. When selecting an engine in the `CombinedListWidget` the information of the selected engine should be shown in the details panel together with an `Install`/`Apply` button.

qparis:
Wine is not installed inside a container, it is the other way around :) A container uses an engine

madoar:
@qparis do we have a method for wine which applies it to an existing container?

madoar:
@qparis that's what I meant with I'm not an expert about wine :P

qparis:
What do you mean by a method which applies it to an existing container?

madoar:
When changing the engine of a container I want to apply a new engine (engine version) on the container. This replaces the previously used engine (engine version) with the new one.

qparis:
Basically you just need to update playonlinux.cfg. Don't know if we need to pass through JS though

madoar:
The question is how transferable this is to other types of engines?
Can we assume that all engine types can be maintained separately from the containers?

qparis:
I think we need to differenciate the engine type (Wine) and the engine instance (wine 2.0.3 for example).

The engine type should not be changed, the instance can be. 
Then, this engine kind exposes stuff that can be changed (like the version)

madoar:
That is fine by me. Still because we want to provide a generic interface for all types of engines on the Java side we need to move the "engine change" code to JS, I think.

qparis:
Yes ideally. The main problem I see is that it takes much much time to bring up new features with this design

madoar:
I don't understand why it takes much time to bring up new features when we provide a generic interface? If we want to provide a new feature we can first provide an empty dummy implementation for the new feature for the interfaces, this should go quite fast. Afterwards we can then provide a concrete implementation later on. The important part is to decide which information are required for the new feature.

plata:
As a matter of fact, you simply cannot know what an engine might have to do to switch the used version in a container. Therefore, the engine interface must provide a function `changeVersion(container, version)` which is implemented by every engine.

qparis:
Or it can be a property like any other (windows version, etc...)

madoar:
The issue is that the Java code requires the `changeVersion` function to be present, otherwise an error could occur when you want to change the version of an engine which doesn't provide such a function.

Like mentioned before I don't see a reason why we shouldn't add the function to the engine interface. If an engine doesn't provide different versions it could provide an empty dummy implementation.

qparis:
I don't think we need to have explicitely a changeVersion method. I think we can implement it exactly the same way we have implementend the "change windows version" setting

qparis:
Also, keep in mind that the behavior is very specific to wine: you can change the version, the distribution (a concept that does not necessarely exist with other engine by the way), but not the architecture

madoar:
So you propose to implement the `changeVersion` method as a verb? I don't think this is intuitive:
- the user can change the wine version inside an installation script. This doesn't seem useful to me?
- in Phoenicis the user needs to change the engine version inside the verbs tab and not the overview tab of a container

qparis:
windows XP is not in the verb tab AFAIK but on "setting tab", which is quite right. The version can be remembered on the overview tab

madoar:
This has the same problem. If we do it like you suggest we would need to implement two exceptions:
- the `change version` setting needs to be filtered out in the settings tab
- the `change version` setting needs to be manually added to the overview tab

I fail to see the advantages of implementing such functionality in an external file as a plugin/verb/setting. I think this functionality belongs to either:
- the engine interface and its implementations (i.e. `Wine` directly) or
- a new interface which is responsible for swapping an engine version of an existing container

Under a certain viewpoint all engine types need to provide to provide a `changeVersion` method, because it can also be used to do the initial setup of an empty container (i.e. an empty folder) with the engine.

qparis:
The concept of version does not exist with all the engine, especially when talking about architecture and distiribution 

I don’t understand what you need to filter out

madoar:
The engine settings are fetched in a generic way, i.e. all scripts located in `engines/tools` are interpreted as engine settings. If we now add a `changeVersion` setting here, which we do **not** want to show in the engine settings tab of a container we need to filter it out.

I don't think it is a problem that not all engines support changing the engine. In this case we simply need to a provide a dummy method for these engine types, for example:

```javascript
function changeVersion(container, version) {
   // do nothing
}
```

Or if we want to do provide an implementation which allows for a once only execution:
```javascript
function changeVersion(container, version) {
   if (!hasEngineInstalled(container)) {
      ...
   }
}
```

qparis:
I don’t see why you don’t want to show it on the setting tab?

madoar:
Because it is not intuitive in my opinion. All information about the used engine is shown on the overview tab. So as a user I would also expect a button `Change engine` or `Change engine version` on this tab.

In addition I wouldn't call a version change as a setting of an engine. I mean a version change could introduce some new settings or remove some old ones, but in my opinion it isn't a setting itself.

qparis:
I think it is intuitive because the engine is “wine” and not “wine 4.0.1”. The architecture also follow this pattern. 

Changing a version won’t introduce a new setting because we did not architectured phoenicis to make differences between versions which make sense

ImperatorS79:
I personally agree with @madoar that is more intuitive to have it on the overview tab. Adding a `changeVersion` method does not seems so difficult to do IMHO. For wine, it is really simple: update `phoenicis.cfg`.

madoar:
If I have the method I can provide a GUI dialog to use it :)

@qparis I think we should distance ourselves from the thought engine = wine. Currently that may be the case, but in the future it won't. We did a lot of work over the last two years in the direction where we can add additional engine types which are not wine. Currently we are not there yet, but I wouldn't like to see us rushing through the implementation of features which are implemented to only work with a single engine type easily. We require solutions that are generalizable and can be implemented for different engine types.

qparis:
The problem is not about the implementation, it is to know if it is a generic setting or a wine setting.

I am almost sure that with the way we have implemented this (architecture - distribution - version) is very specific to wine. 
We have implemented a generic wine engine (which can contains several version) and not a set of different wine engine. Thus, the engine is responsible for the version management. 

Another advantage I see: we can implement it now with very few impact as a start, and respond to the main demand very quickly: the ability to change the wine version of a container. (Better to have it at the wrong place than not have it at all)

qparis:
@madoar This is not what I’m trying to express. My point is that version management in wine is very specific. You do not just need to have a method that sets a version, you also need to have a method that list the possible versions, and this method depends on the architecture and the distribution, two concepts that won’t exist for other engines.
Also, I really doubt that the need of changing the version really exist on other engine (at least it is way weaker than for wine)

madoar:
> My point is that version management in wine is very specific. You do not just need to have a method that sets a version, you also need to have a method that list the possible versions, and this method depends on the architecture and the distribution, two concepts that won’t exist for other engines.

I see no problem with that, when looking at the method specification @plata mentioned earlier:
```javascript
changeVersion(container, version)
```
we can just assume that the `version` parameter is a more complex object than a simple `int` or a `String`. If we for example pass a `Map` to it we won't have the problem with expressiveness.

> Also, I really doubt that the need of changing the version really exist on other engine (at least it is way weaker than for wine)

Why? In my understand we can change the version of every engine. Let's assume we have a `dosbox` only engine. DosBox also provides different release versions, why shouldn't we be able to change it later on? For other engines the same should be the case. The only difference may be whether it makes sense to change the engine version of engines which are not wine. 

I don't think that we can currently make assumptions about whether wine is the only engine for which a version change makes sense. It is simply the case that wine is the only engine we support currently.

qparis:
I'm not making any assumption because my approach does not require any change in the current architecture and does not prevent us from changing it later if the need appears. So it does not generate any technical debt.  It just requires few lines of code (maybe < 10) and no change in java code to provide the feature we want. 

However, saying that other engines will require version - even if it is **very likely to be true** - is an assumption that has not be proven yet and that requires change in our interfaces. Creating a generic version management requires some though to be abstract enough. If we don't think this properly, then either the architecture won't work with wine (which we obviously don't want), or the architecture won't work with other engines (which we want to try to avoid). By the time we think about it, our user cannot benefit of this important feature. 
E.g. your proposal of interface looks good, but is not complete yet: 
* How are we going provide to the end user the list of available versions he can install for a given container; maybe adding a listVersion(engine) 
* Where are we going to store the template of the version so that javafx shows the right UI (one field for dosbox, two fields for wine)?

As long as we are not mature enough, I suggest developing the few JS lines so that the user can change the wine version, even if we consider that it is not in the right place (which is IMHO not a big deal because the current wine version is a "setting", isn't it?). These JS lines will be reused when the interface of the version are stable enough anyway.

madoar:
> So it does not generate any technical debt. It just requires few lines of code (maybe < 10) and no change in java code to provide the feature we want.

This is also no problem with an interface approach. We don't need to add a new interface, we simply need to extend our existing one. Take a look at: https://github.com/PhoenicisOrg/phoenicis/blob/master/phoenicis-engines/src/main/java/org/phoenicis/engines/Engine.java
We simply would need to add the following method signature:

```java
boolean changeVersion(String containerName, Map<String, Object> version);
```

or if we want to define it analogous to `createContainer`:

```java 
boolean changeVersion(String subCategory, String version, String containerName);
```

Afterwards we need to extend https://github.com/PhoenicisOrg/scripts/blob/master/Engines/Wine/Engine/Implementation/script.js with a method body for the newly created `changeVersion` method.

> E.g. your proposal of interface looks good, but is not complete yet:
    * How are we going provide to the end user the list of available versions he can install for a given container; maybe adding a listVersion(engine)
    * Where are we going to store the template of the version so that javafx shows the right UI (one field for dosbox, two fields for wine)?

I would do it the same way we allow the user to install a new engine:
grafik

Instead of an `Install` button we add a `Change Version` or `Apply` button.

qparis:
If you think it can be done easily, then go ahead :-)

madoar:
I can do the GUI part, but I'm not an expert of wine. Therefore I don't think I'm the right person to provide a concrete implementation of the `changeVersion` method. 

For some more context and maybe a better fitting places to continue the discussion:
- the approach for the `changeVersion` method has been introduced in #702
- the generic engine interface approach has been introduced in #1024
- for the GUI part I previously opened #992

ImperatorS79:
AFAIK, you only need to modify in the container the file `phoenicis.cfg` which is containing the information about the wine version, and that's it.

qparis:
To change the version you also need to warn the user, to run wineserver -k before and after the change and yes that’s it.
To enumerate the list of version, this is a bit trickier.

ImperatorS79:
Do you want to list only the already installed verion (like in POL4), or all the available ones ?

qparis:
I think it makes more sense to show all version. Ideally, we could differenciate the one that are installed and the others

plata:
This shouldn't be too much of an issue. We can basically show something similar to the engine tab in a popup.

Regarding the interface: I agree that the current `String subCategory, String version` might need to be refined. However, I support @madoar in following the generic engines approach. To do so, the Wine implementation in Javascript should look something like:
```Javascript
changeVersion: function (containerName, subCategory, version) {
    // run wineserver -k
    this.install(subCategory, version);
    // update phoenicis.cfg
    // run wineserver -k
}
```

qparis:
This is not sufficient enough. You also need a JavaScript interface to enumerate allowed versions

madoar:
@qparis long-term yes, but do we require such a function right now? Currently I think it is enough to show the user all available engine versions and the user must himself decide whether the engine version is compatible with his container or not.

Independently of whether we provide a method to list all allowed versions now or later, how would this method detect whether an engine version is compatible to a container?

qparis:
How can you show _all_ the engine? We need to manage distributions somehow

madoar:
I would use the same approach as in the Engines tab, see my screenshot above. Using:

https://github.com/PhoenicisOrg/phoenicis/blob/1ca2c073277d894f85e8f95356b326fe66af675a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/engines/EnginesView.java#L161-L188

and

https://github.com/PhoenicisOrg/phoenicis/blob/1ca2c073277d894f85e8f95356b326fe66af675a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/engine/skin/EngineSubCategoryPanelSkin.java#L56-L75

Is this approach not applicable to show all engine versions?

plata:
@madoar is correct. Showing the available engine versions is not handled in Javascript. This is not even possible because it would mean that the scripts directly control the GUI. Instead, the Java code uses the Java interface for generic engines (which is implemented in Javascript) to fetch the engine versions and handle the engine version selection. The result of this is then passed to Javascript again (via `changeVersion`).

madoar:
Is the addition of a new a dialog/component to change the engine version of an existing container required for this PR? I propose to move this to a new PR/issue and review this as is. I don't think we should mix the refactoring of our current component with the introduction of a new one in a single PR if possible.

plata:
No, it's not required for this PR. You can create a new issue which bundles the result of this discussion.

madoar:
I've created #1832 to continue the discussion, did I miss something?

plata:
Looks good for me.

madoar:
Then I need an Approve :)

PR #1833, opened 2019-02-13
Include by ID

plata:
This is implemented as an extension to the existing repository hierarchy currently because I didn't want to rewrite the whole code. Also, I'm aware that the implementation could be done in a nicer way in several places (e.g. when fetching the engine settings). However, the purpose of this PR is to provide the "include by ID" mechanism. Everything else is internal stuff which can be improved later on.


phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
```diff
@@ -91,12 +91,12 @@ private String createFetchScript(RepositoryDTO repositoryDTO) {
         script.append("(function () {\n");
         script.append("var settings = {};\n");
         for (CategoryDTO engine : categoryDTOS) {
-            final String engineId = engine.getId();
+            final String engineId = engine.getId().replaceAll("^.*\\.", "");
```
madoar: It may be worth a thought to extract this pattern to a static final variable.

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryManager.java
```diff
@@ -42,6 +42,14 @@
      */
     ScriptDTO getScript(List<String> path);
 
+    /**
+     * This method returns the {@link org.phoenicis.repository.dto.ScriptDTO} with the given ID
+     *
+     * @param id The script ID
+     * @return The found ScriptDTO
+     */
+    ScriptDTO getScript(String id);
```
madoar: Do you plan to remove the `ScriptDTO getScript(List<String> path);` method long-term?

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
```diff
@@ -53,12 +53,12 @@ private ApplicationDTO(Builder builder) {
         this.typeId = builder.typeId;
         this.categoryId = builder.categoryId;
         if (builder.id != null) {
-            if (builder.id.matches("^[a-zA-Z0-9_]+$")) {
+            if (builder.id.matches("^[a-zA-Z0-9_.]+$")) {
```
madoar: Again we should think about moving this pattern to a static final field

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
plata: That's true but I wouldn't know where to put that. Also, this code here should be changed anyways long term. I'm thinking of something like `getScripts("engine.*.settings.*)`.

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryManager.java
plata: That's the idea. However, it's not a simple change so I didn't want to do it now.

phoenicis-repository/src/main/java/org/phoenicis/repository/dto/ApplicationDTO.java
plata: Again not sure where to put it. Also: this problem has not been introduced by this PR. In general, I agree that the current implementation is not the nicest one though.

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryManager.java
qparis: In the same class. The idea is just to give a explainatory name to the pattern

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
```diff
@@ -53,7 +53,7 @@ public void getSetting(String engineId, String settingId, Consumer<EngineSetting
         final InteractiveScriptSession interactiveScriptSession = scriptInterpreter.createInteractiveSession();
 
         interactiveScriptSession.eval(
-                "include([\"engines\", \"" + engineId + "\", \"settings\", \"" + settingId + "\"]); new Setting();",
+                "include(\"engines." + engineId + ".settings." + settingId + "\"); new Setting();",
```
qparis: I’m not sure that it is secure. Have you checked that we cannot inject some code?

phoenicis-repository/src/main/java/org/phoenicis/repository/types/ClasspathRepository.java
```diff
@@ -331,9 +331,9 @@ private ScriptDTO buildScript(String typeId, String categoryId, String applicati
 
             if (StringUtils.isBlank(scriptDTOBuilder.getId())) {
                 if (!StringUtils.isBlank(scriptDTOBuilder.getScriptName())) {
-                    scriptDTOBuilder.withId(scriptDTOBuilder.getScriptName().replaceAll("[^a-zA-Z0-9_]", ""));
+                    scriptDTOBuilder.withId(scriptDTOBuilder.getScriptName().replaceAll("[^a-zA-Z0-9_.]", ""));
                 } else {
-                    scriptDTOBuilder.withId(scriptFileName.replaceAll("[^a-zA-Z0-9_]", ""));
+                    scriptDTOBuilder.withId(scriptFileName.replaceAll("[^a-zA-Z0-9_.]", ""));
```
qparis: why not using ID_REGEX?

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
madoar: That is a good point. The question is whether we can make the code secure at all?
I mean the previous implementation had the same problem in that you could inject code via the `engineId` and `settingId` parameters, right?

qparis: This is true. We can make it secure I think

phoenicis-repository/src/main/java/org/phoenicis/repository/RepositoryManager.java
plata: @qparis in this case we would still have the issue that the pattern is repeated in several classes.

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
plata: Is this something which should be handled in this PR? I agree that we should do it but I don't think it belongs here.

phoenicis-repository/src/main/java/org/phoenicis/repository/types/ClasspathRepository.java
plata: Where shall I put this?

qparis: It seems to exist in the same file

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
qparis: Not necessarely, but we should not forget it. No need to sign the script if we keep this

phoenicis-repository/src/main/java/org/phoenicis/repository/types/ClasspathRepository.java
plata: Updated.

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
plata: Feel free to open an issue for it.

plata:
@qparis I think I've resolved the relevant remarks. Can we merge this?

qparis:
Yes

PR #1836, opened 2019-02-15
Refactor the `LibrarySidebar` class with MVC in mind

madoar:
This PR refactors the `LibrarySidebar` class with MVC in mind. For this the previous `LibrarySidebar` class is split into a `LibrarySidebar` and `LibrarySidebarSkin` class.


madoar:
I am not sure what the second `Queued` Travis Job is about, but according to Travis the build passes so I will merge this now.

PR #1841, opened 2019-02-17
Add pulseaudio permission to flatpak

### LINKED ISSUE #1840: Audio not working in flatpak installation
### Expected behavior
Audio should work.

### Actual behavior
Audio of any kind does not work.

### Steps to reproduce
1. Install Phoenicis using flatpak
2. Install a new application / create a new container
3. Engine Tools / Configure Wine
4. Go to Audio-Tab

Wine shows:
```
Selected driver: winealsa.drv
```
but sound (Test Sound) isn't working.
I tried changing the backend by setting the corresponding registry-key to pulse, but now wine shows:
```
Selected driver: (None)
```
Do you maybe need to include libpulse in flatpak somehow?

### Comments
When running Phoenicis 5.0-alpha.2 directly, outside of flatpak, audio works just fine (It is automatically choosing the pulse backend).

### System information
* software version (git revision) of phoenicis and scripts repositories
> Newest from flathub
* used operating system
> openSUSE Tumbleweed
* Engine version
> Wine Staging 64bit 4.0-rc4 & 4.1

seijikun:
Pulseaudio socket is needed to support audio output/input in flatpak installation.
This fixes #1840

### DIFF
--- a/phoenicis-dist/src/flatpak/org.phoenicis.playonlinux.yml
+++ b/phoenicis-dist/src/flatpak/org.phoenicis.playonlinux.yml
@@ -37,6 +37,7 @@ finish-args:
   - --allow=multiarch
   - --share=ipc
   - --socket=x11
+  - --socket=pulseaudio
   - --share=network
   - --device=dri
 modules:


plata:
We will need the same in the flathub repo. I'm not sure about the best strategy to do this. Shall we always update the two scripts in parallel or only occasionally push all changes to flathub?

madoar:
I would suggest to occasionally push all changes to flathub. More precise I think we should push the changes to flathub after we create a new release version. This ensures that we don't accidentally change the configuration on flathub for a change done to our development branch.

plata:
This would mean that we can never bugfix flatpak issues like this one without releasing a new Phoenicis version.

madoar:
I guess there should always be exceptions, I'm just talking about our standard approach :)

Often changes to the flatpak configuration are only required because of a change in Phoenicis itself. Such configuration changes should only be pushed to flathub if we create a new release version.

plata:
> Often changes to the flatpak configuration are only required because of a change in Phoenicis itself. Such configuration changes should only be pushed to flathub if we create a new release version.

I agree to this part.

So: shall we add the last two flatpak fixes to flathub now or not?

PR #1843, opened 2019-02-17
Improve applications components

madoar:
This PR introduces some new utility methods which can be used together with `Binding` and `Property` objects:
- `flatMap`
- `mutableMap`

In addition the PR adds a new utility class for `BooleanBinding` objects called `BooleanBindings`.

I'm still thinking about whether we should merge this PR at all, because on one hand the suggested extensions to the binding and property utility classes are useful, but on the other hand they are hard to debug and add an additional complexity layer.

@qparis @plata what do you think?

Currently this PR is still WIP and should not be merged.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/application/control/ApplicationSidebar.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/application/control/ApplicationSidebar.java
@@ -2,10 +2,14 @@ package org.phoenicis.javafx.components.application.control;
 
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.ObjectProperty;
+import javafx.beans.property.SimpleObjectProperty;
 import javafx.collections.ObservableList;
 import org.phoenicis.javafx.components.application.skin.ApplicationSidebarSkin;
 import org.phoenicis.javafx.components.common.control.ExtendedSidebarBase;
 import org.phoenicis.javafx.components.common.widgets.utils.ListWidgetType;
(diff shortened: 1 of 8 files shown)


plata:
To be honest, these custom bindings in general are a bit suspicious to me. Why does nobody else seem to need them? Are we maybe doing something wrong?

madoar:
> Why does nobody else seem to need them?

I don't think that nobody else needs or uses such bindings. I know of at least two other libraries that provide similar and comparable custom bindings:
- https://github.com/lestard/advanced-bindings
- https://github.com/TomasMikula/EasyBind

The main issue with the previous mentioned libraries is that I'm not sure how heavily they are maintained anymore, especially now that JavaFX has been decoupled from Java itself. Therefore I prefer the approach that we implement the custom bindings we need ourselves.

> Are we maybe doing something wrong?

I don't think that we're doing something wrong. There are always multiple solutions for the same problems when programming. One example would be the usage of Streams vs. for loops. You can't say that one or the other is better or wrong, it's simply a question of preferences and your use-case.

plata:
Should we make a new repo for the bindings then?

madoar:
Yes long-term, but it is too early now in my opinion. I first want to create all utility functions we require. Ideally I also want to add some tests before we move them to a new repository.

plata:
Ok, then we should ask for @qparis opinion.

madoar:
I'm not sure anymore whether we need all in this PR proposed utility functions anymore.
Therefore I will close this PR for now.

If we later on discover that we need some of the functions proposed in this PR we can then copy them to a new PR.

PR #1857, opened 2019-02-23
Replace Nashorn with GraalVM

plata:


plata:
@madoar I don't know how to replace the `ScriptObjectMirror` by `Value` while still using our `InteractiveScriptSession` functional interface. Do you have an idea?

madoar:
Can you give me an example where you face the problems?

plata:
E.g. in `ShortcutRunner`.

madoar:
My IntelliJ doesn't seem to be able to find the `polyglot` package (for `import org.graalvm.polyglot.Value;`).
Do you have an idea?

madoar:
With the import working it should be working with:

```java
    public void run(ShortcutDTO shortcutDTO, List<String> arguments, Consumer<Exception> errorCallback) {

        interactiveScriptSession.eval("include([\"engines\", \"wine\", \"shortcuts\", \"reader\"]);",
                ignored -> interactiveScriptSession.eval("new ShortcutReader()", output -> {
                    final Value shortcutReader = (Value) output;

                    shortcutReader.invokeMember("of", shortcutDTO);
                    shortcutReader.invokeMember("run", arguments);
                }, errorCallback), errorCallback);
    }
```

plata:
I had tried this before already but I got: PolyglotMap cannot be cast to PolyglotValue. I think you only get `Value` if you use `Context`.

madoar:
After pulling the newest changes from `upstream` I got my branch working and did some tests.
You're right I stumbled across the same error like you (the one with `PolyglotMap` and `Value`).

After debugging the code a little I identified the location of the problem. The problem is caused by the "cast" in: https://github.com/graalvm/graaljs/blob/0385428b1c2f2cc4212a85612435c2ea0f105488/graal-js/src/com.oracle.truffle.js.scriptengine/src/com/oracle/truffle/js/scriptengine/GraalJSScriptEngine.java#L205

As far as I found out `polyglotContext.eval(source)` returns a `Value` object, which is the then converted to a `PolyglotMap` object by the `.as(Object.class)` call.

I think that the best solution is to change our graal `ScriptInterpreter` to use a graal `Context` instead of a `ScriptEngine`.

plata:
Currently, I get
```
TypeError: INVOKE on JavaObject[com.oracle.truffle.js.javaadapters.org.phoenicis.engines.Engine@4c66c7 (com.oracle.truffle.js.javaadapters.org.phoenicis.engines.Engine)] failed due to: java.lang.ClassCastException: Cannot convert '(0) []'(language: JavaScript, type: Array) to Java type 'java.lang.String': Unsupported target type.
```
during an installation.

madoar:
Does this occur when "casting" the `Value` to `Installer` or when accessing one of its methods?
Which application are you executing?

plata:
Notepad++. The installation starts just fine, install the correct Wine version but then it crashes. I'm assuming that it happens in `Wine#run()` but no idea how I could debug this.

plata:
I can confirm my `Wine#run()` theory.

madoar:
I just got:

```
Exception in thread "pool-3-thread-3" java.lang.ClassCastException: class org.graalvm.polyglot.Value cannot be cast to class java.util.Map (org.graalvm.polyglot.Value is in module org.graalvm.sdk of loader 'app'; java.util.Map is in module java.base of loader 'bootstrap')
	at phoenicis.engines@5.0-SNAPSHOT/org.phoenicis.engines.EngineSettingsManager.lambda$fetchAvailableEngineSettings$1(EngineSettingsManager.java:75)
	at phoenicis.scripts@5.0-SNAPSHOT/org.phoenicis.scripts.engine.PhoenicisInteractiveScriptSession.eval(PhoenicisInteractiveScriptSession.java:34)
	at phoenicis.scripts@5.0-SNAPSHOT/org.phoenicis.scripts.interpreter.BackgroundScriptInterpreter.lambda$createInteractiveSession$1(BackgroundScriptInterpreter.java:43)
```

when removing an installed application and container

plata:
`EngineSettingsManager` l.75. Not quite sure how to handle this.

madoar:
About your problem:
I have no clue, where exactly the problem is located. The error message indicates that a javascript `array` is passed to the Java method where a `String` is required.

The problem for me is that the responsible script code for this stuff spans 3+ files, which include both the engine "implementation" and "object" code. After following two or three invocations I don't know anymore if the `run` invocation now targets the "implementation" object, the "object" object or maybe even the Java interface?

plata:
I added some debug output but everything looked ok. The call hierarchy is not that difficult really: `InstallerScript` → `Wine(Object)#run` → Java `Engine` interface → `Wine(Implementation)#run`. The problem occurs when the Java interface is called from `Wine(Object)#run` because this is the place where we go from Javascript to Java.

plata:
I found it. It's a bug in scripts. I will create a PR.

plata:
@qparis could you have a look at the packaging script? We need some more command line parameters. See 
```
--add-modules=ALL-MODULE-PATH --module-path=phoenicis-dist/target/lib -XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCI -classpath phoenicis-dist/target/lib --upgrade-module-path=phoenicis-dist/target/lib/compiler.jar
```

plata:
I think we're ready. Apart from the packaging script. Unfortunately, I cannot update that because I cannot test .deb here.

madoar:
@plata alternatively we set the currently used `PhoenicisScriptEngine` back to nashorn and merge this now. This allows us to change both the script and Phoenicis application code to use interfaces for their interoperability faster. In addition this gives us more time to test Graal before we decide it is stable enough to finally migrate.

plata:
Honestly, I don't really see the advantage of this. From what I've tested, everything works. If there's a bug, it's better to use Graal on master and find it.

madoar:
There are multiple reasons for my suggestion:
- it may still take quite a while until the graal related dependencies are added to the deb and osx installers
- graal.js is still not fully released. It is to be expected that graal.js itself still contains a number of bugs, which may lead to problems in Phoenicis

I'm in favor of merging graal.js as soon as possible to make it better testable. In addition I am inclined to give the interface approach a try, which makes the most sense in my opinion after this PR is merged :)

plata:
> it may still take quite a while until the graal related dependencies are added to the deb and osx installers

Why? Shouldn't it be included by the packaging script? Of course, if that's not possible, I agree with you.

> graal.js is still not fully released. It is to be expected that graal.js itself still contains a number of bugs, which may lead to problems in Phoenicis

Wouldn't we still have this issue because we are using `Value`?

madoar:
I don't know how easy it is to introduce the required configuration into the packaging script, is it enough to add the parameters to the execution command? Are there additional OS related steps which need to be performed? Therefore my assumption that it may take a while until we can merge this.

You're right, we can't continue to use nashorn as the default engine for this PR because of the changes to `Value`.

plata:
Let's wait for @qparis. He should know more about the packaging .

madoar:
Agreed!

plata:
@qparis could you have a look at the packaging script so we can merge this? We need some more command line parameters. See 
```
--add-modules=ALL-MODULE-PATH --module-path=phoenicis-dist/target/lib -XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCI -classpath phoenicis-dist/target/lib --upgrade-module-path=phoenicis-dist/target/lib/compiler.jar
```

madoar:
@qparis what is your opinion on this? :)

qparis:
Have you tested the package or does it need to be done?

plata:
Still needs to be done. I do not have Ubuntu here so I cannot test the .deb creation.

madoar:
It seems there are some good news regarding this PR: It seems like GraalVM 19 was "just" released!
I believe that after we've "fixed" the build script(s) and after we've updated to graal dependencies to the release version we are ready to merge this PR.

plata:
I've also seen the update but on Maven central the latest version seems to be 1.0.0-rc16.

plata:
Sorry the latest changes where not meant to go to this branch.

madoar:
> @qparis could you have a look at the packaging script so we can merge this? We need some more command line parameters. See
>
>```
>   --add-modules=ALL-MODULE-PATH --module-path=phoenicis-dist/target/lib -XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCI -classpath phoenicis-dist/target/lib --upgrade-module-path=phoenicis-dist/target/lib/compiler.jar
>   ```

I think setting the module path will be quite easy. For this it should be enough to copy the corresponding statements to:

https://github.com/PhoenicisOrg/phoenicis/blob/98dd045ea8eb803ba48f781294436ab1eed00e46/phoenicis-dist/src/scripts/phoenicis-create-package.sh#L37-L38

Applying the other options could be harder to add I think. @plata have you tried some tests whether adding the remaining options to `PHOENICIS_JPACKAGER_ARGUMENTS` works?

qparis:
Yep, should be enough

madoar:
@qparis I've done some experiments. Long story short:
I couldn't get the packager running with the additional parameters. 
I've tried adding them with `--java-options` to tell the packager that the arguments need to be passed to the runtime but I got an `Illegal argument` error.

I've just pushed #1950 with my results, any ideas?

qparis:
I don't remember. I need to try it again. 
Does jpackager work without the extra arguments?

madoar:
No, I get basically the same error when removing `--java-options`

plata:
@madoar I couldn't test (no Ubuntu here).

madoar:
You can try with Gitpod

plata:
Right. I didn't think about that option.

docs/_docs/Developers/intellij-idea.md
```diff
@@ -18,7 +18,7 @@ add: JavaFXApplication with:
 * main class: org.phoenicis.javafx.JavaFXApplication
 * VM options: 
 ```
---add-modules=ALL-MODULE-PATH --module-path=phoenicis-dist/target/lib
+--add-modules=ALL-MODULE-PATH --module-path=phoenicis-dist/target/lib -XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCI -classpath phoenicis-dist/target/lib --upgrade-module-path=phoenicis-dist/target/lib/compiler.jar
```
qparis: I don't understand why -classpath would be required for GraalVM and not other dependencies. This might cause package problems

plata: Can you try if it works for you without it?

plata:
I cannot run it anymore:
```
org.phoenicis.scripts.interpreter.ScriptException: org.graalvm.polyglot.PolyglotException: SyntaxError: Error initializing language 'regex' using class 'com.oracle.truffle.regex.RegexLanguage'.
Unnamed:26:9 Expected ; but found )
        });
         ^
Unnamed:27:4 Expected eof but found }
    };
    ^
```

madoar:
Ok, I've experimented a bit. I can reproduce the exception(s). Interestingly they only seem to occur when executing Phoenicis from the IDE. If I install it via a built `deb` archive I don't get the error message and the application starts like it should.

I believe the reason for the exception is a missing parameter which needs to be passed to the JVM. Somehow we set this parameter/option correctly inside the `deb` installer but at least I don't do that inside the IDE.

plata:
That's an interesting find. Nevertheless, I think it's something which should be fixed before we can merge this.

madoar:
I've searched some more but still can't find the reason for the error messages. The only think I found out is that the errors are thrown when parsing a javascript regex. For example an exception is thrown for the lines:

https://github.com/PhoenicisOrg/phoenicis/blob/2ee4666d4625b188c590be9f0b600bbea9f5597e/phoenicis-scripts/src/main/resources/org/phoenicis/scripts/nashorn/builtins/utils.js#L26-L27

I can't find any significant differences between my IntelliJ setup and the options the installer script sets.

qparis:
If you replace ALL_MODULE_PATH with jdk.crypto.ec,java.base,javafx.base,javafx.web,javafx.media,javafx.graphics,javafx.controls,java.naming,java.sql,java.scripting,jdk.scripting.nashorn,jdk.internal.vm.ci,jdk.internal.vm.compiler,org.graalvm.truffle,java.desktop,java.prefs,java.xml 

it should work

qparis:
Actually java.desktop,java.prefs,java.xml are not needed 

@madoar Seems that some module are not detected automatically with ALL-MODULE-PATH

qparis:
@plata Feel free to merge whenever you think we are stable enough. 
Next step is to get rid of nashorn compatibility

plata:
Do we need to update the IntelliJ IDEA documentation and and the flatpak? If so, what's the final list of required modules?

qparis:
The list of package is the same than than in the packaging script. 
I have no idea concerning flatpak. Do you pass the module list somewhere?

qparis:
@plata Can this be merged?

plata:
I'm doing a quick test then I will merge, yes.

qparis:
Does it mean that it will not compile on Ubuntu?

plata:
This is just for the IntelliJ IDEA documentation. Phoenicis itself is compiled with Java 11 already since https://github.com/PhoenicisOrg/phoenicis/commit/b1b1ad5c8ebfdc6dd815e5f43559cf5c0f16f9ea.

qparis:
Ok

plata:
Apart from that: @madoar is using Ubuntu or Linux Mint so we should be fine.

PR #1864, opened 2019-02-26
Refactor the ContainerPanel class with MVC in mind

madoar:
This PR refactors the `ContainerPanel` class with MVC in mind. For this the `ContainerPanel` class has been replaced with a `ContainerInformationPanel` component, which needs to be wrapped in a `DetailsPanel` object.


madoar:
In addition to the comments in #1865:

It is interesting to note that the bindings don't seem to be invalidated (triggered) by:
https://github.com/PhoenicisOrg/phoenicis/blob/41de4372050185e467eeb634be12ff571917d9be/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java#L134-L137

I'm not sure why this is the case. Therefore to "update" the tabs in the container details panel, the container needs to be reselected in the list widget.

plata:
> Therefore to "update" the tabs in the container details panel, the container needs to be reselected in the list widget.

Is this a new issue introduced in this PR?

madoar:
No this hasn't been introduced by this PR. The source for this behavior is the way we pass new information to the `ContainerInformationPanel`, which only occurs when selecting an entry in the list widget.

madoar:
The newest commit fixes #1865.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java
```diff
@@ -51,18 +47,12 @@
...
 
     private boolean firstViewSelection = true;
 
-    public ContainersController(ContainersView containersView,
-            ContainersManager containersManager,
-            ContainerEngineController containerEngineController,
-            RepositoryManager repositoryManager,
-            EngineSettingsManager engineSettingsManager,
-            VerbsManager verbsManager,
+    public ContainersController(ContainersView containersView, ContainersManager containersManager,
```
plata: I think this was more readable before.

madoar: Ok, I've reverted the constructor parameter style for the class.

PR #1866, opened 2019-03-01
Add a new TabIndicator component to inform the user about changes

### LINKED ISSUE #1786: Implement notification system
Trigger notification on:
- installation started
- app install finished
- container created

The notification could be shown like:
notification

Is this something for JavaFX only or can we implement it in the respective modules directly?

see also #1781

madoar:
This PR closes #1786 by implementing a `TabIndicator` component, which informs the user about changes to a tab, which occurred in the time interval since the user visited the tab the last time.

Screenshots:
grafik
grafik
grafik

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsView.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsView.java
@@ -21,7 +21,7 @@ package org.phoenicis.javafx.views.mainwindow.installations;
 import javafx.application.Platform;
 import javafx.beans.Observable;
 import javafx.beans.binding.Bindings;
-import javafx.beans.binding.StringBinding;
+import javafx.beans.binding.IntegerBinding;
 import javafx.beans.property.ObjectProperty;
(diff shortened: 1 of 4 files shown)


madoar:
Currently a `TabIndicator` is used to show the total number of currently active installations.

plata:
That's cool! Maybe the number should be white?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/mainwindow/installations/InstallationsView.java
```diff
@@ -87,6 +89,18 @@ public InstallationsView(ThemeManager themeManager, JavaFxSettingsManager javaFx
 
         this.activeInstallations = createInstallationListWidget();
 
+        // a binding containing the number of currently active installations
+        final StringBinding openInstallations = Bindings.createStringBinding(
+                () -> Integer.toString(this.activeInstallations.getElements().size()),
```
plata: What will happen if there are >9 elements? Can we maybe show a "+" or "i" instead of 10? I assume that more than one character will not fit in the circle.

madoar:
grafik
grafik

PR #1870, opened 2019-03-02
Rename Nashorn related classes to "PhoenicisScript..." 

plata:
In a next step, I want to add an enum for `PhoenicisScriptEngineFactory` which allows you to select either Nashorn or Graal.


plata:
As the PR is vs the graalvm branch, we have not Codacy or Travis results.

madoar:
Does this allow us to use `ScriptEngine` for nashorn and `Context` (i.e. `Polyglot`) for graal?

plata:
I don't think so. I thought we can just use a `ScriptEngine` with "graal.js" as described in https://github.com/graalvm/graaljs/blob/master/docs/user/NashornMigrationGuide.md.

madoar:
I think it would be nice to be able to use any underlying API a scripting solution provides.
Often it is the case that the dedicated solution (i.e. polyglot for graal.js) is more powerful than the Java Scripting API.

plata:
I get your point but does it make sense to do this now? Shouldn't we go for a rather simple transition first? If we need other features later on, we can always extend it.

madoar:
I don't know

plata:
My first approach was indeed to have an interface which is implemented for nashorn and graal. However, the implementations seemed to be too similar. Therefore I decided to have only one engine, one factory etc.

madoar:
I think this is ok for now, but I think we should open a new issue to change this in the future.

PR #1871, opened 2019-03-02
Add ListConfirmDialog class

### LINKED ISSUE #1822: Show confirm dialog when installing verbs via the container details panel
It would be useful to show a confirmation dialog when the user wants to install one or multiple verbs in the container tab. This dialog should contain the following information/components:
- a textual list containing the names of all verbs the user has selected
- a confirm button
- a cancel button

For more details see https://github.com/PhoenicisOrg/phoenicis/pull/1815#issuecomment-460035106

madoar:
This PR:
- renames `ConfirmDialog` to `SimpleConfirmDialog`
- adds a new abstract `ConfirmDialog` class
- adds a `ListConfirmDialog` class
- improves the dialog classes

This PR closes #1822.

Screenshot:
grafik

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/dialogs/ConfirmDialog.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/dialogs/ConfirmDialog.java
@@ -4,16 +4,13 @@ import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.scene.control.Alert;
 import javafx.scene.control.ButtonType;
-import javafx.scene.control.Label;
-import javafx.scene.layout.Region;
-import javafx.stage.Window;
 
 import java.util.Optional;
 
 /**
- * A confirm dialog with two callbacks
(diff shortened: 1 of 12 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/dialogs/ConfirmDialog.java
```diff
@@ -4,16 +4,13 @@
...
-import javafx.stage.Window;
 
 import java.util.Optional;
 
 /**
- * A confirm dialog with two callbacks
+ * An abstract class for confirm dialogs with two callbacks
  */
-public class ConfirmDialog extends Alert {
+public abstract class ConfirmDialog extends Alert {
```
plata: Wouldn't it make sense to already have title, owner etc in this class?

madoar: This information is part of `Alert`.

plata: Ah I messed up with the builder.

PR #1877, opened 2019-03-04
Developer documentation: add coding rules

### LINKED ISSUE #1564: Change the code specification for the keyword `this`
Currently we use the keyword `this` basically whenever we can.
I don't think that this is a good idea, because it distracts from the important lines in the code where it is absolutely required.
In addition it makes the code longer and therefore decreases the readability.

For more details see the discussion in #1556

plata:
fixes #1564

### DIFF
--- a/docs/_docs/Developers/coding-rules.md
+++ b/docs/_docs/Developers/coding-rules.md
@@ -1,0 +1,13 @@
+---
+title: "Coding rules"
+category: Developers
+order: 10
+toc: true
+---
+
+### naming conventions
+* Use camel case.
+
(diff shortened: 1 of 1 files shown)


docs/_docs/Developers/coding-rules.md
```diff
@@ -0,0 +1,13 @@
+---
+title: "Coding rules"
+category: Developers
+order: 10
+toc: true
+---
+
+### naming conventions
+* Use camel case.
```
madoar: Where and for what?

PR #1880, opened 2019-03-04
Remove the ApplicationFilter class

madoar:
This PR removes the `ApplicationFilter` class and replaces it by a set of interfaces.


plata:
Why did you choose this approach?

I'm not sure if it's really a good idea that e.g. `ApplicationInformationPanel` implements `ScriptFilter` because
- it does not fulfill an "is a" relationship
- it violates "composition over inheritance"

madoar:
> Why did you choose this approach?

I had the idea for this approach because of my experiences with #1843 (which would be obsolete with this PR). 

The "problem" I wanted to fix both PRs is the location issue of the filter function and the corresponding properties in combination with working bindings. In #1843 a `mutableMap` is required to update the filter relevant properties when the filter itself changes. This method is very complex and its errors are hard to detect and understand. On the other hand the predicate bindings for `FilteredList` objects require that the same properties are used for the filter function and as the dependencies for `Bindings.createObjectBinding(...)`, otherwise either the filter functions on old values or the binding isn't invalidated correctly.

To solve or better to circumvent these problems I created this PR, which circumvents the whole issue by removing the `ApplicationFilter` class and moving the filter functions and their corresponding properties to the components where they are used.

> it does not fulfill an "is a" relationship

When is a class a filter? In my opinion when it provides filter functionality. A class/component automatically provides filter functionality if it implements one of the interfaces.

> it violates "composition over inheritance"

Java doesn't support inheritance over multiple classes. Therefore a Mixin approach is the next best implementation approach in this case.

plata:
I would like to hear @qparis opinion about this.

madoar:
@qparis what do you think about this PR? :)

PR #1881, opened 2019-03-05
Use Graal Context instead of ScriptEngine 

plata:
With this PR, it's possible to run and stop applications from the library using graal.


madoar:
Found the reason :D
What is working:

```java
        this.context = Context.newBuilder("js")
                .option("js.nashorn-compat", "true")
                .allowHostAccess(true).build();
```

I'll also open a PR in a minute.

plata:
I know. I have this here already XD

PR #1882, opened 2019-03-05
Add phoenicis script engine interface

madoar:
This PR:
- makes `PhoenicisScriptEngine` an interface
- adds two specific implementations for `PhoenicisScriptEngine` (one based on the Java Scripting API and another one for Polyglot engines)

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/JSAScriptEngine.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/JSAScriptEngine.java
@@ -1,17 +1,28 @@
 package org.phoenicis.scripts.engine;
 
 import javax.script.ScriptEngine;
+import javax.script.ScriptEngineManager;
 import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Consumer;
 
+/**
+ * A {@link PhoenicisScriptEngine} wrapping around a {@link ScriptEngine} object defined by the Java Scripting API
+ */
 public class JSAScriptEngine implements PhoenicisScriptEngine {
     private final ScriptEngine scriptEngine;
     private final List<Consumer<Exception>> errorHandlers = new ArrayList<>();
 
-    public JSAScriptEngine(ScriptEngine scriptEngine) {
-        this.scriptEngine = scriptEngine;
+    /**
+     * Constructor
+     *
+     * @param engine The name of the engine
+     */
+    public JSAScriptEngine(String engine) {
+        super();
+
+        this.scriptEngine = new ScriptEngineManager().getEngineByName(engine);
     }
 
     public void eval(InputStreamReader inputStreamReader, Consumer<Exception> errorCallback) {
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/JSAScriptEngine.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/JSAScriptEngine.java
@@ -1,0 +1,66 @@
+package org.phoenicis.scripts.engine;
+
+import javax.script.ScriptEngine;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+public class JSAScriptEngine implements PhoenicisScriptEngine {
+    private final ScriptEngine scriptEngine;
+    private final List<Consumer<Exception>> errorHandlers = new ArrayList<>();
+
+    public JSAScriptEngine(ScriptEngine scriptEngine) {
+        this.scriptEngine = scriptEngine;
+    }
+
+    public void eval(InputStreamReader inputStreamReader, Consumer<Exception> errorCallback) {
+        try {
+            this.scriptEngine.eval(inputStreamReader);
+        } catch (Exception e) {
+            handleError(errorCallback, e);
+        }
+    }
+
+    public void eval(String script, Runnable doneCallback, Consumer<Exception> errorCallback) {
+        try {
+            this.scriptEngine.eval(script);
+            doneCallback.run();
+        } catch (Exception e) {
+            handleError(errorCallback, e);
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


plata:
I guess this should be vs graalvm branch?

plata:
Honestly, I don't think it makes sense to keep the `ScriptEngine` implementation. It probably won't work anyways because you cannot cast it's result to `Value`, can you?

madoar:
For cleanliness purposes I would propose to make `PhoenicisScriptEngine` an interface. This allows us to add other (experimental) implementations (like graal now) easier.

About the adaptability of one engine to another:
This depends on how we want to use the `PhoenicisScriptEngine` in the future.
Long term I would recommend only working with Java interfaces, which we specify in Phoenicis (i.e. `Installer`, `Verb`, `Engine`, ...). Then `PhoenicisScriptEngine` only needs to provide methods which allow such an access, for example:

```java
<E> E evalAndReturn(String script, Consumer<Exception> errorCallback, Class<E> type);
```

The instantiation of `E` could then be done engine specific. (for graal either via `Value#as` or via a [Proxy](https://docs.oracle.com/javase/8/docs/technotes/guides/reflection/proxy.html))

plata:
Ok, that makes sense. Then I will close my PR. Our next step should be to implement your proposal for the library tab (after merging this).

madoar:
Btw I agree that we should remove the nashorn implementation for the graal branch.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/JSAScriptEngine.java
```diff
@@ -0,0 +1,66 @@
+package org.phoenicis.scripts.engine;
+
+import javax.script.ScriptEngine;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+public class JSAScriptEngine implements PhoenicisScriptEngine {
```
plata: What is "JSA"?

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/ScriptEngineType.java
```diff
@@ -1,17 +1,34 @@
...
+        }
+    };
 
     private final String name;
 
     ScriptEngineType(String name) {
         this.name = name;
     }
 
+    public abstract PhoenicisScriptEngine createScriptEngine();
```
plata: Do we want this in an enum? I'd rather put this functionality in the factory.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/JSAScriptEngine.java
madoar: JSA = Java Scripting API
See also https://docs.oracle.com/javase/8/docs/technotes/guides/scripting/prog_guide/api.html

plata: Ok, let me put it differently: Is the abbreviation common enough to know what this class is about (it wasn't for me at least)? At minimum, we should have a comment.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/ScriptEngineType.java
madoar: I also thought about putting this in the factory, my issue was where?
The problem was that `createEngine` in `PhoenicisScriptEngineFactory` does so much more than simply creating an engine object. It also initializes it, i.e. it adds the injectors.

After adding the method to the `ScriptEngineType` I also discovered that it was quite a nice approach, because you can specify the required parameters in the `ScriptEngineType` objects.
For example we could add a graal python engine easily by adding:

```java
    PYTHON("graal.python") {
        @Override
        public PhoenicisScriptEngine createScriptEngine() {
            return new PolyglotScriptEngine("python", Map.of());
        }
    };
```

This approach doesn't require extra `switch case` constructs or anything similar.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/JSAScriptEngine.java
madoar: Done

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/ScriptEngineType.java
plata: It's just that you wouldn't expect functionality in an enum. Also, you basically moved the factory to the enum. I would really prefer `switch case` in the factory. Adding injectors in the factory is perfectly fine because its part of building the object (logically - that it doesn't happen in the constructor is only implementation details).

madoar: I'm fine with moving this functionality to a different place. I just don't know where and how to do this in an elegant manner.

I would suggest to merge this as is for now and change it later on. 
I mean we still need to change the `PhoenicisScriptEngine` implementations multiple times. So we have enough time to think of a better place to instantiate them.

plata: Ok.

plata:
Regarding the next steps: Can you add a new page to the developers documentation explaining the concept of communicating with Javascript only via Java interfaces? Then we can finish the graal implementation without this and tackle it later on. Otherwise we will add too much to the graal topic.

madoar:
Yes I agree, just to summarize:

- in the first step we change Phoenicis to use Graal.js instead of Nashorn with the least required changes
- in a second (or maybe third?) step we change the the way Phoenicis communicates with the scripts to an interface approach

Additionally I would like to introduce more model programming patterns (for example [classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes)) to the `script` repository after we've finished step 1.

plata:
Exactly. And not to forget and for future developers, the concept should be explained in the documentation.

PR #1883, opened 2019-03-05
Remove ScriptObjectMirror 

plata:


phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
```diff
@@ -132,23 +132,22 @@ public void deleteContainer(ContainerDTO container, Consumer<Exception> errorCal
...
-         * });
-         */
+        categories.stream().flatMap(shortcutCategoryDTO -> shortcutCategoryDTO.getShortcuts().stream())
+                .forEach(shortcutDTO -> {
+                    final InteractiveScriptSession interactiveScriptSession = this.scriptInterpreter
+                            .createInteractiveSession();
+                    interactiveScriptSession.eval(
+                            "include(\"engines." + engineId + ".shortcuts.reader\");",
+                            ignored -> interactiveScriptSession.eval("new ShortcutReader()", output -> {
+                                final org.graalvm.polyglot.Value shortcutReader = (org.graalvm.polyglot.Value) output;
```
madoar: No import?

plata: No, because `Value` is already defined by `import org.springframework.beans.factory.annotation.Value;`.

PR #1885, opened 2019-03-05
Add interface communication approach to documentation

madoar:
This PR adds the interface communication approach discussed in #1882 to the documentation

### DIFF
--- a/docs/_docs/Developers/script-interaction.md
+++ b/docs/_docs/Developers/script-interaction.md
@@ -1,0 +1,36 @@
+---
+title: "Interaction between Phoenicis and its Scripts"
+category: Developers
+order: 10
+toc: true
+---
+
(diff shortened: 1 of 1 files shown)


docs/_docs/Developers/script-interaction.md
```diff
@@ -0,0 +1,36 @@
+---
+title: "Interaction between Phoenicis and its Scripts"
+category: Developers
+order: 10
```
plata: Don't we have 10 already?

madoar: Yes, and we have 9 two times...
I don't really understand the `order` field.
I thought when adding the tenth file I'm on the right side with the order `10`.

plata: Ok... Normally we should have each number only once.

PR #1886, opened 2019-03-05
Fix retrieval of engine settings

plata:


plata:
I can see that the engine settings are retrieved correctly from Javascript but they do not show up in the container tab.

madoar:
Even if you wait for a bit?
Is the engine settings tab empty then?

plata:
Apparently, the list in the map still contains `com.oracle.truffle.js.javaadapters.org.phoenicis.engines.EngineSetting` instead of `org.phoenicis.engines.EngineSetting`. With my latest update, the "Engine settings" tab shows up but when I click it, it explodes.

madoar:
Can't we add a method to the `Engine` interface and its implementation to return all `EngineSetting` objects? This should make it easier for graal to convert the objects.

plata:
I'm not sure. At the moment, the engine knows nothing about its settings/plugins/verbs.

madoar:
The approach to find this information is to search the corresponding subfolders of the engine, right?
Can't we implement this in JS as well?

plata:
Can we traverse the repository in JS?

madoar:
We can traverse the folders in the filesystem in JS

plata:
That's not enough. The repository can be merged.

madoar:
Then if we can make the repository accessible from JS we can traverse it in JS.

madoar:
In my opinion it should be possible to fetch the engine settings from the engine JS implementation. 
Later on during the merging process the engine settings then need to be merged separately.

madoar:
An additional question is then:
Can we also perform the merging of multiple repositories in JS? 
Maybe using a filesystem approach... 
Where we create a new repository, which contains all files from the merged repositories.
This also allows for the fastest access times.

plata:
Sounds a bit complicated to me. It should be possible to get the correct type from the adapter somehow.

madoar:
There is another fix:
```java
output -> {
   for (String engineName : engineNames) {
      Value engineSettings = ((Value)output).getMember(engineName);
      List<EngineSetting> settings = new ArrayList<>();
      for (int index = 0; index < engineSettings.getArraySize(); index++) {
         settings.add(engineSettings.getArrayElement(index).as(EngineSetting.class));
      }
      settingsMapJava.put(engineName, settings);
   }
}
```

plata:
Did some more investigation. It's an error in the regedit script.

PR #1887, opened 2019-03-05
Restructure phoenicis-scripts module

madoar:
This PR restructures the `phoenicis-scripts` module and removes the Java Scripting API `PhoenicisScriptEngine` implementations.


plata:
Maybe we should keep the JSA implementation for performance measurements.

madoar:
We can't use it if we switch to `Value`.

plata:
We can use it again once we finished the Java interfaces, right?

plata:
It might also be helpful to test if the Java interface implementation is complete.

madoar:
The big question will be how we can convert the different types of scripts into their corresponding interfaces. If we decide to keep both `PhoenicisScriptEngine` implementations we need to provide solutions for the different script types for both `PhoenicisScriptEngine` implementations. I can think of at least 3 types of scripts that require different handling:

### Executable scripts
Executable scripts are directly executed. They don't return an object at the end, which can be used as as the implementation body for the interface. Our installation scripts belong to this category.

To convert an executable script to an interface, the execution of the script needs to be delayed until the trigger method in the interface has been called (for example the `run` method in the `Installer` interface)

### Prototype scripts
Prototype scripts return a prototype/class object at the end. Our engine scripts (e.g. wine) belong to this category.

To convert a prototype script to an interface, the script can be executed directly. The returned prototype/class object then needs to be instantiated on the Java side. Finally the resulting instance needs to be cast into the desired Java interface.

### Object scripts
Object scripts return an object, which can be directly cast to a Java interface.
Typically scripts returning a json object belong to this category.

plata:
We could decide to make phoenicis-scripts very defined for our special needs. So instead of generic `evalPrototype`: `installApp`, `getEngine` etc. Not sure if it makes sense but for me that's not the point here. The only question is if we want to have a solution which is agnostic of the used scripting implementation (e.g. nashorn or graal) or not. And I think we decided for yes.

madoar:
I thought about adding a `Map<Class<?>, JavaScriptConverter>` to `PolyglotScriptEngine`, which is responsible for converting a given script to the corresponding Java type. I guess this will lead to three specific implementations for `PolyglotScriptEngine`. If we decide to keep `JSAScriptEngine` we will need to implement the same for the Java Scripting Engine or we add empty stubs.

I'm not against retaining the `JSAScriptEngine`, I mainly want to talk about the additional work we need to perform to maintain them later on.

plata:
Let's keep it for now. If it's too much work to maintain it later, we can always delete it.

madoar:
Ok, I will readd the `JSAScriptEngine` class. 

Additionally we shouldn't forget that nashorn is only compatible with a quite old JavaScript standard. This means that a lot of my suggested changes (i.e. JavaScript classes, the filesystem module) don't work using nashorn.

plata:
That's true. Like I said: for me, nashorn is more a proof tester that our interface is good. I'm not really planning to use it anymore.

madoar:
According to IntelliJ it was unused

plata:
Ok.

plata:
I assume it's not called from scripts?

madoar:
I just searched for it in the repository, no match found.

Issue #1800, opened 2019-01-24
Restructure Styles/Themes


madoar:
Currently our themes consist each of two files: `main.less` and `description.less`.
Each of these files contain a flat hierarchy of style class definitions.
This structure is has multiple disadvantages:
- the `main.less` file tends to be quite long and hard to read
- the flat hierarchy makes it difficult to:
    - define dependencies
    - define slight deviations between similar components
    - see which definitions are used together on the same panel/view

Therefore I propose the following changes to our themes:

## Split the Themes in more files
First I propose to split our themes in more than the previously mentioned two files.
For example, I suggest we add a `buttons.less` file, which contains all button definitions, for example definitions like:

```less
.aboutButton {
  -fx-background-image: url('/org/phoenicis/javafx/themes/standard/icons/mainwindow/settings/about.png');
}
```

## Differentiate in **global** and **local** style classes
To solve the issues with the flat hierarchy I propose to differentiate the style classes in **global** and **local** style classes.

### Global style classes
Global style classes are basically what we currently have.
Global style classes are defined at the top most scope in our `main.less` file.
They are applied application wide and therefore look the same in all panels/view.

An example for global style classes are the `.button` and `.label` style classes.
These classes are automatically applied to all `Button` and `Label` objects by JavaFX. 
In addition all buttons and labels should normally look the same independent of where they are shown.

### Local style classes
Local style classes are only applied to individual panels/views.
Local style classes should be defined using a style class nested hierarchy, which specifies to which components in which tab/panel/view the style class is applied.

An example for a local style class are the grid definitions.
Not all grids should look and behave the same. Their look should be customized depending on the panel they are added to (i.e. the list widgets should look different compared to the grids in used in the settings tab)

plata:
Would you compile all less files in one css again? How do other projects handle this?

madoar:
No we don't need to transpile all less files in one css file.
I don't know how other Java projects handle this. In other languages and with other frameworks (for example angular) it is normal to split the stylesheets in multiple files depending on the components to which they are applied.

plata:
Wouldn't this make loading the css in Java quite difficult?

madoar:
I don't think so. The css is loaded via this line:
https://github.com/PhoenicisOrg/phoenicis/blob/bfe4cee2d8b6a7b679dc0d44c2aafb259c0a453c/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/ThemeManager.java#L122

[...]

Issue #1491, opened 2018-11-27
Change css class names from camelCase to kebab-case


madoar:
Currently we use css class names such as `sidebarInside`, which are written in camelCase.
I think it would be better readable and more in line with normal css to use kebab-case (e.g. `sidebar-inside`) for css class names.

plata:
Do you have a source which supports this style for CSS?

madoar:
What do you mean with source?
I couldn't find a styleguide that specified how css class names should look like.
It's just that in my opinion and experience css class names are normally written using kebab-case.
An example for a project that uses kebab-case css class names is [bootstrap](https://getbootstrap.com/).

ovari:
https://google.github.io/styleguide/htmlcssguide.html

https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style

plata:
To quote the google style guide:

> Do not concatenate words and abbreviations in selectors by any characters (including none at all) other than hyphens, in order to improve understanding and scannability.

So, who ever would like to do this: go ahead.

madoar:
Seems like I skipped over this section of the google styleguide yesterday...

PR #1897, opened 2019-03-07
Extract button css to buttons.less

madoar:
This PR addresses the first task of #1800 by splitting the themes into a `buttons.less` and `main.less` file.
I'm currently not sure whether we should use a flat or nested hierarchy for the button css.

### DIFF
--- a/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/buttons.less
+++ b/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/buttons.less
@@ -1,0 +1,147 @@
+/*******************************************************/
+/************************ icons ************************/
+/*******************************************************/
+
+.sidebar {
+  .searchBox {
+    .searchCleanButton {
+      -fx-graphic: url('/org/phoenicis/javafx/themes/breezeDark/icons/mainwindow/general/edit-clear.png');
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


plata:
Wouldn't it be better to group by tab/functionality? In other applications, it makes a lot of sense to group by type but in Phoenicis, many styles depend more on the place where they are used than the type of the control.

madoar:
Yes, this was what I meant by the second step of #1800 :D
The idea was to first extract the images, this makes it easier in my opinion to change the visuals of our buttons. Afterwards I want to overhaul the remaining css according to their functionality/tab.
It is likely that this second step will be combined with #1491.

I'm still open for ideas, because I currently only have a very loose plan, which will likely require some later refinement.

plata:
If we use general.less, apps.less, containers.less etc. later, why introduce buttons.less now?

madoar:
It seems like I misunderstood you. So basically you want to split the main.less file into multiple files according to the targeted functionality?

Either way I think we should continue the discussion in #1800 and let this PR stay open until we have decided how to continue.

madoar:
I don't think that we should go with the approach in this PR anymore. I agree with @plata that it is a better idea to split the styles according to the targeted functionalities, e.g.:

- library
- applications
- container
- engines
- ...

plata:
Do we have an issue to track that we want to do this?

PR #1903, opened 2019-03-12
Refactor the ContainersView class with MVC in mind

madoar:
This PR refactors the `ContainersView` class with MVC in mind. For this the previous `ContainersView` class is split into a `ContainersFeaturePanel` and `ContainersFeaturePanelSkin` class. The new `ContainersFeaturePanel` component is then wrapped inside a Tab object.


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainersFeaturePanel.java
```diff
@@ -0,0 +1,250 @@
...
+            });
+        }
+    }
+
+    public ContainersFilter getFilter() {
+        return filter.get();
+    }
+
+    public ObjectProperty<ContainersFilter> filterProperty() {
+        return filter;
```
plata: `this` (also below)

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainersFeaturePanel.java
```diff
@@ -0,0 +1,250 @@
...
+import org.phoenicis.repository.dto.ApplicationDTO;
+
+import java.awt.*;
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+
+import static org.phoenicis.configuration.localisation.Localisation.tr;
+
+public class ContainersFeaturePanel extends FeaturePanel<ContainersFeaturePanel, ContainersFeaturePanelSkin> {
```
plata: add some Javadoc

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/skin/ContainersFeaturePanelSkin.java
```diff
@@ -0,0 +1,231 @@
...
+import org.phoenicis.javafx.settings.JavaFxSettingsManager;
+import org.phoenicis.javafx.utils.ObjectBindings;
+import org.phoenicis.javafx.utils.StringBindings;
+import org.phoenicis.repository.dto.ApplicationDTO;
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+
+public class ContainersFeaturePanelSkin extends FeaturePanelSkin<ContainersFeaturePanel, ContainersFeaturePanelSkin> {
```
plata: Javadoc

madoar:
This PR still has an open issue:

Normally when successfully completing an installation, a new container should be automatically added to the containers tab. This doesn't work currently in this PR (it works on the master). 

The reason this previously worked is because 

https://github.com/PhoenicisOrg/phoenicis/blob/d096e49c17258deed0f664ef31b4aca4db7fb5db/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java#L108

is called inside the callback:

https://github.com/PhoenicisOrg/phoenicis/blob/d096e49c17258deed0f664ef31b4aca4db7fb5db/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/containers/ContainersController.java#L68

As far as I know this leads to the refreshing of the containers whenever the containers tab is opened.

A more clean approach in my opinion would be to add a second callback to the `SetupUiJavaFXImplementation` class, called `onComplete` or `onSuccess`, which is called when the setup wizard has been completed successfully. This callback should then update the containers in the containers tab.

@plata @qparis what do you think?

plata:
Depends if an installation is really the only case that a new container can be created. I'm not sure if that's really true. I think (at least for now) we should keep the old logic.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainersFeaturePanel.java
madoar: Done

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainersFeaturePanel.java
madoar: Done

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/skin/ContainersFeaturePanelSkin.java
madoar: Done

madoar:
Ok. I've changed it now so that the containers are reloaded whenever the containers tab is opened.
I'll open a new issue to discuss other strategies to reload the containers after this PR has been merged.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainersFeaturePanel.java
```diff
@@ -0,0 +1,293 @@
...
+    public ObservableMap<String, ApplicationDTO> getEngineTools() {
+        return engineTools;
+    }
+
+    public ContainerDTO getSelectedContainer() {
+        return this.selectedContainer.get();
+    }
+
+    public ObjectProperty<ContainerDTO> selectedContainerProperty() {
+        return selectedContainer;
```
plata: `this`

madoar: Done

Issue #1915, opened 2019-04-06
Integrate gitpod?


plata:
It might potentially make it easier for people to get started/contribute. However, in a first quick try I wasn't able to compile.

What do you think? Shall we consider this or not?

plata:
Ok, you need to set `JAVA_HOME` to compile:
```
export JAVA_HOME=/home/gitpod/.sdkman/candidates/java/11.0.2-open/

plata:
Running JavaFX doesn't work right away but it looks like it's possible to work around that limitation:
https://medium.com/gitpod/developing-native-ui-applications-in-gitpod-15af2967c24e

madoar:
`gitpod` integration sounds like an interesting idea. In addition to running JavaFX I see another possible issue. A core part when playing around with the Phoenicis code is running scripts. Is this possible (allowed) with `gitpod`?

plata:
@madoar I just tried installing Notepad++. Worked until the shortcut creation (it cannot find the exe). Maybe we should debug that.

Kreyren:
QA: Assuming gitpot beeing already integrated based on project README.md -> close issue?

plata:
Yes, can be closed.

PR #1917, opened 2019-04-08
Try Gitpod Config

meysholdt:
Hello! 

I'm Moritz, I work on Gitpod. I noticed your questions in https://github.com/PhoenicisOrg/phoenicis/issues/1915, so I thought I should try myself if I get your project working in Gitpod. 

To try this PR in Gitpod, follow this link:
https://gitpod.io/#https://github.com/meysholdt/phoenicis

This PR adds a dockerfile and a `.gitpod.yml`. The Dockerfile is useful to install all tools the projects needs. The docker image is based on Ubuntu, so you can use `RUN apt-get update && apt-get install myfavoritetool`. Right now, I'm only setting the proper Java version via SDKMan.

This also means that all scripts that work under linux should just work in Gitpod. It's just Ubuntu Linux. 

With this config `mvn install` passes just fine.

The Docker image extends `gitpod/workspace-full-vnc`, so yes, there is support for running UI applications in Gitpod. I see you already found [this example](https://medium.com/gitpod/developing-native-ui-applications-in-gitpod-15af2967c24e). However, with your project I got stuck with the following exception:
```
java.lang.UnsupportedOperationException: Internal Error
        at com.sun.glass.ui.gtk.GtkApplication.lambda$new$6(GtkApplication.java:189)
        at java.base/java.security.AccessController.doPrivileged(Native Method)
        at com.sun.glass.ui.gtk.GtkApplication.<init>(GtkApplication.java:171)
        at com.sun.glass.ui.gtk.GtkPlatformFactory.createApplication(GtkPlatformFactory.java:41)
        at com.sun.glass.ui.Application.run(Application.java:144)
        at com.sun.javafx.tk.quantum.QuantumToolkit.startup(QuantumToolkit.java:258)
        at com.sun.javafx.application.PlatformImpl.startup(PlatformImpl.java:269)
        at com.sun.javafx.application.PlatformImpl.startup(PlatformImpl.java:158)
        at com.sun.javafx.application.LauncherImpl.startToolkit(LauncherImpl.java:658)
        at com.sun.javafx.application.LauncherImpl.launchApplication1(LauncherImpl.java:678)
        at com.sun.javafx.application.LauncherImpl.lambda$launchApplication$2(LauncherImpl.java:195)
```

My guess is that Java/GTK fail to talk to the X window server for some reason, but I'm not sure and I didn't analyse it further.  Maybe you already know this kind of issue.

This is how it looks like:
image

Let me know what you think.


plata:
Thanks a ton for the input. I cannot recall seeing that exception before. @madoar do you? We will definitely have to sort this out before we can merge this PR but if we can manage to do so, I'd be happy offering gitpod to the people who want to participate in Phoenicis.

madoar:
@meysholdt yes thanks alot for your input!

@plata no I haven't seen this error before. I guess it is possible that some dependencies are missing on the docker container, which are expected by JavaFX. I think it would be interesting to see if swing UIs work, so that we can exclude Java itself as the culprit. If the error only occurs with JavaFX I think we should open an issue on the OpenJFX github repository to let them know.

madoar:
Another idea would be to try with JavaFX 12.

plata:
OpenJFX is installed by Maven. It should not be required to install it with apt.

meysholdt:
Interesting. For some reason it works with the version installed by `apt`, but not with the one installed by maven. I'm considering to include this in the base image: https://github.com/gitpod-io/workspace-images/pull/81. Should I not?

plata:
I think we can use it for now if it works. Probably worth investigating why the Maven dependency is not sufficient in this Docker container nevertheless.

madoar:
@meysholdt I think it is fine to add it in the base image. Still I believe it is a good idea to investiage why it doesn't work with the maven installation method! The issue is that I think quite a lot of applications want to deliver JavaFX with their applications. 

Delivering JavaFX ensures for example, that the correct version of JavaFX for an application is installed. If JavaFX is via apt it could lead to unintended behavior, like an application working with gitpod but not in a real world scenario where JavaFX is not installed via apt.

meysholdt:
It's working now :)

image

There may be one last hickup: The VNC view disconnects when it's idle for a while, but hitting the "reload" icon in the Gitpod-Preview-View brings it back.

I did install the OpenJFX libraries in the base image, because I don't want anybody to run into this issue...
find them here: https://github.com/gitpod-io/workspace-images/blob/8167b14ee279141d8ef91583687648bcc73f2819/full-vnc/Dockerfile#L7

Regarding the question why pulling OpenJFX in via maven dependency didn't work: I didn't investigate this here, but in past projects I know there were cases where jars related to JavaFX had to be on the boot-classpath of the JVM, not the regular classpath. I don't know if that's the issue here but it could be an explanation.

I noticed that your build spends several minutes downloading JARs and compiling Java code. There is absolutely not need to wait for this, because Gitpod can do it before you open a worksapce. You'd need to install the Gitpod App (https://github.com/apps/gitpod-io) on the repo, and it will build the workspace for branches/PRs on-git-push (just like Travis/Jenkins) so that workspaces are ready-to-be-used when you need them. 

From my side this PR is ready-for-merging. Let me know what you think.

plata:
@meysholdt I've definitely planned to use the app :+1: 

Do you think we should consider the remark from Codacy?
One other thing: Is it possible to zoom in the VNC window?

Otherwise I think we can merge this if @madoar and @qparis agree.

meysholdt:
hey @plata,

sorry for the late response. 

Codacity: yes, I think it's best to ignore the remark for now until https://github.com/gitpod-io/workspace-images/issues/84 is fixed.

Zoom in VNC: There is a feature-request for screen-resize https://github.com/gitpod-io/gitpod/issues/303, but I know that technically zoom can also be possible. Please file a feature request if screen-resize does not fulfill your use-case and screen-zoom would be valuable for you. what works nicely today is to open the VNC-view in a new browser-tab (instead of opening it in Gitpod's preview), because the tab has sufficient screen-estate so that zooming/resizing may not be necessary.

plata:
Ok, then I will merge this. I guess screen-resize should be sufficient for us.

PR #1922, opened 2019-04-18
Fix missing category icons

### LINKED ISSUE #1874: Category icons broken
Screenshot:
grafik

## Most likely reason:
After the script ids have changed, the generated css looks like this:
```css
#applications.accessoriesButton{
-fx-background-image: url('file:~/.Phoenicis/cache/git-331995911/Applications/Accessories/icon.png');
}
```

Correct would be most likely:
```css
#accessoriesButton{
-fx-background-image: url('file:~/.Phoenicis/cache/git-331995911/Applications/Accessories/icon.png');
}
```

plata:
Use full category ID in kebab-case for category button ID. This avoids potential naming conflicts e.g. between applications and engines (there could be something like `applications.wine` and `engines.wine`).

fixes #1874

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/engine/skin/EnginesSidebarToggleGroupSkin.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/engine/skin/EnginesSidebarToggleGroupSkin.java
@@ -36,7 +36,7 @@ public class EnginesSidebarToggleGroupSkin extends
     protected ToggleButton convertToToggleButton(EngineCategoryDTO category) {
         final ToggleButton categoryButton = createSidebarToggleButton(category.getName());
 
-        categoryButton.setId(String.format("%sButton", category.getName().toLowerCase()));
+        categoryButton.setId(String.format("%s-button", category.getName().toLowerCase().replace('.', '-')));
         categoryButton.setOnAction(event -> getControl().setSelectedElement(category));
 
         return categoryButton;
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/engine/skin/EnginesSidebarToggleGroupSkin.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/engine/skin/EnginesSidebarToggleGroupSkin.java
@@ -8,7 +8,7 @@ import org.phoenicis.javafx.components.common.skin.SidebarToggleGroupBaseSkin;
 import java.util.Optional;
 
 /**
- * A {@link SidebarToggleGroupBaseSkin} implementation class used inside the {@link EnginesSidebar}
+ * A {@link SidebarToggleGroupBaseSkin} implementation class used inside the {@link EngineSidebar}
  */
 public class EnginesSidebarToggleGroupSkin extends
         SidebarToggleGroupBaseSkin<EngineCategoryDTO, EnginesSidebarToggleGroup, EnginesSidebarToggleGroupSkin> {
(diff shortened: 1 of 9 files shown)


plata:
@madoar I'm not sure if this could be related to your MVC refactoring but I had to explicitly trigger the callbacks otherwise the library/application category icons only showed up after opening the engines tab (which probably simply forces a repository load).

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
```diff
@@ -566,39 +566,39 @@
     -fx-background-image: url('/org/phoenicis/javafx/themes/breezeDark/icons/mainwindow/apps/all.png');
 }
 
-#customButton {
+#applications-custom-button {
```
madoar: What about the other themes? Don't they require similar changes?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/engine/skin/EnginesSidebarToggleGroupSkin.java
```diff
@@ -36,7 +36,7 @@ public EnginesSidebarToggleGroupSkin(EnginesSidebarToggleGroup control) {
     protected ToggleButton convertToToggleButton(EngineCategoryDTO category) {
         final ToggleButton categoryButton = createSidebarToggleButton(category.getName());
 
-        categoryButton.setId(String.format("%sButton", category.getName().toLowerCase()));
+        categoryButton.setId(String.format("%s-button", category.getName().toLowerCase().replace('.', '-')));
```
madoar: I think it would be useful to add a method `buttonize`, which is responsible for the whole conversion of a string:

```java
public String buttonize(String pattern, String input) {
   return String.format(pattern, input.toLowerCase().replace('.', '-'))
}
```

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibrarySidebarToggleGroupSkin.java
```diff
@@ -43,7 +43,9 @@ public LibrarySidebarToggleGroupSkin(LibrarySidebarToggleGroup control) {
     protected ToggleButton convertToToggleButton(ShortcutCategoryDTO category) {
         final ToggleButton categoryButton = createSidebarToggleButton(category.getName());
 
-        categoryButton.setId(String.format("%sButton", category.getId().toLowerCase()));
+        categoryButton.setId(String.format("applications-%s-button", category.getId().toLowerCase())); // TODO: store
```
madoar: Please move the comment to the line above

madoar:
> @madoar I'm not sure if this could be related to your MVC refactoring but I had to explicitly trigger the callbacks otherwise the library/application category icons only showed up after opening the engines tab (which probably simply forces a repository load).

This is possible. When refactoring the views/tabs I basically inserted the refactored views/tabs into the old view with a lot of "glue". The idea was to rework the content loading mechanism after all views/tabs have been refactored.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
plata: They use the default icons from what I saw but you can check again.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/engine/skin/EnginesSidebarToggleGroupSkin.java
plata: I thought about this but in which class?

madoar: If this is only required in the sidebars I would recommend `SidebarToggleGroupBaseSkin`.
I also think it makes sense to make the method static.

phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/breezeDark/main.less
madoar: I think the standard theme requires similar changes:

https://github.com/PhoenicisOrg/phoenicis/blob/d470dc6ac00e79d629048ca03a7d62be5be84e4e/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/standard/main.less#L868-L870

https://github.com/PhoenicisOrg/phoenicis/blob/d470dc6ac00e79d629048ca03a7d62be5be84e4e/phoenicis-javafx/src/main/resources/org/phoenicis/javafx/themes/standard/main.less#L918-L921

PR #1923, opened 2019-04-18
Avoid concurrent modification of DefaultRepositoryManager

plata:
- callbacks should not be iterated while a new callback is added
- new repositories should not be added concurrently


plata:
I'll merge this then you can do whatever changes you like in your PR.

madoar:
@qparis how do you interact with the `RepositoryManager` on the website? Which methods do you call?

Issue #1960, opened 2019-06-03
Java 12 support


qparis:

plata:
I guess this requires the GraalVM pr to be merged, right?

qparis:
Actually maybe not... However, it requires we use jpackage from here: https://jdk.java.net/jpackage/

Ironically, I could not make it work with graalvm yet: I'm getting this error at runtime java.lang.module.FindException: Module org.graalvm.truffle not found, required by jdk.internal.vm.compiler

qparis:
Oh I see, nashorn is no longer available... So yeah you are right

qparis:
In fact it seems that it is still supported

madoar:
@qparis on https://jdk.java.net/jpackage/ it is mentioned that the build is based on an incomplete version of JDK 13. Does this mean that the built installer contains a Java 13 JRE?

qparis:
Yep, I have tried to bundle a Java 12 JRE and it does not seem to work. With Java 13, it seems to work great though. I suggest we keep the two scripts (Java 11 and Java 13) to anticipate the future release of Jpackage

plata:
Travis is building with OpenJDK 12, so this issue is fixed. If something else should be done (e.g. regarding packaging with Java 12), please open a new issue for that.

PR #1961, opened 2019-06-04
Travis: build with JDK 12

plata:
- add OpenJDK 11
- remove Oracle JDK support
- use OpenJDK 12 for Flatpak

see #1960

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,16 @@ jobs:
       os: linux
       dist: trusty
       jdk: oraclejdk11
+    - stage: Build + test
+      name: Linux OpenJDK 12
+      os: linux
+      dist: trusty
+      jdk: openjdk12
+    - stage: Build + test
+      name: Linux Oracle JDK 12
+      os: linux
+      dist: trusty
+      jdk: oraclejdk12
 
     # packages
     - stage: Packages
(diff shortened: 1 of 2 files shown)


plata:
Do you see any reason to maintain Oracle JDK support?

madoar:
If it is easy to retain I would suggest retaining it. I agree that OpenJDK is the most important one to test. So if it takes too much effort to maintain OracleJDK support in travis we can remove.

plata:
At least it didn't seem like OracleJDK 12 is available/works.

plata:
@qparis are you ok with removing OracleJDK support as well?

qparis:
Yes

qparis:
Did you manage to run the project with OpenJDK 12?

plata:
Didn't try that yet. Just wanted to prepare the Travis part so that we have the build already. You are right, I should have checked for the flatpak. So blame me if I broke it.

qparis:
I suspect it is broken with Java 12

plata:
Ok, I will revert the flatpak change and separate it to a new pr. Should have done this anyways...

plata:
However, this also indicates that our tests are not really good enough.

PR #1964, opened 2019-06-04
Flatpak: use OpenJDK 12

plata:
TODO: test if it works

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
       name: Flatpak
       os: linux
       dist: xenial
-      jdk: openjdk11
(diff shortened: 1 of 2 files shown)


qparis:
Now it should work

madoar:
@plata any news about this? :)

plata:
I didn't check. However, https://github.com/flathub/org.freedesktop.Sdk.Extension.openjdk is using Java 13 by now. I think it makes sense to stick to the LTS versions for the flatpak. Therefore I would suggest to close this.

madoar:
Ok then let us close this

PR #1976, opened 2019-06-07
Add winbind and freetype to the .deb

ImperatorS79:
They are not provided by the runtime due to crash.


plata:
Looks ok for me. @qparis please check.

PR #1984, opened 2019-06-10
Gitpod: add freetype

qparis:


plata:
Have you tried running a Wine app in Gitpod?

qparis:
Cannot. The workspace needs to be built

plata:
Ok. Let's see if it works then.

qparis:
This was a mistake...

PR #2002, opened 2019-06-15
Simplify the Installer Scripts

madoar:
This PR allows executing simplified installer scripts like proposed in https://github.com/PhoenicisOrg/scripts/pull/1011.

The new installer script execution process is as followed:
1) set the necessary environment variables for the installer script (`TYPE_ID`, `CATEGORY_ID`, ...)
2) return a JavaScript object implementing the Installer interface from the script (i.e. the object needs to contain a `go` method) 
3) execute the `go` method of the returned JavaScript object

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/Installer.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/Installer.java
@@ -7,5 +7,5 @@ public interface Installer {
     /**
      * runs the installer
      */
-    void run();
+    void go();
 }
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
@@ -1,12 +1,12 @@
 package org.phoenicis.scripts.engine.injectors;
 
-import org.phoenicis.scripts.engine.implementation.PhoenicisScriptEngine;
 import org.phoenicis.scripts.interpreter.ScriptException;
 import org.phoenicis.scripts.interpreter.ScriptFetcher;
+import org.phoenicis.scripts.engine.implementation.PhoenicisScriptEngine;
 
 import java.util.HashSet;
 import java.util.Set;
-import java.util.function.Function;
+import java.util.function.Consumer;
 
 /**
  * Injects Include() function into a Script Engine
@@ -22,18 +22,16 @@ public class IncludeInjector implements EngineInjector {
     public void injectInto(PhoenicisScriptEngine phoenicisScriptEngine) {
         final Set<String> includedScripts = new HashSet<>();
 
-        phoenicisScriptEngine.put("include", (Function<String, Object>) argument -> {
+        phoenicisScriptEngine.put("include", (Consumer<String>) argument -> {
             final String script = scriptFetcher.getScript(argument);
             if (script == null) {
                 throwException(new ScriptException(argument + " is not found"));
             }
 
             if (includedScripts.add(argument)) {
-                return phoenicisScriptEngine.evalAndReturn("//# sourceURL=" + argument + "\n" + script,
+                phoenicisScriptEngine.eval("//# sourceURL=" + argument + "\n" + script,
                         this::throwException);
             }
-
-            return null;
         }, this::throwException);
     }
 }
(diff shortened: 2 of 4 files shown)


phoenicis-scripts/src/main/java/org/phoenicis/scripts/Installer.java
```diff
@@ -7,5 +7,5 @@
     /**
      * runs the installer
      */
-    void run();
+    void go();
```
plata: Why do you prefer `go` over `run`?

madoar: I don't prefer `go` over `run`. Using `go` just required less changes because it has been defined in the quick script implementations. So if we want to continue using `run` we would need to change the quick scripts to provide a `run` method instead of a `go` method.

plata: Ok.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
```diff
@@ -22,16 +22,18 @@ public IncludeInjector(ScriptFetcher scriptFetcher) {
     public void injectInto(PhoenicisScriptEngine phoenicisScriptEngine) {
         final Set<String> includedScripts = new HashSet<>();
 
-        phoenicisScriptEngine.put("include", (Consumer<String>) argument -> {
+        phoenicisScriptEngine.put("include", (Function<String, Object>) argument -> {
```
plata: Will this work in cases where the script e.g. also defines variables? Does it affect somehow the way `include` is used in scripts?

madoar: > Will this work in cases where the script e.g. also defines variables? Does it affect somehow the way include is used in scripts?

What do you mean by variables? This change only changes the semantics of the `include(<scriptId>)` command. To be more exact it changes the signature of the method from:

```java
void include(String scriptId);
```

to

```java
Object include(String scriptId);
```

All our existing scripts shouldn't be affected by this because they don't use the "result" of the `include` command.

plata: I was thinking of something like:
```Javascript
var a = "hello world";

new OnlineInstallerScript() {
    ...
}
```
Is `a` still accessible if I include this?

madoar: It should be. I mean the script is still fully executed. The only change is, that the result of the last "line" is returned as the result of the `include(script)` command.

plata: Does this mean that the following would not work?
```Javascript
class MyVerb { ... }
var a = "test";
```

madoar: Yes

plata: Should we use an explicit `return`? Like `return new OnlineInstallerScript`? I think this would be clearer then deliberately usin the last statement. Also it's in line with similar stuff in other script languages e.g. Lua modules.

madoar: `return` doesn't work, because it is not a function from which we are returning.
The equivalent would be an `export`, but when using an `export` you need to use a named `import` when using the exported expression. We can't use `import` because it requires a name of the imported expression, which we don't have/know.

plata: I see. It probably would make sense to split the change for `include` and the installer simplification as we cannot merge the installer simplification before all scripts are done. It would also allow us to test all types (plugins, verbs etc.) first.

madoar: See #2004

plata:
I would like to see this working for the engine implementation as well. After that I think we can go for it.

PR #2004, opened 2019-06-16
Change Signature of the "include" Command to return an Object

madoar:
This PR changes the signature of the `include` operation to return an object. In other words it changes:

```java
void include(String scriptId);
```

to:

```java
Object include(String scriptId);
```

The returned value of the `include` command is the result of the last executed command inside the referenced script.


plata:
@qparis please check as well.

madoar:
@qparis can you take a short look at this PR. This PR is basically the prerequisite for #2006, #2007, #2009 and #2010.

madoar:
@qparis can you take a short look at this PR :)

PR #2006, opened 2019-06-18
Simplify the Verb Scripts

madoar:
This PR is the Java counterpart to https://github.com/PhoenicisOrg/scripts/pull/1015.

Additionally I've slightly modified the `Verb` interface to indicate that a `ScriptException` can be thrown from inside. Strangely I wasn't able to catch the `InterruptedException` which is thrown when an installer script is canceled...


qparis:
Looks good to me

plata:
To be merged after #2004.

PR #2014, opened 2019-06-21
Remove libncurses, libfreetype, libbsd0 from .deb and add libc6:i386

ImperatorS79:
Needs https://github.com/PhoenicisOrg/phoenicis-winebuild/pull/98 and https://github.com/PhoenicisOrg/scripts/pull/1019.


ImperatorS79:
libc6:i386 is needed because without it one cannot launch 32 bits apps.

PR #2018, opened 2019-06-23
GraalVM sandbox implementation

qparis:


phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PhoenicisSandbox.java
```diff
@@ -0,0 +1,33 @@
+package org.phoenicis.scripts.engine.implementation;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PhoenicisSandbox {
```
madoar: Please add some JavaDoc

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PhoenicisSandbox.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public boolean isSafe(String identifier) {
+        LOGGER.debug("Loading {} in javascript context", identifier);
+        if (identifier.startsWith("org.phoenicis")) {
+            return true;
+        }
+
+        if (identifier.startsWith("java.lang")) {
+            // FIXME: This should be more fine-tuned later
+            // Contains process builder
```
madoar: I don't think that we should make `ProcessBuilder` directly visible. This has multiple reasons:
- the `ProcessBuilder` API is designed for Java usage
- to start a process and wait until it finished takes quite some operations using the `ProcessBuilder` API
- the `ProcessBuilder` API exposes **all** applications you've installed on your system to the outside. A malicious script could make use of this

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PolyglotScriptEngine.java
```diff
@@ -33,16 +35,19 @@
...
      * Constructor
      *
+     * @param phoenicisSandbox a Phoenicis Sandbox bean
      * @param language The language name
      * @param options A map of options for the Polyglot context
      */
-    public PolyglotScriptEngine(String language, Map<String, String> options) {
+    public PolyglotScriptEngine(PhoenicisSandbox phoenicisSandbox, String language, Map<String, String> options) {
         super();
+        this.phoenicisSandbox = phoenicisSandbox;
```
madoar: This should be moved down a few lines to not break the formatting

qparis: mvn formatter is ok with this

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PhoenicisSandbox.java
qparis: Yes, this is why there is a fixme. However, if I block ProcessBuilder, it will break the scripts. 
This class is just a first proof of concept ; it needs to be tuned later

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PolyglotScriptEngine.java
madoar: I meant it breaks the (my) reading flow :)
I normally have a single line break between `super()` and the later assignments, followed by:
1) the constructor parameter assignments
2) the generated values assignments (often based on constructor parameters)

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PhoenicisSandbox.java
madoar: Then it seems I misunderstood you. I thought you wanted to retain the ProcessBuilder but exclude other classes

qparis: Yep, we are going step by step.
At this end, I'd like to introduce a permissions attribute in the script settings, so that we can manage permission properly

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PolyglotScriptEngine.java
qparis: Done

plata:
Please add a description to the PR about how this works, what's the final goal etc.

PR #2019, opened 2019-06-23
Refactor the "FileUtilities" Class

madoar:
This PR refactors the `FileUtilities` class. This contains a change to the method signatures to accept Strings instead of File objects. In addition some additional file access methods are now forwarded from JS to Java to make the accessibility-split more clean.

### DIFF
--- a/phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
+++ b/phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
@@ -40,9 +40,6 @@ import org.springframework.util.CollectionUtils;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
--- a/phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
+++ b/phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
@@ -28,8 +28,8 @@ import org.phoenicis.library.LibraryManager;
 import org.phoenicis.library.ShortcutManager;
 import org.phoenicis.library.dto.ShortcutCategoryDTO;
 import org.phoenicis.library.dto.ShortcutDTO;
-import org.phoenicis.scripts.session.InteractiveScriptSession;
 import org.phoenicis.scripts.interpreter.ScriptInterpreter;
+import org.phoenicis.scripts.session.InteractiveScriptSession;
 import org.phoenicis.tools.config.CompatibleConfigFileFormatFactory;
 import org.phoenicis.tools.config.ConfigFile;
 import org.phoenicis.tools.files.FileUtilities;
@@ -120,9 +120,7 @@ public class GenericContainersManager implements ContainersManager {
     @Override
     public void deleteContainer(ContainerDTO container, Consumer<ContainerDTO> onSuccess, Consumer<Exception> onError) {
         try {
-            final File containerPath = new File(container.getPath());
-
-            this.fileUtilities.remove(containerPath);
+            this.fileUtilities.remove(container.getPath());
         } catch (IOException e) {
             LOGGER.error("Cannot delete container (" + container.getPath() + ")! Exception: " + e.toString());
             onError.accept(e);
--- a/phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
+++ b/phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
@@ -20,6 +20,7 @@ package org.phoenicis.containers;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.commons.io.FileUtils;
 import org.phoenicis.configuration.security.Safe;
 import org.phoenicis.containers.dto.ContainerCategoryDTO;
 import org.phoenicis.containers.dto.ContainerDTO;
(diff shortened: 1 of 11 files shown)


phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
```diff
@@ -120,9 +120,7 @@ public void fetchContainers(Consumer<List<ContainerCategoryDTO>> onSuccess, Cons
     @Override
     public void deleteContainer(ContainerDTO container, Consumer<ContainerDTO> onSuccess, Consumer<Exception> onError) {
         try {
-            final File containerPath = new File(container.getPath());
-
-            this.fileUtilities.remove(containerPath);
+            this.fileUtilities.remove(container.getPath());
```
madoar: I'm not sure if we should call the functions which we export for ease of use to the JS scripts internally in our main Application. I think it would be cleaner to directly call Java APIs to perform the deletion, or we forward the call somehow to a JS script.

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -176,7 +175,7 @@ public RepositoryDTO fetchInstallableApplications() {
     @Override
     public void onDelete() {
         try {
-            fileUtilities.remove(this.localFolder);
+            fileUtilities.remove(this.localFolder.getAbsolutePath());
```
madoar: See above

madoar: Do we also need to delete the lock file?

phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
plata: I agree.

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
plata: I don't think it's required.

plata:
We should discuss about the structure of our JS API. I think it would be better to have a separate module `phoenicis-api` which contains the complete API instead of spreading it around like now. This is somewhat similar to the remarks from @madoar.
@qparis @madoar what do you think?

madoar:
Isn't `phoenicis-tools` such a module? My main issue with `phoenicis-tools` is that it accepts and returns Java objects (e.g. `File` etc.) instead of JS objects (e.g. `String`, `Array`, etc.).

qparis:
> We should discuss about the structure of our JS API. I think it would be better to have a separate module `phoenicis-api` which contains the complete API instead of spreading it around like now. This is somewhat similar to the remarks from @madoar.
> @qparis @madoar what do you think?

I do not share this point of view. IMHO, the app needs to stay modular: a module = an independant set of feature. Otherwise, a module can grow and grow and grow and grow forever, we should avoid this.

With this way of cutting this, we could imagine later having a system of plugin; a plugin would just be a maven module exposing its own API (and why not, its own JavaFX classes)

plata:
I can understand your point but this will make it really hard to
* figure out which functions are provided by the API
* implement the API outside of Phoenicis

qparis:
Annotations are made for this :)

plata:
Can we generate an API description based on annotations?

qparis:
Yes, this is how swagger works basically

Zemogiter:
https://github.com/PhoenicisOrg/phoenicis/blob/0c3831d51d2099cb801c002cccbcebeab2186661/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java#L178 this line results in
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/jonasz/Pobrane/phoenicis-master/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java:[178,66] incompatible types: java.lang.String cannot be converted to java.io.File
[INFO] 1 error
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for phoenicis 5.0-SNAPSHOT:
[INFO] 
[INFO] phoenicis .......................................... SUCCESS [  0.948 s]
[INFO] phoenicis-configuration ............................ SUCCESS [  2.120 s]
[INFO] phoenicis-entities ................................. SUCCESS [  0.179 s]
[INFO] phoenicis-multithreading ........................... SUCCESS [  0.332 s]
[INFO] phoenicis-win32 .................................... SUCCESS [  0.558 s]
[INFO] phoenicis-tools .................................... SUCCESS [  1.041 s]
[INFO] phoenicis-repository ............................... FAILURE [  0.631 s]
[INFO] phoenicis-scripts .................................. SKIPPED
[INFO] phoenicis-library .................................. SKIPPED
[INFO] phoenicis-engines .................................. SKIPPED
[INFO] phoenicis-settings ................................. SKIPPED
[INFO] phoenicis-containers ............................... SKIPPED
[INFO] phoenicis-javafx ................................... SKIPPED
[INFO] phoenicis-cli ...................................... SKIPPED
[INFO] phoenicis-dist ..................................... SKIPPED
[INFO] phoenicis-tests .................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  6.009 s
[INFO] Finished at: 2019-06-26T09:08:54+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project phoenicis-repository: Compilation failure
[ERROR] /home/jonasz/Pobrane/phoenicis-master/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java:[178,66] incompatible types: java.lang.String cannot be converted to java.io.File
```

plata:
If that's true, how can it compile in Travis?

madoar:
@Zemogiter this looks like a build error on your end

Zemogiter:
@madoar you are correct. Tried to apply this and and some other PR to my local repo at the same time. That was probably the reason. With just this PR the code compiles normally. I'd say we can merge it.

phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
madoar: Done

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
madoar: Done

PR #2022, opened 2019-06-28
Fix a "TypeError" Exception when deleting a Wine Container

madoar:
This PR changes the called method on `ShortcutReader` from `container()` to `getContainer()` during the deletion of a container. The reason for the change is that `container` has been defined as a getter method which needs to be accessed like a member field, which is semantically wrong.

This PR needs to be merged together with https://github.com/PhoenicisOrg/scripts/pull/1042, which also contains additional information.


plata:
What exactly is the effect of this? Can the pr title be more specific than "bug"? (same for scripts pr)

Issue #2021, opened 2019-06-25
PC Building Simulator script crashes


Zemogiter:
### Expected behavior
After the `.preInstall` step the script should launch steam with the game ID argument and start the game installation
### Actual behavior
Somewhere in that step I get this error:
```
java.lang.IllegalAccessException: access to public member failed: sun.awt.SunToolkit.getScreenSize[Ljava.lang.Object;@2dcb37ed/invokeVirtual, from public Lookup
	at org.graalvm.truffle/com.oracle.truffle.polyglot.HostMethodDesc$SingleMethod$MethodMHImpl.makeMethodHandle(HostMethodDesc.java:318)
	at org.graalvm.truffle/com.oracle.truffle.polyglot.HostMethodDesc$SingleMethod$MHBase.invoke(HostMethodDesc.java:258)
	at org.graalvm.truffle/com.oracle.truffle.polyglot.HostExecuteNode$1.executeImpl(HostExecuteNode.java:776)
	at org.graalvm.truffle/com.oracle.truffle.polyglot.GuestToHostRootNode.execute(GuestToHostRootNode.java:87)
	at <js> :anonymous(Unnamed:23:998-1049)
	at <js> go(engines.wine.quick_script.steam_script:145:7919-7953)
	at <js> run(Unnamed:12-27:353-1192)
	at org.graalvm.sdk/org.graalvm.polyglot.Value.executeVoid(Value.java:358)
	at org.phoenicis.scripts.Installer$$JSJavaAdapter.run(Unknown Source)
	at <js> :program(Unnamed:34:1318-1338)
	at org.graalvm.sdk/org.graalvm.polyglot.Context.eval(Context.java:367)
	at org.phoenicis.scripts.engine.implementation.PolyglotScriptEngine.eval(PolyglotScriptEngine.java:63)
	at org.phoenicis.scripts.interpreter.PhoenicisScriptInterpreter.runScript(PhoenicisScriptInterpreter.java:36)
	at org.phoenicis.scripts.interpreter.BackgroundScriptInterpreter.lambda$runScript$0(BackgroundScriptInterpreter.java:37)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
[...]

PR #2023, opened 2019-06-29
Add ScreenManager

Zemogiter:
Since we shouldn't make direct java calls in javascript I've made this. Untested


phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
```diff
@@ -0,0 +1,33 @@
...
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.tools.system;
+
+import org.phoenicis.configuration.security.Safe;
+import org.phoenicis.entities.OperatingSystem;
+
+public class setVirtualDesktop {
```
madoar: In Java a class needs to be named the same as the file it is contained inside, i.e. `ScreenManager` in this case.

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     *Fetch the screen resolution and sets the virtual desktop to match it
+     *
+     *@return the current screen resolution
+     */
+    public virtualDesktop fetchCurrentScreenResolution() {
```
madoar: `getScreenSize()` returns a `Dimension` object, see also https://docs.oracle.com/javase/9/docs/api/java/awt/Toolkit.html#getScreenSize--

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
```diff
@@ -0,0 +1,33 @@
...
+        return screenSize = java.awt.Toolkit.getDefaultToolkit().getScreenSize();
```
madoar: A `return` statement normally does not contain an assignment, try:
`return java.awt.Toolkit.getDefaultToolkit().getScreenSize();`

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
madoar: You need a `@Safe` annotation before the class definition, i.e.:
```java
@Safe
public class ScreenManager {
   ...
}
```

Adding `@Safe` is one of two things that are required to make this class available from the scripts :)

madoar: In addition to `@Safe` you also need to add a method to https://github.com/PhoenicisOrg/phoenicis/blob/master/phoenicis-tools/src/main/java/org/phoenicis/tools/ToolsConfiguration.java:

```java
    @Bean
    public ScreenManager screenManager() {
        return new ScreenManager();
    }
```

This method is then used to fetch the `ScreenManager` instance when calling `Bean("screenManager")` inside a script.

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.phoenicis.tools.system;
+
+import org.phoenicis.configuration.security.Safe;
+import org.phoenicis.entities.OperatingSystem;
+
+public class setVirtualDesktop {
+    /**
+     *Fetch the screen resolution and sets the virtual desktop to match it
```
madoar: Please add a space between the text and the `*` on the left

madoar:
@Zemogiter can you please link to the issues containing the related discussions?

Zemogiter:
https://github.com/PhoenicisOrg/phoenicis/issues/2021
Affects every script that sets virtual desktop

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
Zemogiter: Should it be converted to text?

madoar: What do you mean by `to text`?

Zemogiter: Something like `java.awt.Toolkit.getDefaultToolkit().getScreenSize().toText()` because a `Dimension` object won't be recorginzed by Wine (at least that's what I think)

madoar: In which format is the screen size needed in the scripts? Is it always the same?

plata: I suggest two methods for x and y resolution returning `int`.

qparis: I suggest creating a Resolution object, may be easier to use after

Zemogiter: Or getting the x and y from an already existing `Dimension` object.

qparis: In general, we try to avoid that because you cannot exactly control what can Dimension do

plata: And we do not want to expose standard Java classes by default. So yes, either the `Resolution` object or x/y to avoid a separate class.

Zemogiter: @madoar the format would be screenWidth x screenHeight, always the same

madoar: I think I prefer a separate class, this reduces the number of necessary calls from JS to Java.

madoar: @Zemogiter you need to change the the line above to `public Dimension fetchCurrentScreenResolution() {` to make it work

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
```diff
@@ -0,0 +1,37 @@
...
+public class ScreenManager {
+    /**
+     * Fetch the screen resolution and sets the virtual desktop to match it
+     *
+     * @return the current screen resolution
+     */
+    public virtualDesktop fetchCurrentScreenResolution() {
+        return getDefaultToolkit().getScreenSize();
+        int width = System.out.println(getScreenSize().width);
+        int height = System.out.println(getScreenSize().height);
```
madoar: It is not allowed to execute code after a `return` statement, because it is never executed, correct is either:

```java
public virtualDesktop fetchCurrentScreenResolution() {
   return getDefaultToolkit().getScreenSize();
}
```

or 

```java
public double getWidth() {
   return getDefaultToolkit().getScreenSize().getWidth();
}

public double getHeight() {
   return getDefaultToolkit().getScreenSize().getHeight();
}
```

madoar:
@plata should we fetch the screen size using AWT or JavaFX? In either case I think it is better to move the class to the javafx module, what do you think?

Zemogiter:
The way I see it we do now have two functions that extract screen height and width. Now we have to turn them into int values and modify the virtual desktop plugin.

plata:
@madoar I can understand your reasoning but actually this functionality has nothing to do with the GUI. So we need this also from Phoenicis CLI. Therefore I'm not sure if it makes sense to move it to JavaFX.

madoar:
That is true I forgot the CLI tool. Still I think using JavaFX is a better choice that AWT. This leads to the question: do we want a JavaFX dependency in the tools module?

qparis:
Hmm that’s a bad idea. This feature will not work on CLI mode. 
I’d rather find another solution to grab the screen resolution

madoar:
That is true... So we're stuck with AWT?
If we continue to use AWT for this I think we should look at https://stackoverflow.com/questions/3680221/how-can-i-get-screen-resolution-in-java to ensure we fetch the screen size of the correct monitor in a multi-monitor setup

qparis:
System calls are acceptable in this situation I think

Zemogiter:
I'm confused by Codacy errors. I already did make explicit statements.

madoar:
@Zemogiter you misunderstood my last comment. I didn't mean to replace the whole code. My suggestion was only to use `GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice()` instead of `Toolkit.getDefaultToolkit().getScreenSize()`:

```java
    public double getScreenWidth() {
        return GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice().getWidth();
    }

     public double getScreenHeight() {
        return GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice().getHeight();
    }
```

madoar:
Please look again at the sample code I edited it because something was missing

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
```diff
@@ -0,0 +1,37 @@
...
+import org.phoenicis.configuration.security.Safe;
+
+@Safe
+public class ScreenManager {
+    /**
+     * Fetch the screen resolution and sets the virtual desktop to match it
+     *
+     * @return the current screen resolution
+     */
+    public int double getScreenWidth() {
```
madoar: only `int` not `int double`

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
```diff
@@ -0,0 +1,37 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.tools.system;
+
+import org.phoenicis.configuration.security.Safe;
```
madoar: You also need to import `GraphicsEnvironment`.
Try `import java.awt.GraphicsEnvironment;`

madoar:
After you've fixed the two open comments you should be able to use the new manager in the scripts via:

```javascript
const width = Bean("screenManager").getScreenWidth();
const height = Bean("screenManager").getScreenHeight();
```

Zemogiter:
Why is Travis complaining?

plata:
Looks like you did not build.

madoar:
@Zemogiter the formatting is not correct. Try `mvn clean compile`.

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
```diff
@@ -0,0 +1,38 @@
...
+import java.awt.GraphicsEnvironment;
+
+@Safe
+public class ScreenManager {
+    /**
+     * Fetch the screen resolution and sets the virtual desktop to match it
+     *
+     * @return the current screen resolution
```
plata: it returns only the width (not the resolution)

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public int getScreenWidth() {
+        return GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice().getDisplayMode().getWidth();
+    }
+
+    public int getScreenHeight() {
```
plata: add Javadoc

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
Zemogiter: You want it to look like this:
```
@return the current screen width
@return the current screen height
```
?

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
Zemogiter: Where?

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
plata:
```Java
/**
 * Fetches the screen width in px
 *
 * @return the current screen width
*/
public int getScreenWidth() {
```

phoenicis-tools/src/main/java/org/phoenicis/tools/system/ScreenManager.java
plata:
```Java
/**
 * Fetches the screen height in px
 *
 * @return the current screen height
*/
public int getScreenHeight() {
```

PR #2025, opened 2019-06-30
Add a "ShortcutReader" interface

madoar:
This PR adds a `ShortcutReader` interface for type safety reasons


phoenicis-library/src/main/java/org/phoenicis/library/ShortcutReader.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Interface for all shortcut reader implementations
+ */
+public interface ShortcutReader {
+    /**
+     * Sets the shortcut
+     *
+     * @param shortcut The shortcut
+     */
+    void of(ShortcutDTO shortcut);
```
madoar: I'm still thinking whether it is better to add the functionality of this method to the constructor. 

Changing the constructor to have the signature:
```java
public ShortcutReader(ShortcutDTO shortcut)
```

has the benefit that the object usage is cleaner. In addition it has the disadvantage that we can't add the constructor requirement to the interface.

madoar:
I'm also thinking about whether it makes sense to move the `WineShortcutReader` implementation to a new file. 

This leads me to a second question, why do we have both a `WineShortcutReader` and a `ShortcutReader` class with the same method signatures? In theory we could call the `WineShortcutReader` directly if we know that the shortcut is for a wine container.

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutReader.java
plata: If it is part of the constructor, it cannot be defined in the interface, right?

plata:
I can't really answer this because this functionality has been there before I joined. So this is a question for @qparis. However, I would agree that implementing the `ShortcutReader` for each engine should be sufficient.

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutReader.java
madoar: Exactly

madoar:
I have just created #2033, https://github.com/PhoenicisOrg/scripts/issues/1068 and https://github.com/PhoenicisOrg/scripts/issues/1069 to continue the open discussions. 

I think we can merge this as is, and work on the created issues separately.

madoar:
Any news on this? It would be nice if we could merge this before https://github.com/PhoenicisOrg/phoenicis/pull/2050

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutRunner.java
```diff
@@ -43,24 +43,30 @@ public void run(ShortcutDTO shortcutDTO, List<String> arguments, Consumer<Except
         final InteractiveScriptSession interactiveScriptSession = scriptInterpreter.createInteractiveSession();
 
         interactiveScriptSession.eval("include(\"engines.wine.shortcuts.reader\");",
-                ignored -> interactiveScriptSession.eval("new ShortcutReader()", output -> {
-                    final Value shortcutReader = (Value) output;
+                result -> {
```
plata: This code is the same everywhere. Can we extract it to a separate method?

madoar: My idea was to do this later on in a centralized method in `PhoenicisScriptEngine`:

```java
public interface PhoenicisScriptEngine {
    <E> E evalAndReturn(String script, Class<E> type);
    
    <E> void eval(String script, Class<E> type, Consumer<E> onSuccess, Consumer<Exception> onError);
}
```

See also our [documentation](https://phoenicisorg.github.io/phoenicis/Developers/script-interaction/).

For now I would suggest to retain the redundancy and fix this later for all script types (i.e. engines, installers, plugins, shortcuts, verbs etc.) at once.

plata: I'm fine with that.

plata:
You need to resolve some conflicts.

madoar:
@plata the conflicts are resolved. I would like to merge this soon because this PR blocks other useful changes to the script engine.

PR #2029, opened 2019-07-02
Fix wizard browse step

plata:
fixes https://github.com/PhoenicisOrg/scripts/issues/1054

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/wizard/UiSetupWizardImplementation.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/wizard/UiSetupWizardImplementation.java
@@ -283,8 +283,9 @@ public class UiSetupWizardImplementation implements SetupWizard {
      */
     @Override
     public String browse(String textToShow, String directory, List<String> allowedExtensions) {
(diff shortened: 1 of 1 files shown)


madoar:
When looking at our Java - JavaScript interop code I discover that we have almost no input/output validation (i.e. is `directory` a valid directory?, is the result for a browse operation valid?). 

I think we should add such checks to the code.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/wizard/UiSetupWizardImplementation.java
```diff
@@ -283,8 +283,9 @@ public String browse(String textToShow) {
      */
     @Override
     public String browse(String textToShow, String directory, List<String> allowedExtensions) {
+        final List<String> copiedAllowedExtensions = allowedExtensions != null ? List.copyOf(allowedExtensions) : null;
```
madoar: Are you sure it should be null and not an empty List?

plata: It's the same behavior like before.

plata:
Please open a new issue for the particular input/output validation you would like to have.

madoar:
See #2032

PR #2034, opened 2019-07-07
Remove Nashorn

madoar:
This PR removes Nashorn from Phoenicis. The main reason for this is that our Scripts repository is now incompatible with the JS standard supported by Nashorn.


madoar:
I think we can also remove some dependencies from the build scripts, but I don't know which ones

PR #2035, opened 2019-07-07
Add Support for automatic Object Conversion from Java to JS

madoar:
This PR adds support for some more automatic conversions from Java types to JS types (for example Java `List` to JS `Array`). See also https://github.com/graalvm/graaljs/issues/88 for more details.

I expect this to allow us to remove almost all `Java.from` operations.


phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PolyglotScriptEngine.java
```diff
@@ -43,7 +43,8 @@ public PolyglotScriptEngine(String language, Map<String, String> options) {
         this.language = language;
         this.context = Context.newBuilder(language)
                 .allowExperimentalOptions(true)
-                .options(options).allowHostAccess(true).build();
+                .allowHostAccess(true)
```
madoar: I tried to replace by `allowHostAccess(true)` by something else because it seems to be deprecated but I couldn't find a working replacement operation.

PR #2040, opened 2019-07-08
Fix JS includes (avoid null return if script already included)

plata:
With the old implementation, the `IncludeInjector` would return `null` if the same path has already been included before.

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
@@ -27,7 +27,7 @@ public class IncludeInjector implements EngineInjector {
             if (!includedScripts.containsKey(argument)) {
                 final String script = scriptFetcher.getScript(argument);
                 if (script == null) {
-                    throwException(new ScriptException(argument + " is not found"));
+                    throwException(new ScriptException("Script '" + argument + "' is not found"));
                 }
 
                 includedScripts.put(argument,
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
@@ -24,12 +24,12 @@ public class IncludeInjector implements EngineInjector {
         final Map<String, Object> includedScripts = new HashMap<>();
 
         phoenicisScriptEngine.put("include", (Function<String, Object>) argument -> {
-            final String script = scriptFetcher.getScript(argument);
-            if (script == null) {
-                throwException(new ScriptException(argument + " is not found"));
-            }
-
             if (!includedScripts.containsKey(argument)) {
+                final String script = scriptFetcher.getScript(argument);
+                if (script == null) {
+                    throwException(new ScriptException(argument + " is not found"));
+                }
+
                 includedScripts.put(argument,
                         phoenicisScriptEngine.evalAndReturn("//# sourceURL=" + argument + "\n" + script,
                                 this::throwException));
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
@@ -4,8 +4,8 @@ import org.phoenicis.scripts.engine.implementation.PhoenicisScriptEngine;
 import org.phoenicis.scripts.interpreter.ScriptException;
 import org.phoenicis.scripts.interpreter.ScriptFetcher;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.function.Function;
 
 /**
@@ -20,7 +20,8 @@ public class IncludeInjector implements EngineInjector {
 
     @Override
     public void injectInto(PhoenicisScriptEngine phoenicisScriptEngine) {
-        final Set<String> includedScripts = new HashSet<>();
+        // store included scripts (include path -> JS object)
+        final Map<String, Object> includedScripts = new HashMap<>();
 
         phoenicisScriptEngine.put("include", (Function<String, Object>) argument -> {
             final String script = scriptFetcher.getScript(argument);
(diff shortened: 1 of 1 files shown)


plata:
Even with this fix, I still get issues in `EngineSettingsManager` but I think they are not related.

plata:
@madoar please give this a try.

I'm wondering if we could find a way to use ES modules with `import` instead of our own `include` system...

plata:
I'm not sure that `Context#eval` does what we expect.

For
```
class test { myMethod() {}}
```
it gives "undefined".
For
```
class test { myMethod() {}}
test
```
it gives "class test { myMethod() {}}".

plata:
An additional problem is that `as(EngineSetting.class)` does not care if the methods in the interface are implemented. Actually, what happens right now is: The last include is `regedit` so this is returned by `eval`. Then, we try to create a new instance of the `regedit` function and return it as `EngineSetting`...

madoar:
> I'm not sure that Context#eval does what we expect.
> For
> `class test { myMethod() {}}`
> it gives "undefined".
> For
> `class test { myMethod() {}}`
> test
> it gives "class test { myMethod() {}}".

This behavior of graaljs makes no sense to me. Are you sure about this? If this is really the case I think we should open an issue on the graaljs repository and ask about the behavior of `eval`.

Another thing we could try are anonymous classes:

```javascript
class { 
   myMethod() {}
}
```

madoar:
> I'm wondering if we could find a way to use ES modules with import instead of our own include system...

I thought about this previously too. The issue with ES modules is that we can't perform an "anonymous" import. We need know the names of the imported classes/functions/values to be able to use them, which we don't do when we load a script from Java.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
```diff
@@ -20,20 +20,22 @@ public IncludeInjector(ScriptFetcher scriptFetcher) {
 
     @Override
     public void injectInto(PhoenicisScriptEngine phoenicisScriptEngine) {
-        final Set<String> includedScripts = new HashSet<>();
+        // store included scripts (include path -> JS object)
+        final Map<String, Object> includedScripts = new HashMap<>();
```
madoar: Maybe also add a check for circular dependencies (i.e. `A` -> `B` -> `A`)?

This should be relatively easy to implement with a `Stack`

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
```diff
@@ -20,20 +20,22 @@ public IncludeInjector(ScriptFetcher scriptFetcher) {
...
 
         phoenicisScriptEngine.put("include", (Function<String, Object>) argument -> {
-            final String script = scriptFetcher.getScript(argument);
-            if (script == null) {
-                throwException(new ScriptException(argument + " is not found"));
+            if (!includedScripts.containsKey(argument)) {
+                final String script = scriptFetcher.getScript(argument);
+                if (script == null) {
+                    throwException(new ScriptException(argument + " is not found"));
```
madoar: `"Script " + argument + " is not found"`

plata:
> Are you sure about this?

At least that's what I see in the debugger. I would really like if you could have a look at it as well because for me this behavior is really strange.
I've tried the anonymous class but there I only get `""` as return.

plata:
Just tested again. It seems that the behavior changed from 19.0.0 to 19.1.0. I have opened an issue for graaljs to clarify (https://github.com/graalvm/graaljs/issues/191).

Zemogiter:
Can we merge this now?

madoar:
According to the graaljs developers this doesn't seem to be a bug but a feature.

For now I recommend to revert our used graaljs version to `19.0.2`. Long-term we will need to rethink how we pass classes between scripts or from a script to Phoenicis. The best solution in my opinion is to either modify our `include` construct or change to ES 6 modules (i.e. their `import`/`export` syntax).

What do you think about this @qparis?

qparis:
I agree yes

plata:
Wouldn't it be better to use one of their proposed solutions
(class C {}) or class C {}; C? Especially the first one does not clutter the syntax too much and we could still use the most recent graaljs with all bug fixes.

madoar:
I'm fine with that as well, but it seems like a larger number of changes for a quickfix which we basically only need as an interim solution. 

I'm also not aware of any problems we have with the "old" graaljs version.

plata:
I'm not so sure that it's an interim solution only. Right now, we do not even know if it is actually possible to use ES6 modules and it might very well be that it's not.

madoar:
Even if we are unable use ES6 modules, we can still extend our current include mechanism with an `export` mechanism, see also [here](https://www.w3schools.com/nodejs/nodejs_modules.asp).

plata:
From my understanding, `exports` will only work in combination with `require`. Or would you want to implement the exports mechanism on our own? If so, I don't think that's an easy/fast task and honestly I don't quite see the advantages.

madoar:
Yes I would want to implement this on our own. I think it is comparable easy. First we would need to "inject" an empty "module" object in the executed script, which the script then populates with the exported results. This could be done like the following (pseudocode):

```javascript
const tmpScript = function(module) {
   // the included script content
}

val tmpModule = {};
// populate the tmpModule through the script
tmpScript(tmpModule);

// return tmpModule to the script calling `include(...)`
tmpModule
```

The calling script can access the result then like this:

```javascript
const fileSystem = include("id.to.filesystem");
```

Additionally we could add a default export like this:

```javascript
const tmpScript = function(module) {
   // the included script content
}

val tmpModule = {};
// populate the tmpModule through the script
tmpScript(tmpModule);

if (tmpModule.default) {
   // return only the default export
   tmpModule = tmpModule.default;   
}

// return tmpModule to the script calling `include(...)`
tmpModule
```

The included scripts would then look like the following:

```javascript
...
// multiple exports:
module.value = "Exported String";
module.function = (x, y) => x + y;

module.default = class X {
   ...
};
```

plata:
Can we merge this and continue the discussion in a separate issue? The fix in this pr is required anyways.

madoar:
I think we can merge this as is, but I think it would be useful to do the two fixes I mentioned above during last review :)

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
plata: Can we do that later? It doesn't really belong to the issue which is fixed in this PR.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
plata: Done.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
madoar: As long as we don't forget it I'm fine with that. I think this is an important check because it makes it easier to find certain types of errors.

plata: See https://github.com/PhoenicisOrg/phoenicis/issues/2048.

madoar:
Feel free to merge :)

Issue #2015, opened 2019-06-22
Prevent access to Java classes from JS


plata:
All functionality used in Javascript shall be implemented as a `Bean` such that we have a well defined API.

madoar:
Examples:
- `ProcessBuilder`

plata:
@qparis do you know how this white listing could be done? For our API we have `@Safe` but what about the normal Java classes.

qparis:
Maybe Context#allowHostClassLookup

plata:
Could you do a proof of concept (for me it's not quite clear currently)?

qparis:
See PR #2018

PR #2044, opened 2019-07-12
Add a ProcessUtils Class

madoar:
This PR adds a new `ProcessUtils` class which can be used as a `Bean` from the scripts to start new processes (see also #2015)

### DIFF
--- a/phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcess.java
+++ b/phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcess.java
@@ -1,0 +1,53 @@
+package org.phoenicis.tools.processes;
+
+import org.apache.commons.io.IOUtils;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+
+/**
+ * A wrapper class around the Java {@link Process} class allowing for a save access to the underlying process from the scripts
+ */
+public class PhoenicisProcess {
+    /**
+     * The process underneath
+     */
+    private final Process process;
+
+    /**
+     * Constructor
+     *
+     * @param process The wrapped process
+     */
+    public PhoenicisProcess(Process process) {
+        super();
+
+        this.process = process;
+    }
+
+    /**
+     * Waits for the underlying process to terminate
+     *
+     * @return This object
+     * @throws InterruptedException thrown if the process is interrupted
+     */
+    public PhoenicisProcess waitFor() throws InterruptedException {
+        this.process.waitFor();
+
+        return this;
+    }
+
+    /**
+     * Gets the output of the process as a String
+     *
+     * @return The output of the process
+     * @throws IOException thrown if an error during the result fetching occurs
+     */
+    public String getOutput() throws IOException {
+        if (this.process.isAlive()) {
+            throw new IllegalStateException("The process hasn't been terminated yet (hint try \"waitFor()\")");
+        }
+
+        return IOUtils.toString(process.getInputStream(), Charset.defaultCharset());
+    }
+}
--- a/phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcess.java
+++ b/phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcess.java
@@ -6,7 +6,8 @@ import java.io.IOException;
 import java.nio.charset.Charset;
 
 /**
- * A wrapper class around the Java {@link Process} class allowing for a save access to the underlying process from the scripts
+ * A wrapper class around the Java {@link Process} class allowing for a save access to the underlying process from the
+ * scripts
  */
 public class PhoenicisProcess {
     /**
--- a/phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcessBuilder.java
+++ b/phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcessBuilder.java
@@ -29,13 +29,6 @@ public class PhoenicisProcessBuilder {
      */
     private String outputPath;
 
-    /**
-     * Constructor
-     */
-    public PhoenicisProcessBuilder() {
-        super();
-    }
-
     public PhoenicisProcessBuilder withCommand(final List<String> command) {
         this.command = command;
 
@@ -81,8 +74,8 @@ public class PhoenicisProcessBuilder {
             throw new IllegalArgumentException("Base path is either does not exist or is no directory");
         }
 
-        if (this.environment == null) {
-            throw new IllegalArgumentException("Environment can't be null");
+        if (environment == null) {
+            environment = Map.of();
         }
 
         final ProcessBuilder processBuilder = new ProcessBuilder()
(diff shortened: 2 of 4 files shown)


phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcess.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.commons.io.IOUtils;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+
+/**
+ * A wrapper class around the Java {@link Process} class allowing for a save access to the underlying process from the scripts
+ */
+public class PhoenicisProcess {
```
madoar: We could think about moving this whole class to JS. This would mean that `ProcessUtils#startProcess` would return a Java `Process` object, which is then wrapped inside a JSObject in a proxy script.

qparis: In general, is better to have our own class because it makes us sure that there are no unsafe things that are exposed. So I like your implementation

madoar: My personal dilemma is the assessment between safety (like you mentioned) and ease of use.

If we implement this on the Java side we have full control over how things are processed and whether something should be prevented/disallowed for certain input.

If we implement this on the JavaScript side, we lose a lot of control because scripts can change or reload other scripts without Phoenicis being aware of this. On the other hand it is more more clear what functionality is available from inside the Scripts and how it needs to be accessed syntactically.

qparis:
Well not necessarily because we could generate a documentation from our @Safe annotated classes. This would help a lot.
Also, one day, we could try to use typescript

phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcess.java
plata: In this case, I would vote for safety. Actually, this might not even be safe enough because it still allows you to spawn random processes (not that I would know any alternative to that in our special case).
One other point to consider: returning a Java object ties the scripts to Java whereas a Bean could be implemented in any language.

phoenicis-tools/src/main/java/org/phoenicis/tools/processes/ProcessUtils.java
```diff
@@ -0,0 +1,61 @@
...
+     * Starts a new system process with the given command
+     *
+     * @param command     The command to be executed
+     * @param basePath    The location from where the command is executed
+     * @param environment A map of environemnt variables
+     * @param outputPath  An optional path where the output should be saved
+     * @return The created process
+     * @throws IOException thrown if the process couldn't be started
+     */
+    public PhoenicisProcess startProcess(List<String> command, String basePath, Map<String, String> environment, String outputPath) throws IOException {
```
madoar: @plata @qparis should pass all parameters in a single call like I did here, or should be pass them using a builder API (i.e. using `withCommand`, `withBasePath` etc methods)?

qparis:
It depends on how many parameter you have. More than 5-6 is enough to justify a builder

plata:
Additionally, many of these parameters (like environment) are optional so I think a builder would be good.

phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcess.java
madoar: Done

madoar:
Ok, I have changed the implementation and added a `PhoenicisProcessBuilder` to set the parameters.

phoenicis-tools/src/main/java/org/phoenicis/tools/processes/ProcessUtils.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        final File directory = new File(basePath);
+
+        if (!directory.exists() || !directory.isDirectory()) {
+            throw new IllegalArgumentException("Base path is either does not exist or is no directory");
+        }
+
+        if (environment == null) {
+            throw new IllegalArgumentException("Environment can't be null");
```
madoar: @plata if I read your previous comment correctly you want this to be:

```java
if (environment == null) {
   environment = Map.of();
}
```

right?

phoenicis-tools/src/main/java/org/phoenicis/tools/processes/ProcessUtils.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        final ProcessBuilder processBuilder = new ProcessBuilder()
+                .command(command)
+                .directory(directory)
+                .inheritIO();
+
+        processBuilder.environment().putAll(environment);
+
+        if (outputPath != null) {
```
madoar: Should I add some more optional parameters which we may need later on?

phoenicis-tools/src/main/java/org/phoenicis/tools/processes/ProcessUtils.java
plata: Yes. I don't think it's invalid to not specify any environment variables.

phoenicis-tools/src/main/java/org/phoenicis/tools/processes/ProcessUtils.java
plata: Like?

madoar: I don't know that is why I'm asking :D

plata: I can't think of anything currently.

phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcess.java
qparis: +1

phoenicis-tools/src/main/java/org/phoenicis/tools/processes/PhoenicisProcessBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A process builder to be used from inside the scripts
+ */
+public class PhoenicisProcessBuilder {
```
plata: In the DTO's, the DTO and the builder are in the same file. Should it be the same here?

madoar: I think the code is better readable if we split both classes in two separate files. But I'm fine either way.

plata: I don't care too much but it might make sense to be consistent.

madoar: After thinking some more about this I'm not sure this would work, because I don't know if it is possible to access inner Java classes from JavaScript.

madoar:
@plata I would suggest to add this class for now. We can still adjust it later on if we discover some hard edges.

Issue #2056, opened 2019-08-11
Enhance include System with more detailed Exceptions


madoar:
Currently when an included script throws an error the shown error message doesn't tell the user where the error occurred, i.e. in which script the error occurred. We should add this information to the shown error message to ensure script developers an easier time fixing their scripts.

plata:
Can you give more detailed information on how this could be done?

madoar:
In theory we only need to execute 

https://github.com/PhoenicisOrg/phoenicis/blob/727481fa7ee714125539b96c12002add7e50fc4f/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java#L36-L43

in a `try/catch` block. 

When catching an exception in this block we can then wrap it in a new `ScriptIncludeException` which we then again throw. This new `ScriptIncludeException` should contain a `scriptId` field, that contains information about the script in which the exception occurred.

PR #2063, opened 2019-08-22
Improve User Friendliness of the "include" Operation

### LINKED ISSUE #2048: Handle circular JS includes
i.e. A → B → A

to be implemented in `IncludeInjector` (e.g. with a `Stack`)

madoar:
This PR:
- adds three new exceptions to the script engine: CircularIncludeException, IncludeException and ScriptNotFoundException
- checks for circular includes (fixes #2048)
- throws a dedicated exception when an include fails (#2056)
- moves the exceptions to a new subpackage

Open issues:
The PR works, but our exception processing seems to be a bit strange. In some cases it seems like only parts of the exceptions are shown to the user. Sometimes the terminal contains a different GUI-based exception instead of the `ScriptException`. In addition it seems like the previous `throwException` wrapped every given exception into an additional `ScriptException` which makes understanding the shown exceptions more difficult.

### DIFF
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
@@ -2,11 +2,15 @@ package org.phoenicis.scripts.engine.injectors;
 
 import org.graalvm.polyglot.Value;
 import org.phoenicis.scripts.engine.implementation.PhoenicisScriptEngine;
-import org.phoenicis.scripts.interpreter.ScriptException;
+import org.phoenicis.scripts.exceptions.CircularIncludeException;
+import org.phoenicis.scripts.exceptions.IncludeException;
+import org.phoenicis.scripts.exceptions.ScriptException;
+import org.phoenicis.scripts.exceptions.ScriptNotFoundException;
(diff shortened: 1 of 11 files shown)


phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/injectors/IncludeInjector.java
```diff
@@ -21,34 +25,49 @@ public IncludeInjector(ScriptFetcher scriptFetcher) {
...
     public void injectInto(PhoenicisScriptEngine phoenicisScriptEngine) {
+        // store scripts that are currently in progress
+        final Stack<String> includeStack = new Stack<>();
         // store included scripts (include path -> JS object)
         final Map<String, Object> includedScripts = new HashMap<>();
 
         phoenicisScriptEngine.put("include", (Function<String, Object>) argument -> {
+            // prevent circular includes
+            if (includeStack.contains(argument)) {
+                throw new CircularIncludeException(argument, includeStack);
```
plata: Won't this occur also if a->b->c and a->c?

madoar: No, because `a -> b -> c` is first evaluated in complete. Only after this returns the second include (i.e. `a -> c`) is evaluated.

PR #2064, opened 2019-08-22
Parallelize "unzip"

### LINKED ISSUE #2058: Certain zip files cannot be extracted
### Expected behavior
`ZipScript` extracts .zip.

### Actual behavior
Extracting stops somewhere during the process or takes forever.

### Steps to reproduce
see https://github.com/PhoenicisOrg/scripts/pull/1081

### System information
* current master
* Manjaro 18
* Java 11

madoar:
This PR tries to fix #2058

### DIFF
--- a/phoenicis-tools/src/main/java/org/phoenicis/tools/archive/Zip.java
+++ b/phoenicis-tools/src/main/java/org/phoenicis/tools/archive/Zip.java
@@ -19,85 +19,76 @@
 package org.phoenicis.tools.archive;
 
 import com.google.common.io.CountingInputStream;
-import org.apache.commons.compress.archivers.ArchiveInputStream;
-import org.apache.commons.compress.archivers.ArchiveStreamFactory;
 import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
-import org.apache.commons.compress.utils.IOUtils;
+import org.apache.commons.compress.archivers.zip.ZipFile;
 import org.apache.commons.io.FileUtils;
 import org.phoenicis.configuration.security.Safe;
 import org.phoenicis.entities.ProgressEntity;
-import org.phoenicis.tools.stream.CursorFinderInputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.*;
+import java.io.File;
+import java.io.IOException;
 import java.nio.file.Files;
-import java.util.LinkedList;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 @Safe
 public class Zip {
     private static final String ZIP_ERROR_MESSAGE = "Unable to open input stream";
-    private static final byte[] ZIP_MAGICK_BYTE = new byte[] { 0x50, 0x4B, 0x03, 0x04 };
     private final Logger LOGGER = LoggerFactory.getLogger(Zip.class);
 
-    List<File> uncompressZipFile(File inputFile, File outputDir, Consumer<ProgressEntity> stateCallback) {
-        try (CountingInputStream inputStream = new CountingInputStream(new FileInputStream(inputFile))) {
-            final long finalSize = FileUtils.sizeOf(inputFile);
-            List<File> files = uncompress(inputStream, inputStream, outputDir, finalSize, stateCallback);
+    public List<File> uncompressZipFile(File inputFile, File outputDir, Consumer<ProgressEntity> stateCallback) {
+        final long finalSize = FileUtils.sizeOf(inputFile);
+        final AtomicLong counter = new AtomicLong(0);
 
-            return files;
+        LOGGER.info(String.format("Attempting to unzip file \"%s\" to directory \"%s\".",
+                inputFile.getAbsolutePath(), outputDir.getAbsolutePath()));
+
+        Path targetDirPath = outputDir.toPath();
+        try (ZipFile zipFile = new ZipFile(inputFile)) {
+            return Collections.list(zipFile.getEntries()).parallelStream()
+                    .map(e -> unzipEntry(zipFile, e, targetDirPath, counter, finalSize, stateCallback))
+                    .collect(Collectors.toList());
         } catch (IOException e) {
             throw new ArchiveException(ZIP_ERROR_MESSAGE, e);
         }
     }
 
-    /**
-     * Uncompress a tar
-     *
-     * @param countingInputStream to count the number of byte extracted
-     * @param outputDir The directory where files should be extracted
-     * @return A list of extracted files
-     * @throws ArchiveException if the process fails
-     */
-    private List<File> uncompress(final InputStream inputStream, CountingInputStream countingInputStream,
-            final File outputDir, long finalSize, Consumer<ProgressEntity> stateCallback) {
-        final List<File> uncompressedFiles = new LinkedList<>();
-        try (InputStream cursorInputStream = new CursorFinderInputStream(inputStream, ZIP_MAGICK_BYTE);
-                ArchiveInputStream debInputStream = new ArchiveStreamFactory().createArchiveInputStream("zip",
-                        cursorInputStream)) {
-            ZipArchiveEntry entry;
-            while ((entry = (ZipArchiveEntry) debInputStream.getNextEntry()) != null) {
-                final File outputFile = new File(outputDir, entry.getName());
-                if (entry.isDirectory()) {
-                    LOGGER.info(
-                            String.format("Attempting to write output directory %s.", outputFile.getAbsolutePath()));
+    private File unzipEntry(ZipFile zipFile, ZipArchiveEntry entry, Path targetDir, AtomicLong counter,
+            double finalSize, Consumer<ProgressEntity> stateCallback) {
+        try {
+            Path targetPath = targetDir.resolve(Paths.get(entry.getName()));
 
-                    if (!outputFile.exists()) {
-                        LOGGER.info(String.format("Attempting to createPrefix output directory %s.",
-                                outputFile.getAbsolutePath()));
-                        Files.createDirectories(outputFile.toPath());
-                    }
-                } else {
-                    LOGGER.info(String.format("Creating output file %s.", outputFile.getAbsolutePath()));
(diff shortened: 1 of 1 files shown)


phoenicis-tools/src/main/java/org/phoenicis/tools/archive/Zip.java
```diff
@@ -19,86 +19,77 @@
...
 
-                stateCallback
-                        .accept(new ProgressEntity.Builder()
-                                .withPercent(
-                                        (double) countingInputStream.getCount() / (double) finalSize * (double) 100)
-                                .withProgressText("Extracting " + outputFile.getName()).build());
+                    double count = counter.addAndGet(in.getCount());
 
+                    stateCallback
+                            .accept(new ProgressEntity.Builder()
```
madoar: This call (i.e. `stateCallback.accept(...)`) leads to an exception caused by graaljs, because multiple threads try to access the same graal context at the same time.

The reason for this is that `stateCallback` is a JS object (managed by graaljs) that is cast to the corresponding `Consumer<ProgressEntity>` java interface. If you comment line 83 - 86 out you get no exception.

I think we will need to provide a wrapper around `stateCallback` that ensures the calls to the JS object are only done from a single thread. A possible solution would be `Platform.runLater(() -> ...)` or the addition of a single thread `Executor` that manages the access to `stateCallback`.

plata: Does this mean that the progress bar is not working now? How do you handle race conditions if multiple files try to create the same directory?

madoar: > Does this mean that the progress bar is not working now?

Yes. To make it work we will need to decide how we can ensure that only a single thread performs the progress bar updates.

> How do you handle race conditions if multiple files try to create the same directory?

I think race conditions are handled by `Files.createDirectories`

plata: Can you try a separate single thread to update the progress bar?

madoar: Done, can you test it?

plata: Just tested with Hourglass: The extract is really fast but the progress bar is weird. It seems to go from 1 to 100% directly (actually this could even be true). However, the wizard panel was changing its size/content very rapidly. It's quite hard to describe, I think you must try yourself to see what I'm talking about.

madoar: Yes I also saw that yesterday but I wasn't sure whether this is an issue with my setup which doesn't seem to be the case...

When looking at the code we assume that the result of the extraction has the same size as the `zip` archive. I think this is the reason for the strange behavior of the progress bar. While the zip archive is around 1GB large the unzipped folder seems to be around 2GB large.

I think we will need to find a way to get the uncompressed size of the archive beforehand. Maybe [`ZipArchiveEntry#getSize`](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/zip/ZipArchiveEntry.html#getSize--) may be help here.

madoar: I've done some tests. It seems like `ZipArchiveEntry` only returns the correct uncompressed size after the whole entry has been decompressed (i.e. copied to the target folder). 

As an alternative we could use the number of finished entries as the progress indicator shown in the progress bar, i.e. `finished entries / total entries`. @plata what do you think?

plata: It will not be very accurate regarding the time estimation but I don't see a better alternative either.

madoar: We have another issue with the progress bar. It contains information about the last extracted file, but not about the file we are currently extracting. This was also the case before this PR.

plata: Is it an issue of how the progress bar is used here or of the progress bar implementation itself?

madoar: Ok I think I fixed the issue with the progress bar for now. @plata can you please try the PR again?

> Is it an issue of how the progress bar is used here or of the progress bar implementation itself?

I think both mentioned points are an issue.

> Is it an issue of how the progress bar is used here

Yes, if we could set the progress value and the description text separately (using two methods) it would help quite a lot. In this case we could set the text to `Extracting X` before we start extracting the entry and change the progress value afterwards.

> or of the progress bar implementation itself?

This is an issue on its own. If we had direct access to the progress bar (without the need to go through the graaljs consumer) we wouldn't need to use the `Future` trick, where we repeatedly query in the "main thread" whether the future object has been finished. Instead we could use `Platform.runLater(...)` to update the progress bar in a synchronous and thread-safe way.

madoar: When taking a look at the progress bar be aware that the zip archive of `Hourglass` contains a file that is 2.2GB huge.

plata: That's better.

So current state:
- speed is good
- progress bar based on file number instead of size not good with big files (like in this case)
- progress bar shows "extracted" instead of "extracting" for reasons explained above

@qparis what do you think about this?

madoar: > progress bar based on file number instead of size not good with big files (like in this case)

No the progress bar shows the progress based on the number of processed compressed bytes. (it seems like `Queue#drainTo` does strange stuff or doesn't clear the target list, the current solution doesn't depend on `Queue#drainTo`)

What I think is strange is the large speedup of the uncompress operation if one of the contained files makes up ~80\% of the contained data. This is strange because this file is still uncompressed inside a single thread. 

Maybe the speedup is also thanks to the use of `Files.copy(in, targetPath, StandardCopyOption.REPLACE_EXISTING);` instead of the "old" `IOUtils.copy(debInputStream, outputFileStream);` call. @plata can you try if the original implementation speeds up if you replace the `IOUtils` call with the `Files` call? If this is the case I think we will need to replace all `IOUtils` calls with corresponding `Files` calls. In addition wasn't there an old issue about a memory size limit for the `IOUtils` copy operations?

plata: > What I think is strange is the large speedup of the uncompress operation if one of the contained files makes up ~80% of the contained data. This is strange because this file is still uncompressed inside a single thread.

Then I misinterpreted this behavior (in assuming that it is a result of feeding the progress based on number of files).

plata: Just tried
```Java
Files.copy(debInputStream, outputFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
```
It's extremely slow.

madoar: Hmm. I'm just guessing at this point, but maybe it is one of the custom `InputStream`s that were used in the previous implementation.

plata: If that's really the case and we can make the old solution fast enough, we would not need to worry about the progress bar.

madoar: We should at least change the implementation to use `ZipFile`. The authors of `apache-commons` [recommend](https://commons.apache.org/proper/commons-compress/zip.html) this themselves:

> If possible, you should always prefer ZipFile over ZipArchiveInputStream.

plata: Can you try that?

madoar: Ok I've changed implementation back to a serial one. @plata can you give it a try?

plata: It's faster (and I would say fast enough) but the progress bar is still strange.

madoar: What do you mean by strange? The progress bar updates every time a contained file has been completely uncompressed. This means that you will see a small progress (up to `1/5` or `1/8`) then it will begin uncompressing the 2.2GB file... Afterwards the progress bar will be full (or almost full)

plata: I had the impression that e.g. the picture in the panel header was resizing/jumping around. Apart from that: Can't we show the real progress of the file copy now (like discussed e.g. [here](https://stackoverflow.com/questions/11182114/how-does-one-show-progress-of-a-java-file-copy-operation-in-icefaces))?

madoar: We could do that but this would make our file copy method much more complicated.
While we only need a single call at the moment, i.e.:

```java
Files.copy(in, targetPath, StandardCopyOption.REPLACE_EXISTING);
``` 

We would need multiple lines to do the same if we want to show real-time progress (taken from your link):

```java
long counter = 0;
int r = 0;
byte[] b = new byte[1024];
fin  = new FileInputStream(filein);
fout = new FileOutputStream(fileout);
while( (r = fin.read(b)) != -1) {
   counter += r;
   // update the progress bar
   fout.write(b, 0, r);
}
```

Additionally I expect this would cost us some performance because we would need to update the progress bar every x bytes (in the example every 1024 bytes)

plata: I think we should try it. Right now you can get the impression that something stopped if a big file is extracted. We have quite powerful machines but if you have a slow CPU and old HDD, it might be even worse.

madoar: @plata let's try changing the value inside the progress bar to something more meaningful in a different issue/PR. This PR is primary about speeding up the performance which we achieve by removing the used input streams.

plata: Ok, but then the title is wrong now. @qparis what do you think about this PR?

madoar: @plata I've changed the title to something more fitting

PR #2065, opened 2019-08-23
Sort the Verbs in the ContainerVerbsPanel according to their Names

madoar:
This PR sorts the verbs in the container verbs panel according to the verb names. This is something that has annoyed me for a long time now!

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainerVerbsPanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainerVerbsPanel.java
@@ -13,6 +13,8 @@ import org.phoenicis.javafx.utils.CollectionBindings;
 import org.phoenicis.repository.dto.ApplicationDTO;
 import org.phoenicis.repository.dto.ScriptDTO;
 
+import java.util.Comparator;
+
 /**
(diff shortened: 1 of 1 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/control/ContainerVerbsPanel.java
```diff
@@ -33,7 +35,7 @@
     private final ObjectProperty<ApplicationDTO> verbs;
 
     /**
-     * A list of all verbs
+     * A list of all verbs sorted according to their names
```
plata: Shouldn't we use `SortedList` instead of `ObservableList`?

madoar: We can. I just didn't want to use a more concrete type, should I?

plata: If we want that the list is sorted, we should also make sure it is in the class contract.

madoar: Done

Issue #2062, opened 2019-08-22
Add a new blocking "create" Method to UiQuestionFactory


madoar:
Currently there is only a non-blocking way to create a `question` dialog. After the user has provided an input (either yes or no), either the yesCallback or the noCallback is invoked.
This approach makes it hard to write sequential scripts. 

Therefore I propose we add a blocking `create` method to the `UiQuestionFactory` implementations. This new method should have the following signature:

```java
/**
 * Asks the user a question where he/she can answer either yes or no
 *
 * @param questionText The question text
 * @return True if the user answered yes and false if he answered no
 */
boolean create(String questionText);
```

See also https://github.com/PhoenicisOrg/scripts/pull/1040 for a use-case

plata:
Is this still relevant after #2067 has been closed?

PR #2067, opened 2019-08-23
Add a blocking UIQuestionFactory#create Method

madoar:
This PR adds a blocking `create` method to the `UIQuestionFactory` class as described in #2062

### DIFF
--- a/phoenicis-cli/src/main/java/org/phoenicis/cli/scriptui/UiQuestionFactoryCLI.java
+++ b/phoenicis-cli/src/main/java/org/phoenicis/cli/scriptui/UiQuestionFactoryCLI.java
@@ -32,6 +32,17 @@ public class UiQuestionFactoryCLI implements UiQuestionFactory {
 
     @Override
     public void create(String questionText, Runnable yesCallback, Runnable noCallback) {
+        final boolean answer = create(questionText);
+
+        if (answer) {
+            yesCallback.run();
+        } else {
+            noCallback.run();
+        }
+    }
+
+    @Override
+    public boolean create(String questionText) {
         String answer = "";
 
         while (!"yes".equals(answer) && !"no".equals(answer)) {
@@ -40,15 +51,8 @@ public class UiQuestionFactoryCLI implements UiQuestionFactory {
 
             Scanner input = new Scanner(System.in);
             answer = input.nextLine();
-
-            switch (answer) {
-                case "yes":
-                    yesCallback.run();
-                    break;
-                case "no":
-                    noCallback.run();
-                    break;
-            }
         }
+
+        return "yes".equals(answer);
     }
 }
--- a/phoenicis-cli/src/main/java/org/phoenicis/cli/scriptui/UiQuestionFactoryCLI.java
+++ b/phoenicis-cli/src/main/java/org/phoenicis/cli/scriptui/UiQuestionFactoryCLI.java
@@ -21,7 +21,11 @@ package org.phoenicis.cli.scriptui;
 import org.phoenicis.configuration.security.Safe;
 import org.phoenicis.scripts.ui.UiQuestionFactory;
 
+import java.util.List;
+import java.util.Map;
 import java.util.Scanner;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 @Safe
 public class UiQuestionFactoryCLI implements UiQuestionFactory {
@@ -54,5 +58,30 @@ public class UiQuestionFactoryCLI implements UiQuestionFactory {
         }
 
         return "yes".equals(answer);
+    }
+
+    @Override
+    public String create(String questionText, List<String> choices) {
+        Map<String, String> choiceMap = IntStream
+                .range(0, choices.size())
+                .boxed()
+                .collect(Collectors.toMap(
+                        index -> Integer.toString(index + 1),
+                        choices::get));
+
+        String answer = "";
+        while (!"0".equals(answer) && !choiceMap.containsKey(answer)) {
+            System.out.println(questionText);
+            System.out.print("Please enter:");
+            System.out.println("0\tCancel");
+            for (Map.Entry<String, String> entry : choiceMap.entrySet()) {
+                System.out.println(String.format("%s\t%s", entry.getKey(), entry.getValue()));
+            }
+
+            Scanner input = new Scanner(System.in);
+            answer = input.nextLine();
+        }
+
+        return "0".equals(answer) ? null : choiceMap.get(answer);
     }
 }
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/UiQuestionFactory.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/UiQuestionFactory.java
@@ -20,6 +20,8 @@ package org.phoenicis.scripts.ui;
 
 import com.google.common.util.concurrent.Runnables;
 
+import java.util.List;
+
 public interface UiQuestionFactory {
     /**
      * Creates a question UI (yes/no decision)
@@ -37,6 +39,14 @@ public interface UiQuestionFactory {
      * @return True if the user selects "yes", false if the user selects "no"
      */
     boolean create(String questionText);
+
+    /**
+     * Creates a question UI (multiple-choice decision)
+     * @param questionText The question text
+     * @param choices The choices
+     * @return The selected choice or null if the user canceled the question
+     */
+    String create(String questionText, List<String> choices);
 
     default void create(String questionText, Runnable yesCallback) {
         create(questionText, yesCallback, Runnables.doNothing());
--- a/phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/UiQuestionFactory.java
+++ b/phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/UiQuestionFactory.java
@@ -18,17 +18,27 @@
 
 package org.phoenicis.scripts.ui;
 
+import com.google.common.util.concurrent.Runnables;
+
 public interface UiQuestionFactory {
     /**
-     * creates a question UI (yes/no decision)
-     * @param questionText question
-     * @param yesCallback called when user selects "yes"
-     * @param noCallback called when user selects "no"
+     * Creates a question UI (yes/no decision)
+     *
+     * @param questionText The question text
+     * @param yesCallback The callback for when the user selects "yes"
+     * @param noCallback The callback for when the user selects "no"
      */
     void create(String questionText, Runnable yesCallback, Runnable noCallback);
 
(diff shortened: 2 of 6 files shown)


plata:
Shouldn't this be: "Add a blocking..."?

madoar:
Yes, my bad

phoenicis-cli/src/main/java/org/phoenicis/cli/scriptui/UiQuestionFactoryCLI.java
```diff
@@ -40,15 +55,33 @@ public void create(String questionText, Runnable yesCallback, Runnable noCallbac
...
 
-            switch (answer) {
-                case "yes":
-                    yesCallback.run();
-                    break;
-                case "no":
-                    noCallback.run();
-                    break;
+    @Override
+    public String create(String questionText, List<String> choices) {
```
plata: Don't we have another widget for this functionality already?

madoar: We have something similar in the installation wizard, i.e.:

https://github.com/PhoenicisOrg/phoenicis/blob/master/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/scriptui/StepRepresentationMenu.java

Which is used in

https://github.com/PhoenicisOrg/scripts/blob/c1bf3755e2c8cb6831695fb8ef98038e13fcd1a7/Applications/Games/League%20of%20Legends/Online/script.js#L34

The issue with this functionality is that it is contained inside the wizard, which can be quite easily overlooked when unexpected messages like a yes/no question for some unexpected situation is shown. A dialog that appears as a popup does not have this problem.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/UiQuestionFactory.java
```diff
@@ -18,17 +18,37 @@
...
+     */
+    boolean create(String questionText);
+
+    /**
+     * Creates a question UI (multiple-choice decision)
+     * @param questionText The question text
+     * @param choices The choices
+     * @return The selected choice or null if the user canceled the question
+     */
+    String create(String questionText, List<String> choices);
```
madoar: Should we return the selected choice or its index?

plata: How is the menu handled?

phoenicis-cli/src/main/java/org/phoenicis/cli/scriptui/UiQuestionFactoryCLI.java
plata: Doesn't a dialog break the concept of the wizard? Also: If you have multiple open wizards, it will be very hard to figure out where the dialog belongs to.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/UiQuestionFactory.java
madoar: As far as I know it returns the input string. The problem with this approach are translations. If the user can select a localized string it is better to work with indices.

phoenicis-cli/src/main/java/org/phoenicis/cli/scriptui/UiQuestionFactoryCLI.java
madoar: Both are true, but why do we have `UiQuestionFactory` then?

phoenicis-scripts/src/main/java/org/phoenicis/scripts/ui/UiQuestionFactory.java
plata: I remember some scripts which use two arrays (one with and one without translation) to circumvent this issue. Really not that nice.

phoenicis-cli/src/main/java/org/phoenicis/cli/scriptui/UiQuestionFactoryCLI.java
madoar: I think we have two choices here, either:
- we remove the `UIQuestionFactory` and move all "question" dialogs inside the installation wizards. I think this would work, but could be a bit problematic in case a user overlooks a question. (is there perhaps a way to make alert the user of a change that requires user input in the wizard?)
- we support some categories of dialogs, e.g. question dialogs for now.

@plata @ImperatorS79 @Zemogiter @qparis what do you think?

Zemogiter: I'm against removing UIQuestionFactory because of the problems you mentioned. Also because we would have to change every script that uses it.

plata: If we keep it, we need a clear guideline when to use it and when not.

madoar: I'm fine anyway but I think we should come to a decision relatively soon so that we can merge the blocked script PRs.

Zemogiter: @plata I don't think we really need a guideline for that because it's simply for asking questions to the user and letting said user choose from x options.

plata: @Zemogiter you could also use the menu step for that purpose.

However, don't feel blocked by this. We can still decide this in general later on and then change some scripts if required.

ImperatorS79:
@madoar This PR seems to block DXVK and D9VK PR that I would like to update afterwars with the git-fetched version list PR. Any news about this ?

madoar:
I still don't know how to proceed with this. We still have two choices on the table:
- we go with the already existing menu step
- we add a separate dialog factory

The dialog factory seems kind of redundant to the existing menu infrastructure. I would currently prefer the first solution where we go with the existing menu step. If some menu functionality is missing we can add this as well, I just think that a second redundant question infrastructure (based on a dialog) leads to more confusion

plata:
I agree that we should not duplicate functionality.

madoar:
The current tendency is to use existing functionality of the menu step to ask the user for input. Therefore I would propose to close this PR. If you believe that dialog is absolutely required we need some rules when to use a dialog and when to use the menu step.

Zemogiter:
The thing is I don't know how to use the existing menu functionality in a way that is fitting for my PRs. Therefore we need this PR.

plata:
@Zemogiter what exactly is the reason why you cannot use the existing menu?

Zemogiter:
@plata I don't know how to make an three-option menu screen with current menu functionality.

plata:
Did you check scripts which use the menu? It provides a list of n options.

Zemogiter:
There are only shortcut reader and D9VK/DXVK/VK9 that have `uiQuestionFactory`.

madoar:
@Zemogiter see https://github.com/PhoenicisOrg/scripts/blob/b9973a72768bb9c6714861cc8b00b880c43e9099/Applications/Custom/LocalInstaller/Local/script.js#L14-L34 for an example

Zemogiter:
`uiQuestionFactory` method looks cleaner

madoar:
Can you elaborate? To me the methods look pretty much the same:

```java
String create(String questionText, List<String> choices);
```

vs

```java
MenuItem menu(String textToShow, List<String> menuItems);

MenuItem menu(String textToShow, List<String> menuItems, String defaultValue);
```

Zemogiter:
With blocking `UIQuestionFactory` method everything is contained within single `if` instruction while that script have:

- a defined table of choices
- another defined table of choices that is visible to the end user (one table should be enough)
- a function that displays the choices
- variable that makes little sense to me at the end

madoar:
I don't think I fully understand your arguments. Why can't you have the things you're mentioning?

Let's take a look at the current code in https://github.com/PhoenicisOrg/scripts/pull/1039:

```javascript
            const answer = uiQuestionFactory.create(
                tr("D9VK is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Select how do you want to approach this situation."),
                ["YES, continue with D9VK installation regardless", "NO, quit script alltogether", "Exit D9VK Installer, but continue with the script"]
            );
            if (!answer || answer == "Exit D9VK Installer, but continue with the script") {
                return this;
            }
            if (answer == "NO, quit script alltogether") {
                throw new Error("User aborted the script.");
            }
```

You can write this in exactly the same manner with the menu API:

```javascript
const answer = wizard.menu(
   tr("D9VK is currently unsupported on non-Linux operating systems due to MoltenVK implementation being incomplete. Select how do you want to approach this situation.",
   ["YES, continue with D9VK installation regardless", "NO, quit script alltogether", "Exit D9VK Installer, but continue with the script"]
);

if (answer.index === 2) { // or if (answer.text === "Exit D9VK Installer, but continue with the script") {
   return this;
}

if (answer.index === 1) { // or if (answer.text === "NO, quit script alltogether") {
   throw new Error("User aborted the script.");
}
```

> variable that makes little sense to me at the end

The last variable is the default selection. Normally a dropdown menu either selects no choice or the first known choice. If you don't want this behaviour you can pass it a default selection.

The example I gave previously was only intended as a guideline I don't want to force you to use the menu API exactly like in the given example, if you want to implement the solution a bit more compact I'm also fine with that as long as it remains readable.

Zemogiter:
Now you got me confused. If we can do multi-choice questions with current implementation the way that is readable by me, then why this PR exists.

madoar:
> Now you got me confused. If we can do multi-choice questions with current implementation the way that is readable by me, then why this PR exists.

This is exactly the reason why @plata and I think we don't need this PR to provide the mentioned functionality :)

Do I understand you correctly that you are fine with using the `wizard.menu(...)` API instead of the `uiQuestionFactory`?

Zemogiter:
Yeah as long as it works and looks the same

madoar:
It won't look the same:
- `uiQuestionFactory` opens a dialog/popup to ask the user for the input
- `wizard.menu` in comparison asks the user for the input inside the installation wizard

Zemogiter:
I'm wondering which one will be better. Do we want a separate window with input options to open in case user minimizes the main window to do something else? On the other hand this may be too intrusive for some.

plata:
We shouldn't open new windows if possible. Just imagine you are running several installations in parallel. In this case you would be spammed with opening windows.

madoar:
I agree. I think we should try to use `wine.menu` for now. If we discover that we absolutely require a dialog later on we can still add it.

Issue #1824, opened 2019-02-09
Remember installed verbs for a container


madoar:
Currently Phoenicis doesn't remember which verbs were installed in a container. This contains both the verbs installed via a script and the ones installed via the container tab.

To help the user know the current state of his container(s) it is a good idea to save the installed verbs for each container in a file in the container folder. The content of this file could then be used to:
- skip the installation of already installed verbs when executing a script
- display the already installed verbs in the container tab differently (e.g. disable the checkbox, add an "Already installed" text)

plata:
Any preferred technical solution how to store this? Is there maybe a standard solution for this problem?

madoar:
I don't think that there is a standard solution. I think the easiest and best fitting way in our case is in a plain text file, maybe using the `json` format? @qparis do you have a suggestion?

qparis:
We need a bean to read and write into a global configuration file. The implementation is nor important, it could be a local json file as long as we take care about race conditions

plata:
Global? Wouldn't it be better per container?

qparis:
Yep, same principle. We already have a config file per container

plata:
Just to be sure ;)

madoar:
I would like to implement this soon. I think the issue can be divided in two parts:

- changes in Phoenicis:
   - add two new methods to `Engine`:
      ```java
      void registerVerb(String verbId);
      boolean isVerbRegistered(String verbId);
      ```
   - disable the installation checkbox in the container verbs panel when `isVerbRegistered` returns `true` for the given verb
- changes in the Scripts:
   - implement the two new methods `registerVerb` and `isVerbRegistered` for `Wine`
   - add a `registerVerb` call to every `Verb` implementation in the scripts.

Open issues:
Currently the verbs have no knowledge about their script id which is required to call `registerVerb`. I think it would be useful to inject this information long-term in each script. For now I think it is ok to hardcode the script id into each script.

@ImperatorS79 do you want to help with the necessary script changes?

ImperatorS79:
> disable the installation checkbox in the container verbs panel when isVerbRegistered returns true for the given verb

If a verb installation fails somehow, I would like to be possible to reinstall the verb. Maybe add a v next to the verb to say it is already installed ?

madoar:
I would suggest to call `registerVerb` as the last operation in the verb installation. This automatically means that the value is only set when the installation succeeds.

[...]

PR #2070, opened 2019-08-24
Support remembering previously completed Verb Installations

madoar:
This PR does the Phoenicis part discussed in #1824.

I'm not sure yet whether we want to add the new methods `registerVerb` and `isVerbRegistered` to the engine class, because they seem somehow unrelated to the engine to me.

### DIFF
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/Engine.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/Engine.java
@@ -127,7 +127,8 @@ public interface Engine {
      * Checks whether a given verb has been registered as being "installed"
      *
      * @param verbId The ID of the verb
+     * @param container The container where the verb has been installed
      * @return True if the verb has been registered as being "installed", false otherwise
      */
-    boolean isVerbRegistered(String verbId);
+    boolean isVerbRegistered(String verbId, String container);
 }
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/Engine.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/Engine.java
@@ -115,4 +115,19 @@ public interface Engine {
      * @param wizard setup wizard
      */
     void setWizard(SetupWizard wizard);
+
+    /**
+     * Registers as verb as being "installed"
+     *
+     * @param verbId The ID of the verb
+     */
+    void registerVerb(String verbId);
+
+    /**
+     * Checks whether a given verb has been registered as being "installed"
+     *
+     * @param verbId The ID of the verb
+     * @return True if the verb has been registered as being "installed", false otherwise
+     */
+    boolean isVerbRegistered(String verbId);
 }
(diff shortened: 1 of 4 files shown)


plata:
I agree. A `Container` class makes more sense.

madoar:
I thought about adding a new script to `Utils` maybe named `container_utils`. This script would export the above mentioned two methods. This has the benefit that the behavior is independent of the engine and it is located in a fixed location (under a fixed script id). 

What do you think?

plata:
Why should it be a script and not in Java?

madoar:
That is a good point. It doesn't need to be a script it could also be implemented in Java. For me it is more intuitive to implement it in Java because we will access the `registerVerb` method from inside the scripts and not from Java.

plata:
> For me it is more intuitive to implement it in Java because we will access the registerVerb method from inside the scripts and not from Java.

Did you really mean "implement it in Java" or "implement it in Javascript"?

madoar:
implement it in JavaScript

plata:
Ok, then the sentence makes more sense. Nevertheless, for me this should be in Java because it's nothing you would want to influence with custom scripts (it's similar to e.g. `ShortcutManager`).

madoar:
So basically you would add the two methods to a new with @Safe annotated Java class?

I.e.:

```java
@Safe
public class VerbUtils {
   void registerVerb(String verbId) {
      ...
   }

   boolean isVerbRegistered(String verbId) {
      ...
   }
}
```

plata:
Yes, something like that. We can decide if it belongs more to a verb or to a container.

madoar:
I have some problems getting the container information based on its name/id...

In a nutshell I want to add the following class to `phoenicis-tools`:

```java
package org.phoenicis.tools.container;

import org.phoenicis.configuration.security.Safe;

@Safe
public class ContainerUtils {
    void registerVerb(String containerId, String verbId) {
        ...
    }

    boolean isVerbRegistered(String containerId, String verbId) {
        ...
    }
}
```

The first thing I need to do for both methods is fetching the content of the `phoenicis.cfg` file from the container, but how do I do that? We also don't seem to have the `ContainerDTO` class available in `phoenicis-tools`.

plata:
Why don't you put it to `phoenicis-containers`?

madoar:
Do we have @Safe accessible from `pheonicis-containers`? Can we make the class accessible from the scripts if we put it in `phoenicis-containers`?

plata:
From my understanding, @qparis intention was to always put the JS API into the package where it belongs logically (and not all API together in one place). So yes, this should be possible.

madoar:
I think that it may be a good idea to wait with this PR until we have changed the repository structure to a more flexible one (I can't find the corresponding issue). Afterwards I think it will also be easier to fetch an entity (i.e. application, script, container etc.) for a given id/name. This should solve a lot of the problems I have in this PR.

plata:
Ok.

Issue #2061, opened 2019-08-21
Fetch graphics card info and make it available to script


ImperatorS79:
It should be possible, for example, to at least fetch the graphic card manufacturr (intel/nvdia/amd) in order, for example, to set VK_ICD_FILENAME correctly for Vulkan application.

Fetching OpenGL core context version and Vulkan version could also be interesting, simply to test if drivers are installed (in 32 bits and 64 bits).

plata:
This is somewhat related to #1854.

ImperatorS79:
#1854 is more general, but yes.
I see three ways of doing it:
- Use LWJGL to get the rendering infos (advantage: pure java)
- Use glxinfo and vulkaninfo (and call the program fro Java) (advantage: nearly nothing to do)
- Compile a simple C++ test program to check this (**only** advantage: can be compiled in 32 bits and 64 bits to check if the user have installed OpenGL/Vulkan 32 bits drivers and Vulkan 64 bits driver)

ImperatorS79:
I have normally a working LWJGL code, which compiles and execute in js, but I do not know how to check the returned value. Current code (most part are simply copy paste from LWJGL tutorial ^^):
```java
package org.phoenicis.tools.system;

import org.phoenicis.configuration.security.Safe;

import org.lwjgl.*;
import org.lwjgl.glfw.*;
import org.lwjgl.opengl.*;
import org.lwjgl.system.*;
import org.lwjgl.vulkan.*;

import java.nio.*;

import static org.lwjgl.glfw.Callbacks.*;
import static org.lwjgl.glfw.GLFW.*;
import static org.lwjgl.opengl.GL11.*;
import static org.lwjgl.system.MemoryStack.*;
import static org.lwjgl.system.MemoryUtil.*;
import static org.lwjgl.glfw.GLFWVulkan.*;

@Safe
class GraphicsProperties {
    public GraphicsProperties() {
        Vendor = "Unknown";
        OpenGLVersion = "Unsupported";
        VulkanVersion = "Unsupported";
    }

    public String Vendor;
    public String OpenGLVersion;
    public String VulkanVersion;
}

@Safe
// Code from LWJGL tutorial
public class GraphicsPropertiesFetcher {
    private long window;

    private void init() {
        // Setup an error callback. The default implementation
        // will print the error message in System.err.
        GLFWErrorCallback.createPrint(System.err).set();

        // Initialize GLFW. Most GLFW functions will not work before doing this.
[...]

PR #2071, opened 2019-08-25
Add a GraphicsPropertiesFetcher class

ImperatorS79:
Fetch: 
- Graphics card vendor and name
- OpenGL version
- Vulkan version

Limitation : cannot distinguish 32 bits and 64 bits support.


ImperatorS79:
How do I format the code correctly ?

ImperatorS79:
Hum, I have currently a bug when I launch two times an application using this. The second time it cannot load the OpenGL library :/.

ImperatorS79:
Fixed :wink:.

ImperatorS79:
What should I use for the exception ?

phoenicis-tools/pom.xml
```diff
@@ -126,5 +127,71 @@
...
+            <classifier>${lwjgl.natives}</classifier>
+        </dependency>
     </dependencies>
+    <profiles>
+        <profile>
+            <id>lwjgl-natives-linux-amd64</id>
+            <activation>
+                <os>
+                    <family>unix</family>
+                    <arch>amd64</arch>
```
madoar: Are we limiting us in any way if we specify `amd64` here?

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsProperties.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Safe
+public class GraphicsProperties {
+    public String vendor;
+    public String renderer;
+    public String openGLVersion;
+    public String vulkanVersion;
+    
+    public GraphicsProperties() {
+        vendor = "Unknown";
```
madoar: `this.vendor`, the same for the values below

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsProperties.java
```diff
@@ -0,0 +1,36 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package org.phoenicis.tools.system;
+
+import org.phoenicis.configuration.security.Safe;
+
+@Safe
+public class GraphicsProperties {
+    public String vendor;
```
madoar: @plata do we want the scripts to access objects of this type via getter methods or direct field access?

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsPropertiesFetcher.java
```diff
@@ -0,0 +1,116 @@
...
+
+import static org.lwjgl.glfw.Callbacks.*;
+import static org.lwjgl.glfw.GLFW.*;
+import static org.lwjgl.opengl.GL11.*;
+import static org.lwjgl.system.MemoryUtil.*;
+import static org.lwjgl.glfw.GLFWVulkan.*;
+
+// Code from LWJGL tutorial
+@Safe
+public class GraphicsPropertiesFetcher {
```
madoar: Please provide some more detailed comments over the class documentation

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsPropertiesFetcher.java
```diff
@@ -0,0 +1,116 @@
...
+
+        glfwTerminate();
+        glfwSetErrorCallback(null).free();
+        this.window = NULL;
+    }
+
+    /**
+     * Fetch graphics card vendor and OpenGL version
+     */
+    private void FetchVendorRendererOpenGLVersion(GraphicsProperties graphicsProperties) {
```
madoar: Method names are written in lower case in Java

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsPropertiesFetcher.java
```diff
@@ -0,0 +1,116 @@
...
+        graphicsProperties.renderer = glGetString(GL_RENDERER);
+        graphicsProperties.openGLVersion = glGetString(GL_VERSION);
+        graphicsProperties.openGLVersion = graphicsProperties.openGLVersion.substring(0,
+                graphicsProperties.openGLVersion.indexOf(' '));
+    }
+
+    /**
+     * Fetch Vulkan version
+     */
+    private void FetchVulkanVersion(GraphicsProperties graphicsProperties) {
```
madoar: Method names are written in lower case in Java

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsPropertiesFetcher.java
```diff
@@ -0,0 +1,116 @@
...
+                String.valueOf((version >> 12) & 0x3ff) + "." +
+                String.valueOf(version & 0xfff);
+    }
+
+    /**
+     * Fetch the current graphics properties (vendor, OpenGL version, Vulkan version) of the system
+     * 
+     * @return The current graphics properties inside a GraphicsProperties object
+     */
+    public GraphicsProperties GetProperties() {
```
madoar: Method names are written in lower case in Java

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsPropertiesFetcher.java
```diff
@@ -0,0 +1,116 @@
...
+        GraphicsProperties graphicsProperties = new GraphicsProperties();
```
madoar: `final`

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsProperties.java
```diff
@@ -0,0 +1,36 @@
...
+    public String renderer;
+    public String openGLVersion;
+    public String vulkanVersion;
+    
+    public GraphicsProperties() {
```
madoar: Please add a class comment

madoar:
What exception do you mean?

phoenicis-tools/pom.xml
ImperatorS79: This comes straight from LWJGL so I do not know.

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsProperties.java
ImperatorS79: How should it look like ?

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsPropertiesFetcher.java
ImperatorS79: Is not `final` similar to `const` in C++ ?

ImperatorS79:
Codacy is not happy due to `RuntimeException`. I also have an `IllegalStateException`. That code comes straight fro LWJGL start guide. I do not know what I should do here.

ImperatorS79:
I am also interested if other people could test it by adding:
```js
.preInstall(function() {
        const graphicsProperties = Bean("graphicsPropertiesFetcher").getProperties();
        print(graphicsProperties.vendor);
        print(graphicsProperties.renderer);
        print(graphicsProperties.openGLVersion);
        print(graphicsProperties.vulkanVersion);
    });
```
in the 7-zip script for example. My output looks like:
```
Intel Open Source Technology Center
Mesa DRI Intel(R) HD Graphics 630 (Kaby Lake GT2) 
3.0
1.1.0
```

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsPropertiesFetcher.java
madoar: Not exactly. In C++ `const` can be used to enforce two things:
- to enforce that the `const` value itself can't be replaced, i.e. the following is forbidden:

   ```c++
   const X *x = ...;

   x = new X(...); // forbidden
   ```
- the `const` value can't be changed, i.e. the following is forbidden:
   
   ```c++
   const X *x = ...;
 
   x->a = ...; // forbidden
   ```

In Java `final` can be used to enforce that the `final` value itself can't be replaced.

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsProperties.java
madoar:
```java
/**
 * What does GraphicsProperties contain? What is it used for? Where should it be used?
 */
class GraphicsProperties {
   ...
}
```

madoar:
The `IllegalStateException` is fine. The `RuntimeException` is bad because it is quite generic. Replace it with an `IllegalStateException` as well and you should be fine I think.

phoenicis-tools/pom.xml
ImperatorS79: On linux and macOS LWJGL is only available in 64 bits.

ImperatorS79:
Seems maven cannot find LWJGL on macos, but it should :thinking:.

phoenicis-tools/pom.xml
madoar: What would happen if you run the code with Linux 32 bit or an older mac?

ImperatorS79: I guess it would not work, obviously. De we really want to support such old architecture ? If so, the only possible solution is to write that as an external C++ code or to simply use glxinfo/vulkaninfo as stated in  #2061.

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsProperties.java
plata: Getters to be consistent with the rest of the API.

ImperatorS79: Why make things complicated ? Getters are really superfluous here.

ImperatorS79:
Now with this
```js
.preInstall(function() {
        const graphicsProperties = Bean("graphicsPropertiesFetcher").getProperties();
        print(graphicsProperties.vendor);
        print(graphicsProperties.renderer);
        print(graphicsProperties.openglVersion);
        print(graphicsProperties.openglCoreVersion);
        print(graphicsProperties.vulkanVersion);
    });
```
in the 7-zip script for example, you should get something like this:
```
Intel Open Source Technology Center
Mesa DRI Intel(R) HD Graphics 630 (Kaby Lake GT2) 
3.0
4.5
1.1.0
```

phoenicis-tools/pom.xml
madoar: I think we should try to support as many platforms as we comfortably can. In my opinion it is likely that quite a lot of people still use older 32 bit hardware, with an older Linux 32 bit OS, but i think we should ask for some more opinions, @plata @qparis what do you think about this?

phoenicis-tools/src/main/java/org/phoenicis/tools/system/GraphicsProperties.java
madoar: @ImperatorS79 it is our code style. Either all data classes (Bean classes) use getter and setter methods to access their content or none...

madoar:
I will test this the next days when I have more time

Zemogiter:
I'll test it in DXVK/D9VK/VK9 PRs once https://github.com/PhoenicisOrg/phoenicis/pull/2067 is merged.

ImperatorS79:
I would like to know, before continuing here, how we want to handle the graphics properties fetching.
It is really a useful tool IMO (for example to set VK_ICD_FILENAME).

madoar:
We need to have a solution that either:
- works for all supported operating systems/architectures
- works only for a subset of the supported OSes/architectures, but doesn't crash on other OSes/architectures

Correct me if I'm wrong but I think your proposed solution does not support 32 bit architectures and even leads to a crash on it.

ImperatorS79:
It does not work on pure/only 32 bits architecture simply because LWJGL is 64 bits only on linux.
If we do not use a pure java solution the other solution is to use (on linux) glxinfo and vulkaninfo (adding mesa-utils and vulkanutils as deps)

madoar:
I agree that writing a wrapper for these functions is something we should not do. Therefore my option two, i.e. simply do a check in the fetcher class where you check the architecture and OS. If the OS or architecture are unsupported by the used library return an object containing "unsupported" values

ImperatorS79:
> I agree that writing a wrapper for these functions is something we should not do
What do you mean ? ^^

> simply do a check in the fetcher class where you check the architecture and OS
I do not know how it works in java, but how can I compile without LWJGL on unsupported platform ?

madoar:
>> I agree that writing a wrapper for these functions is something we should not do
>
> What do you mean ? ^^

We have two choices if we want to provide these functions for 32 bit:
- we use some other applications which return the result in text format (I think `glxinfo` and `vulkaninfo` go in this direction)
- we write a small C/C++ library to fetch the necessary information, compile it for all necessary platforms and integrate it via JNI

Both solutions have their (large) issues. The first one requires that the used applications are available. The second one requires C and JNI. Both areas I would like to circumvent if we can.

>> simply do a check in the fetcher class where you check the architecture and OS
>
> I do not know how it works in java, but how can I compile without LWJGL on unsupported platform ?

I expect maven to work an an unsupported platform as well. It just won't add `LWJGL` to the build output. This means that if you actually try to access one of its methods during runtime you will stumble over an error. Therefore we need to execute a check **before** any LWJGL methods are called to ensure that the library is actually available.

ImperatorS79:
And how do I do that check ?

madoar:
I think https://github.com/PhoenicisOrg/phoenicis/blob/master/phoenicis-tools/src/main/java/org/phoenicis/tools/system/OperatingSystemFetcher.java and https://github.com/PhoenicisOrg/phoenicis/blob/master/phoenicis-tools/src/main/java/org/phoenicis/tools/system/ArchitectureFetcher.java will help

madoar:
Another important note: I think there is no more Java 32bit version for a while now. So this may actually be the reason why LWJGL does not support 32bit?

ImperatorS79:
@madoar I do not know why the packaging fails on macOS, the pom.xml is directly taken from lwjgl website. If you have any idea...

madoar:
I have no idea as well. Maven suggests that all files exist (see https://repo1.maven.org/maven2/org/lwjgl/lwjgl-vulkan/3.2.2/ for example). Maybe the cache used by travis is out of date. I don't know if we can force an update manually. Maybe updating lwjgl to version `3.2.3` helps?

ImperatorS79:
Now it fails under macOS with
```
[ERROR]     'dependencies.dependency.version' for org.lwjgl:lwjgl:jar:${lwjgl.natives} is missing. @ org.phoenicis:phoenicis-tools:[unknown-version], /Users/travis/build/PhoenicisOrg/phoenicis/phoenicis-tools/pom.xml, line 146, column 21

[ERROR]     'dependencies.dependency.version' for org.lwjgl:lwjgl-glfw:jar:${lwjgl.natives} is missing. @ org.phoenicis:phoenicis-tools:[unknown-version], /Users/travis/build/PhoenicisOrg/phoenicis/phoenicis-tools/pom.xml, line 151, column 21

[ERROR]     'dependencies.dependency.version' for org.lwjgl:lwjgl-opengl:jar:${lwjgl.natives} is missing. @ org.phoenicis:phoenicis-tools:[unknown-version], /Users/travis/build/PhoenicisOrg/phoenicis/phoenicis-tools/pom.xml, line 156, column 21
```
Again I just copy-paste the pom.xml fromlwjgl website

plata:
What's the status here?

ImperatorS79:
I do not know why the macOS build fails. The lines added to the pom come directly from lwjgl website, and it works on linux.

plata:
@qparis could you check the Mac OS build?

qparis:
Ok, I'll have a look once I have managed to build wine 32on64. This is a difficult task that is taking most of the time I'm devoting to Phoenicis at the moment

ImperatorS79:
Ok I think I found where the error came from. Arch for macos is x86_64 instead of amd64. Which lwjgl website did not mention.

plata:
What's the state of this pr?

ImperatorS79:
It compiles everywhere and works under linux. Needs to be tested on macos.

madoar:
> It compiles everywhere and works under linux. Needs to be tested on macos.

I think it may be worth it to just merge this as is. If later on someone can test it on mac and finds out that it does not work it is still not too late to add a separate solution for macos. @plata what do you think?

plata:
It's fine for me.

PR #2076, opened 2019-08-31
Add a new Key-Attribute List Control/Component

### LINKED ISSUE #1962: Change shortcut display and edit to deal with new environment variables setup
If you to edit a shortcut with the new shortcut code in https://github.com/PhoenicisOrg/scripts/pull/956, it will messed up the string stored in the shortcut file (deleting the " and replacing : by =) and then the shortcut will be unreadable. 

envVar

madoar:
This PR adds a new key-attribute list control/component and fixes #1962.

@ImperatorS79 please be sure to give this PR a thorough test.

Some screenshots:
grafik

grafik

I think we will need to do some more fine tuning on the look and feel, but otherwise I think we are good to go with this. If you have any improvement suggestions feel free to tell me


plata:
Is it possible to size the table to the required size (height)? Is there a scroll bar if it becomes too large?

madoar:
> Is it possible to size the table to the required size (height)?

I don't know. Actually I like it more if the table has a fixed height independent of the number of contained rows. This has the benefit that the buttons don't "jump" because you have added or removed a row.

> Is there a scroll bar if it becomes too large?

There should be

plata:
Ok. If it's tested by @ImperatorS79 and works like expected, I'm fine.

ImperatorS79:
I cannot add an environment variable because my changes are not remembered when I edit the value after the key and vice-versa.

madoar:
You currently need to apply the changes by clicking "enter" after doing them.
See also https://docs.oracle.com/javafx/2/ui_controls/table-view.htm (somewhere in the lower third)

ImperatorS79:
Ok it works for me.

plata:
Shall we merge this as it is and postpone the apply mechanism as future improvement?

madoar:
@ImperatorS79 the issue should be fixed now. Can you try again?

ImperatorS79:
Working for me.

PR #2077, opened 2019-09-01
Introduce the Module System to the Verbs

madoar:
This PR is required to install verbs directly from the GUI after the changes in https://github.com/PhoenicisOrg/scripts/pull/1112.

### DIFF
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
@@ -67,10 +67,8 @@ public class VerbsManager {
                 output -> {
                     final Value verbClass = (Value) output;
 
-                    final Verb verb = verbClass.newInstance().as(Verb.class);
-
                     try {
-                        verb.install(container);
+                        verbClass.invokeMember("install", container);
                     } catch (ScriptException se) {
                         errorCallback.accept(se);
                     }


phoenicis-engines/src/main/java/org/phoenicis/engines/VerbsManager.java
```diff
@@ -67,10 +67,8 @@ public void installVerb(String engineId, String container, String verbId, Runnab
                 output -> {
                     final Value verbClass = (Value) output;
 
-                    final Verb verb = verbClass.newInstance().as(Verb.class);
-
                     try {
-                        verb.install(container);
+                        verbClass.invokeMember("install", container);
```
madoar: It seems like this is the only way to access javascript `static` methods from Java.

plata: Why does it have to be `static` again?

madoar: There is no real reason why we use a `static` install method, there is just no better alternative in my opinion. Here some alternatives and why I think they are worse:

- export a second class/function containing the `install` method, next to the primary verb class. This would require us to return the two classes by name, which is bad during the `include`, because they would need to have very generic names like `Verb` (for Script usage) and `UIVerb` (for Phoenicis usage)
- provide a second script exporting a class/function with the `install` method. This approach would require twice the maintenance effort because we always need to watch out that **both** files making up a verb are updated in sync

plata: Good.

madoar: If we find an alternative in the future that is better than using a `static` method, I think it is still possible to change the verbs at that time. If you absolutely dislike the idea of using `static` we can also explore the alternatives some more. I mean they also have some benefits. For example the first alternative could also be used to **enforce** the name of the verb implementation used in the scripts. For example the verb adobeair could then export the following:

```javascript
// to be used from the Scripts
module.Adobeair = class Adobeair {
   ...
};

// to be used from Phoenicis
module.install = function () {
   ...
}

While the vcrun2008 verb could export:

```javascript
// to be used from the Scripts
module.Vcrun2008 = class Adobeair {
   ...
};

// to be used from Phoenicis
module.install = function () {
   ...
}

This in turn would then force the script developers to use `Adobeair` and `Vcrun2008` as the variable names during the include, while we can fetch the `install` method using `const { install } = include("verb.id");`

plata: > If you absolutely dislike the idea of using static we can also explore the alternatives some more

That's not the case. I'm perfectly fine using what you have now.

PR #2081, opened 2019-09-09
Add cabextract to gitpod

ImperatorS79:


plata:
@ImperatorS79 please always delete branches which are not required anymore.

PR #2093, opened 2019-10-02
Avoid exception if engine setting cannot be fetched

### LINKED ISSUE #1655: Container is unreadable if user.reg does not exist
### Steps to reproduce
* Create a container
* Remove user.reg 

An exception is thrown when trying to open the side panel

plata:
fixes #1655

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/skin/ContainerEngineSettingsPanelSkin.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/skin/ContainerEngineSettingsPanelSkin.java
@@ -14,6 +14,8 @@ import org.phoenicis.containers.dto.ContainerDTO;
 import org.phoenicis.engines.EngineSetting;
 import org.phoenicis.javafx.components.common.skin.SkinBase;
 import org.phoenicis.javafx.components.container.control.ContainerEngineSettingsPanel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static org.phoenicis.configuration.localisation.Localisation.tr;
 
@@ -22,6 +24,8 @@ import static org.phoenicis.configuration.localisation.Localisation.tr;
  */
 public class ContainerEngineSettingsPanelSkin
         extends SkinBase<ContainerEngineSettingsPanel, ContainerEngineSettingsPanelSkin> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ContainerEngineSettingsPanelSkin.class);
+
     /**
      * Constructor
      *
@@ -85,7 +89,15 @@ public class ContainerEngineSettingsPanelSkin
 
             // if the container is not specified set no default values
             if (container != null) {
-                engineSettingComboBox.setValue(engineSetting.getCurrentOption(container.getName()));
+                try {
+                    engineSettingComboBox.setValue(engineSetting.getCurrentOption(container.getName()));
+                } catch (Exception e) {
+                    engineSettingComboBox.getSelectionModel().select(0);
+                    LOGGER.warn(
+                            String.format("Could not fetch current option for engine setting \"%s\", will use default.",
+                                    engineSetting.getText()));
+                    LOGGER.debug("Caused by: ", e);
+                }
                 engineSettingComboBox.valueProperty().addListener((Observable invalidation) -> Platform.runLater(() -> {
                     getControl().setLockEngineSettings(true);
 


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/container/skin/ContainerEngineSettingsPanelSkin.java
```diff
@@ -85,7 +89,15 @@ private void updateEngineSettingsGrid(final GridPane engineSettingsGrid) {
 
             // if the container is not specified set no default values
             if (container != null) {
-                engineSettingComboBox.setValue(engineSetting.getCurrentOption(container.getName()));
+                try {
+                    engineSettingComboBox.setValue(engineSetting.getCurrentOption(container.getName()));
+                } catch (Exception e) {
```
madoar: Is it possible to catch something more specific here? I would expect that only scripting exceptions are thrown here right?

plata: I don't think so. For example, we can also have `PolyglotException`.

madoar: I think we should only call certain types of exceptions. Everything else links to a more drastic problem which we can't solve in this catch block

plata: I'm not sure. In my opinion, this is exactly the desired behavior: Try to get the currently used option. If that's not possible for whatever reason, fall back to default. I doesn't make sense to affect any other parts of Phoenicis by not catching the exception here.

madoar: In my opinion there are exceptions that should lead to a software crash. Or at least not to a fallback to a default option, for example security related exceptions.

plata: Like e.g.?

madoar: For example: https://docs.oracle.com/javase/7/docs/api/java/lang/SecurityException.html

plata: Even in that case: why should we allow this to crash Phoenicis? I don't see how anything bad can happen if we catch the exception.

madoar: You are right in that we can (and maybe even should) catch such exceptions as well. The issue is just that we shouldn't simply do a fallback to a default option if such an exception is thrown. At least a popup showing an error message to the user should be opened

plata: I'm not sure what we could show in the popup. What do we expect the user to do?

madoar: Check his repositories if they are trustworthy and/or updates are available. The official repositories should not throw such exceptions.

plata: It's not necessarily related to the repository. If you check the initial issue, this can also happen the container is broken.

madoar: Yes you're right!

madoar: Back to the main topic:
Where does this leave us? Do we want to catch two "groups" of exceptions:
- the first one allows for a fallback to the "default" option
- the second one requires at least an additional "ok" click from the user

Or do you prefer another solution?

plata: In my opinion, there's nothing useful the user could do at this point of time. Therefore, I would merge this PR as it is now.
What we could think about is: Do we need a detection if a container is broken? Is that check even possible?

madoar: I think we could do at least a really basic check where we check that all files required by the engine and phoenicis are available. Don't we already have another issue about this?

plata: I'm not sure. I will check and open a new issue if not.

PR #2094, opened 2019-10-02
JavaFX: set window decorations explicitly

### LINKED ISSUE #1745: Missing windows frame
5.0 Alpha 2
Elementary OS Juno

Missing the windows frame, can't even click on anything.

wine

plata:
fixes #1745

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
@@ -22,6 +22,7 @@ import javafx.application.Application;
 import javafx.scene.image.Image;
 import javafx.scene.text.Font;
 import javafx.stage.Stage;
+import javafx.stage.StageStyle;
 import org.phoenicis.javafx.controller.MainController;
 import org.phoenicis.multithreading.ControlledThreadPoolExecutorServiceCloser;
 import org.springframework.context.ConfigurableApplicationContext;
@@ -43,6 +44,7 @@ public class JavaFXApplication extends Application {
     @Override
     public void start(Stage primaryStage) {
         primaryStage.getIcons().add(new Image(getClass().getResourceAsStream("views/common/phoenicis.png")));
+        primaryStage.initStyle(StageStyle.DECORATED);
         primaryStage.setTitle("Phoenicis");
         loadFonts();
         ConfigurableApplicationContext applicationContext = new AnnotationConfigApplicationContext(


plata:
To be honest, I couldn't check if this is really fixing the linked issue because I'm not using this DE and I didn't get feedback. Nevertheless, this change should not hurt us so I think we can just use it and hope for the best.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/JavaFXApplication.java
```diff
@@ -43,6 +44,7 @@ public static void main(String[] args) {
     @Override
     public void start(Stage primaryStage) {
         primaryStage.getIcons().add(new Image(getClass().getResourceAsStream("views/common/phoenicis.png")));
+        primaryStage.initStyle(StageStyle.DECORATED);
```
madoar: This is the default according to the API: https://docs.oracle.com/javase/8/javafx/api/javafx/stage/Stage.html#initStyle-javafx.stage.StageStyle-

plata: I know but it's the only thing I could imagine to be wrong. Or do you have any idea what else could explain the behavior described in the issue?

madoar: I wouldn't eliminate a misconfiguration of the user somewhere. I think we should let the issue stay open and add a `help wanted` tag. Maybe someone else has a similar issue. This would allow us to better isolate the issue.

plata: In that case we are blocked for alpha 3.

madoar: I don't see a reason why this needs to be contained in Alpha 3? I think it is fine to move this to a later alpha version especially if we are unsure of the cause and can't reproduce the issue and therefore test a bugfix.

plata: Currently it's marked for alpha 3. I'm fine doing it later.

madoar: Then let us move it to a later release

lucsoft:
this needs more work  (the program can't be used on elemntary os 5.1)

lehnertz85:
I would like to add a temporary fix to the issue which may lead to a permanent fix. I discovered a similar issue when installing POL 4.2.3 which required the addition of an env variable in the .desktop. Here's a [link](https://github.com/elementary/gala/issues/244) to the issue reported to Elementary OS.

Replacing the `Exec=phoenicis %F` in Phoenicis.desktop with `Exec=env GTK_CSD=0 phoenicis %F` will allow the window to display correctly in Elementary OS 5.1

For those finding this from Google, .desktops are located in `/usr/share/applications/`.

image

plata:
@lehnertz85 thanks for the input. Is this something specifically for Elementary OS? In this case, it should probably be handled in the Elementary OS package. If it is something general, we can consider changing the start script.

lehnertz85:
I'm guessing since other distros aren't reporting this issue that it's only happening on a small subset and Elementary OS may be the only one with issues. If that is the case, then I agree that eOS should handle the fix. Looking through the issue, it appears Eclipse on Fedora also has/had issues. Eclipse fixed it through their app. It looks like a lot of finger pointing and no one seems to know why there are issues.

plata:
If the options do not have a bad effect on other systems, I'm fine changing it in our script.

plata:
Doesn't solve the issue.

lucsoft:
NOPE ist solves the problem

plata:
@giovannicaligaris @lehnertz85 can you confirm this?

If so, what's the correct solution? This or #2177?

PR #2095, opened 2019-10-02
Do not show hidden directories in containers tab

plata:
fixes PhoenicisOrg#1733

### DIFF
--- a/phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
+++ b/phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
@@ -170,33 +170,35 @@ public class GenericContainersManager implements ContainersManager {
 
         if (containerDirectories != null) {
             for (File containerDirectory : containerDirectories) {
-                final ConfigFile configFile = compatibleConfigFileFormatFactory
-                        .open(new File(containerDirectory, "phoenicis.cfg"));
-                final String containerPath = containerDirectory.getAbsolutePath();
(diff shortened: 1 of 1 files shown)


phoenicis-containers/src/main/java/org/phoenicis/containers/GenericContainersManager.java
```diff
@@ -170,33 +170,35 @@ public void deleteContainer(ContainerDTO container, Consumer<ContainerDTO> onSuc
...
-                        }).collect(Collectors.toList());
-
-                if (directory.getName().equals("wineprefix")) {
-                    containers.add(new WinePrefixContainerDTO.Builder().withName(containerDirectory.getName())
-                            .withPath(containerPath).withInstalledShortcuts(shortcutDTOS)
-                            .withArchitecture(configFile.readValue("wineArchitecture", ""))
-                            .withDistribution(configFile.readValue("wineDistribution", ""))
-                            .withVersion(configFile.readValue("wineVersion", ""))
-                            .build());
+                if (!containerDirectory.isHidden()) {
```
plata:
```Java
if (!containerDirectory.isHidden()) {
```
is the only change. The rest is indentation.

madoar: By indentation you mean spaces from/to tabs?

plata: No. The code is in the `if` now, so it's indented one level more.

PR #2096, opened 2019-10-02
Remove unused jgettext dependency

plata:
`slf4j` is required by `Localisation`


madoar:
For what was `jgettext` used previously?

plata:
I believe for the translations. We are using `gettext-commons` now.

PR #2097, opened 2019-10-05
Travis: also build with OpenJDK 13

### LINKED ISSUE #2089: Java 13
Now that Java 13 has been released we should test against it :)

plata:
fixes #2089


plata:
@madoar so apparently Travis has Java 13.

PR #2103, opened 2019-10-06
Remove remaining Nashorn stuff

### LINKED ISSUE #2038: Remove all Nashorn related Modules from the module path
After removing the Nashorn script interpreter in #2034 we also should remove the Nashorn related modules from the module path.

plata:
fixes #2038

I'm not sure about https://github.com/PhoenicisOrg/phoenicis/blob/730a6b85b0743bc2cfc393904bd12e711ec0cdd6/phoenicis-tests/src/main/java/org/phoenicis/tests/PhoenicisTestsApp.java#L70.

### DIFF
--- a/docs/_docs/Developers/intellij-idea.md
+++ b/docs/_docs/Developers/intellij-idea.md
@@ -19,7 +19,7 @@ add: JavaFXApplication with:
 * Main class: `org.phoenicis.javafx.JavaFXApplication`
 * VM options: 
 ```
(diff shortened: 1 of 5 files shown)


madoar:
That line needs to also be removed. The bean does not exist anymore

plata:
But what does `PhoenicisTestApp`do then? Obviously it's never executed because Travis is successful.

madoar:
> But what does PhoenicisTestApp do then? Obviously it's never executed because Travis is successful.

That is a good question. I think we should investigate this and if the class is actually not used remove it.

plata:
@qparis can you clarify?

plata:
@qparis can you tell something regarding `PhoenicisTestApp`?

qparis:
It is an old POC that aimed to show that we could run script automatically without any user interaction. It can be rewritten later on

plata:
So we keep it? Or should it be deleted?

qparis:
It can be deleted

PR #2104, opened 2019-10-06
Flatpak: do not use Wine runtime

plata:
e.g. Notepad++ crashes otherwise


ImperatorS79:
This is not a good idea if you cannot ensure that the version of the libs contained in the flatpak runtime are the same that the version of the libs that you have used to build wine.

I will take the example of libavcodec and libavutil, which prevent you to **install** league of legends if you do not have the exact same version.

So by hardcoding this, notepad++ will work but other apps will definitely not work. We should better look at why notepad++ does not work (which is strange indeed, we would like to test under buster to really see if it is the runtime the problem or a missing lib).

plata:
Feel free to try it. The only thing I can say is: When using the runtime, Wine does not work at all on Manjaro and at least certain apps do not work in Flatpak while it works without the runtime.
There's also another point with flatpak: If there's a missing dependency, it should be built in the flatpak environment and not used from the Wine runtime.

ImperatorS79:
> Wine does not work at all on Manjaro

This should be investigated further to know why. It is strange because steam also use a runtime for SteamPlay, and they do not have any problems. You could try to launch league of legends installation (new script) and see if it works (note that you could be the lucky guy with the version of libavcodec working which will not demonstrate that it works for all libs).

ImperatorS79:
So I believe we should add the option to the shortcut so that people can decide to use it or not.

madoar:
> If there's a missing dependency, it should be built in the flatpak environment and not used from the Wine runtime.

This is easier said than done. If we build it in the flatpak environment we go in the direction where we need to provide a separate runtime for each execution environment (currently: debian/ubuntu/arch/manjaro/macos/flatpak/etc). If we do this it will be really hard to maintain the whole project because there is always someone for whom it won't work. 

I tend to agree with @ImperatorS79 that using an "exception" mechanism seems to be the wrong choice.

plata:
> If we build it in the flatpak environment we go in the direction where we need to provide a separate runtime for each execution environment

I think it's rather the other way around. From the experiences with Manjaro and flatpak, I assume that it will not be possible to build a runtime which works on all distributions (also the Steam runtime must be enabled via environment variable explicitly). Therefore I believe that building the runtime in flatpak and using flatpak to **avoid** that "there is always someone for whom it won't work".

plata:
Also some questions/remarks regarding the Steam runtime:
1. Is there any reason why we do not use it directly?
2. There LD path setup is a bit more complex: https://github.com/ValveSoftware/steam-runtime/blob/master/templates/run.sh.
3. Most importantly: The Steam runtime works because Steam has been built inside that runtime. You cannot assume that an application which has not been built for that runtime will work. See especially the comment in `run.sh` "If ldd on a program shows any library in the system path, then that program may not run in the Steam runtime.".

ImperatorS79:
1) Because you need the exact same version of certain libs which you used to build wine in order for wine to works (lbavcodec, libavutil)

3) That is exactly why we provide the libs with which we builded wine : in order for wine to run properly

Our runtime is still however missing a lot of libraries/config files currently imo.

plata:
Your explanation sounds plausible. However, it does not explain the observed behavior (even if libraries are missing in the runtime).

madoar:
I think we need to find a common approach which we want to use for all platforms to ensure that all runtime dependencies are available. @plata @ImperatorS79 do you have an idea how we could achieve this for flatpak, i.e that flatpak uses the same runtime approach as we use for the normal installers (e.g. `deb`)? In addition how do we provide the runtime for OSX?

plata:
OSX should not be a problem as there are no distributions. So if the runtime is built for OSX, it should always work.
Regarding flatpak: It will work if we build Wine and the runtime in the flatpak environment.

ImperatorS79:
>  Regarding flatpak: It will work if we build Wine and the runtime in the flatpak environment.

Currently wine is build on debian stretch. I think that building it in the flatpak runtime will make it work perflectly under flatpak but not necessarly for non flatpak user (like if you build it under debian some of the things could not work for non debian user).  

However I tested a new runtime taking most of the libs from debian stretch, and still notepad++ does not work. I wonder then if it just simply a "bug" inside one of the debian lib.

I think we should rebuild wine under buster since it is out now. 

We should also look at how far we want to keep wine builds and rebuild them.

plata:
> like if you build it under debian some of the things could not work for non debian user

That's exactly the point and one of the main reasons why flapak, snap etc. exist.They give you an environment which is the same regardless of the distribution you are using. Building Wine and the runtime on Debian can never guarantee that it will work correctly/at all on any other distribution (maybe not even on a different Debian version).

ImperatorS79:
And you could do the same if you provide the environment from debian.

So if you have a bug in a debian lib or a flatpak lib -> some things will not work.

madoar:
> And you could do the same if you provide the environment from debian.

@ImperatorS79 what do you mean by `environment`? I don't think we can simply "provide" everything. The biggest issue is that we are kind of memory bound. In addition I don't think it is a good idea to download/install to many external dependencies later on (i.e. after Phoenicis has been installed).

> So if you have a bug in a debian lib or a flatpak lib -> some things will not work.

I think you have this issue independent of what we do, because we require on a source for the libs anyway and this source could contain bugs as well.

> Regarding flatpak: It will work if we build Wine and the runtime in the flatpak environment.

@plata how do you build Wine and the runtime in the flatpak environment? Does this need to be done once somewhere remove (i.e. by us) or is this done during the installation of the flatpak package? If it is done during the installation of the flatpak package is this possible with the number of libraries we would need to build? Would we require any additional system dependencies? How long would the installation process take in this case?

ImperatorS79:
> I don't think we can simply "provide" everything. The biggest issue is that we are kind of memory bound

You can note that kind of what flatpak and snap do, so that you end up with three times the same libs everywhere on your computer.

madoar:
That is why I asked what you meant by "everything". If you only mean libs I can live with that, but if we also require other dependencies in the future (e.g. a certain linux kernel version or whatever) then I don't think this is a feasible approach.

plata:
> And you could do the same if you provide the environment from debian.

If you do that (with everything your really need to have a complete environment - only some libs will simply not be enough), you will have exactly flatpak/snap once you are finished.

> @plata how do you build Wine and the runtime in the flatpak environment? Does this need to be done once somewhere remove (i.e. by us) or is this done during the installation of the flatpak package?

It can be done e.g. in winebuild. We cannot do it in the flatpak package because then we would be limited to the Wine versions available at build time (and we would have to ship all Wine versions).

> Would we require any additional system dependencies?

Maybe yes. Those can be part of the flatpak installation or installed later via runtime (though that doesn't make too much sense).

> How long would the installation process take in this case?

I'm assuming that this is in case we would install Wine as part of the flatpak? If so, not applicable for the reasons described above.

plata:
So what do we do about this PR? I don't see that we will have a flatpak runtime soon (at least not before the next release) and it simply doesn't work with runtime right now.

madoar:
Can you give me a short summarization how a user can de-/activate the runtime?

plata:
You can set the property on command line (like I did here) or in the config file.

PR #2107, opened 2019-10-12
Rewrite the Script Engine API

madoar:
This PR is still work in progress and at the current time only meant as a preview for discussion purposes!

### DIFF
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingScriptEngine.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingScriptEngine.java
@@ -1,0 +1,22 @@
+package org.phoenicis.engines;
+
+import org.graalvm.polyglot.Value;
+import org.phoenicis.scripts.engine.implementation.PhoenicisScriptEngine;
+import org.phoenicis.scripts.engine.implementation.TypedScriptEngine;
+
+public class EngineSettingScriptEngine implements TypedScriptEngine<EngineSetting> {
+    private final PhoenicisScriptEngine<Value> scriptEngine;
+
+    public EngineSettingScriptEngine(PhoenicisScriptEngine<Value> scriptEngine) {
+        super();
+
+        this.scriptEngine = scriptEngine;
+    }
+
+    @Override
+    public EngineSetting evaluate(String scriptId) {
+        final String script = String.format("include(\"%s\")", scriptId);
+
+        return scriptEngine.evaluate(script).newInstance().as(EngineSetting.class);
+    }
+}
--- a/phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
+++ b/phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
@@ -18,12 +18,12 @@
 
 package org.phoenicis.engines;
 
-import org.graalvm.polyglot.Value;
 import org.phoenicis.repository.dto.ApplicationDTO;
 import org.phoenicis.repository.dto.CategoryDTO;
 import org.phoenicis.repository.dto.RepositoryDTO;
-import org.phoenicis.scripts.engine.PhoenicisScriptEngineFactory;
-import org.phoenicis.scripts.engine.implementation.PhoenicisScriptEngine;
+import org.phoenicis.scripts.engine.implementation.TypedScriptEngine;
+import org.phoenicis.scripts.engine.implementation.TypedScriptEngineFactory;
+import org.phoenicis.scripts.exceptions.ScriptException;
 
 import java.util.List;
 import java.util.Map;
@@ -35,20 +35,20 @@ import java.util.stream.Collectors;
  * Manages the engine settings
  */
 public class EngineSettingsManager {
-    private final PhoenicisScriptEngineFactory phoenicisScriptEngineFactory;
+    private final TypedScriptEngineFactory typedScriptEngineFactory;
     private final ExecutorService executorService;
 
     /**
      * Constructor
      *
-     * @param phoenicisScriptEngineFactory The used script engine factory
-     * @param executorService The executor service to allow for parallelization
+     * @param typedScriptEngineFactory The used script engine factory
+     * @param executorService          The executor service to allow for parallelization
      */
-    public EngineSettingsManager(PhoenicisScriptEngineFactory phoenicisScriptEngineFactory,
-            ExecutorService executorService) {
+    public EngineSettingsManager(TypedScriptEngineFactory typedScriptEngineFactory,
+                                 ExecutorService executorService) {
         super();
 
-        this.phoenicisScriptEngineFactory = phoenicisScriptEngineFactory;
+        this.typedScriptEngineFactory = typedScriptEngineFactory;
         this.executorService = executorService;
     }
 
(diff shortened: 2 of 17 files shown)


phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PhoenicisScriptEngine.java
```diff
@@ -18,23 +18,25 @@
...
+    void put(String key, Object value);
+
+    /**
+     * Evaluates the given script and returns its result
+     *
+     * @param script The script
+     * @return The result of the evaluated script
+     */
+    R evaluate(String script);
 }
```
madoar: A `script engine` now provides a blocking method `evaluate` which returns the result of the evaluated script in a blocking manner. The result is of a generic type which depends on the concrete script interpreter used underneath (e.g. `Value` for Graal)

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/TypedScriptEngine.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+public interface TypedScriptEngine<T> {
+    /**
+     * Evaluates the script with the given id and returns its result cast to the given type
+     *
+     * @param scriptId The id of the evaluated script
+     * @return The cast result of the evaluated script
+     */
+    T evaluate(String scriptId);
+}
```
madoar: This class functions as an intermediate class for direct usage. It returns the result of an evaluated script (this time given by the script id) as an application type (e.g. `EngineSetting`, `Installer` etc.)

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/TypedScriptEngineFactory.java
```diff
@@ -0,0 +1,15 @@
...
+public interface TypedScriptEngineFactory {
+    /**
+     * Creates a new {@link TypedScriptEngine} instance for the given `type` input
+     *
+     * @param type The result type {@link Class} for the created script engine
+     * @param <T>  The result type
+     * @return The created script engine
+     */
+    <T> TypedScriptEngine<T> createScriptEngine(final Class<T> type);
+}
```
madoar: This class is a factory for `TypedScriptEngine` instances. At the end I plan to have a single implementation of the factory class which is able to return `TypedScriptEngine` for all application classes, but because the application classes are currently split into multiple modules we currently need to provide multiple `TypedScriptEngineFactory` implementations, one for each application class.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/PhoenicisScriptEngine.java
```diff
@@ -18,23 +18,25 @@
...
-
-    default void eval(String script, Consumer<Exception> errorCallback) {
-        eval(script, Runnables.doNothing(), errorCallback);
-    }
-
-    Object evalAndReturn(String line, Consumer<Exception> errorCallback);
-
-    void put(String name, Object object, Consumer<Exception> errorCallback);
-
-    void addErrorHandler(Consumer<Exception> errorHandler);
```
madoar: Because I'm removing the `error handlers` contained in the script engine we need to provide them nearer to where they are needed in specific try/catch blocks. I think this is a better approach because it is currently very hard to know what error handlers are called for a script because they are often specified in an entirely different code region.

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/GraalEngineType.java
```diff
@@ -0,0 +1,20 @@
...
+        return new PolyglotScriptEngine("js",
+                Map.of("js.nashorn-compat", "true",
+                        "js.experimental-foreign-object-prototype", "true"));
+    }
+
+    @Override
+    public String toString() {
+        return "graal.js";
+    }
+}
```
madoar: Here the `graal.js` specific implementation of the `ScriptEngineType`

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
```diff
@@ -35,49 +35,52 @@
...
      * Constructor
      *
-     * @param phoenicisScriptEngineFactory The used script engine factory
-     * @param executorService The executor service to allow for parallelization
+     * @param typedScriptEngineFactory The used script engine factory
+     * @param executorService          The executor service to allow for parallelization
      */
-    public EngineSettingsManager(PhoenicisScriptEngineFactory phoenicisScriptEngineFactory,
-            ExecutorService executorService) {
+    public EngineSettingsManager(TypedScriptEngineFactory typedScriptEngineFactory,
```
madoar: This class shows an example how I intend to use the `script engine` in the future.

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingScriptEngine.java
```diff
@@ -0,0 +1,22 @@
...
+        this.scriptEngine = scriptEngine;
+    }
+
+    @Override
+    public EngineSetting evaluate(String scriptId) {
+        final String script = String.format("include(\"%s\")", scriptId);
+
+        return scriptEngine.evaluate(script).newInstance().as(EngineSetting.class);
+    }
+}
```
madoar: Here the `TypedScriptEngine` for the `EngineSetting` application class

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/TypedScriptEngineFactory.java
plata: What do you mean by "application class"?

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
plata: Does it mean that we cannot access e.g. applications and engines via the same script engine?

plata:
Is this something you want to do before alpha-3?

phoenicis-scripts/src/main/java/org/phoenicis/scripts/engine/implementation/TypedScriptEngineFactory.java
madoar: By application class/type I mean interfaces like `Installer`, `Engine`, `EngineSetting`, `Verb` etc.

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
madoar: Yes. We will need to provide a dedicated `TypedScriptEngine` implementation for every application type (i.e. `Installer`, `Engine`, etc)

madoar:
Ideally yes, but I'm not sure about our schedule for alpha-3.

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
plata: Understood. So this is also because we want to keep the APIs in the modules and not centralize them in phoenicis-scripts (we discussed this somewhere else).

plata:
I have no clue regarding the schedule. There are some packaging issue which cannot fix.

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
madoar: > So this is also because we want to keep the APIs in the modules and not centralize them in phoenicis-scripts (we discussed this somewhere else).

No the reason why I want to implement a dedicated `TypedScriptEngine` class for every application type is because it divides the functionality better. It follows the approach divide and conquer in that we implement independent behavior in different modules/classes. For example what have the processing of `Installer` script and `Verb` script in common except that they are both executed with graaljs and use the module system? Therefore we divide the corresponding functionality in different classes.

I still plan to move all application types (and ideally also the `TypeScriptEngine` implementations) to the `phoenicis-scripts` module. I didn't do this yet because I think the main changes of this PR are easier comprehensible if I leave as much of the related refactoring as possible for later.

madoar:
I think we will need to wait until @qparis has more time before we can fix those.

I'm also currently a bit limited on time I can invest in this PR. Therefore feel free to give the PR an in-depth look so that we have a stable and durable solution when we finally merge the PR!

phoenicis-engines/src/main/java/org/phoenicis/engines/EngineSettingsManager.java
plata: > I still plan to move all application types

I thought @qparis doesn't like this?

PR #2149, opened 2019-12-29
Fix library category icon handling

### LINKED ISSUE #1626: Library: store category icons
Currently, the library relies on the category icons provided by the repository. As a consequence, the repository must be loaded to properly use the library (which takes some time).

Solution: During app installation, store category icon from repository in library (if it doesn't exist).

Implement a hierarchical shortcut directory like this:
- shortcuts
    - development
        - icon.png
        - notepadplusplus
            - Notepad++.info
            - Notepad++.shortcut

The library icon hierarchy would be:
- standard theme
- shortcut directory
- selected theme (if theme specifies icon for given category)

plata:
fixes #1626
requires https://github.com/PhoenicisOrg/scripts/pull/1159

We cannot use the default category buttons (of the apps) in the library tab for several reasons:
- It cannot be guaranteed that the repository which contained the category of a certain app does still exist after the app has been installed (e.g. the repository might have been deleted).
 - It should not be required to fetch the repository to use the library.

Therefore, use the following approach:
- When installing an app, store the icon for its category together with the shortcut (i.e. in the library).
- Use this icon as fallback if the theme does not specify an icon for this category.

#### Why I store the category icon for every shortcut
It is necessary to store the category icon per shortcut, otherwise category icons in the library could be overwritten if multiple shortcuts from different sources use the same category. This could lead to a strange behavior in the following case:

1. an app A with category C and category icon I(A) from repository R1 is installed
2. if the theme does not specify an icon, I(A) is shown in the library
3. an app B with category C and category icon I(B) from repository R2 is installed such that I(A) != I(B) (even though the category C is the same)
4. if the theme does not specify an icon, I(B) is shown in the library
5. app B is uninstalled
6. if the theme does not specify an icon, I(B) is still shown in the library even though the situation is the same as in 2.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibrarySidebarToggleGroupSkin.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibrarySidebarToggleGroupSkin.java
@@ -44,9 +44,13 @@ public class LibrarySidebarToggleGroupSkin extends
         final ToggleButton categoryButton = createSidebarToggleButton(category.getName());
         // TODO: store category ID in shortcut.info?
         categoryButton.setId(
-                String.format("applications-%s", SidebarToggleGroupBaseSkin.getToggleButtonId(category.getId())));
+                String.format("applications-%s", LibrarySidebarToggleGroupSkin.getToggleButtonId(category.getId())));
         categoryButton.setOnMouseClicked(event -> getControl().setSelectedElement(category));
 
         return categoryButton;
+    }
+
+    public static String getToggleButtonId(String categoryId) {
+        return String.format("%s-library-button", categoryId.toLowerCase().replace('.', '-'));
     }
 }
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibrarySidebarToggleGroupSkin.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibrarySidebarToggleGroupSkin.java
@@ -43,10 +43,19 @@ public class LibrarySidebarToggleGroupSkin extends
     protected ToggleButton convertToToggleButton(ShortcutCategoryDTO category) {
         final ToggleButton categoryButton = createSidebarToggleButton(category.getName());
         // TODO: store category ID in shortcut.info?
-        categoryButton.setId(
-                String.format("library-%s", SidebarToggleGroupBaseSkin.getToggleButtonId(category.getId())));
+        categoryButton.setId(LibrarySidebarToggleGroupSkin.getToggleButtonId(category.getId()));
         categoryButton.setOnMouseClicked(event -> getControl().setSelectedElement(category));
 
         return categoryButton;
+    }
+
+    /**
+     * Creates a button ID which can be used e.g. to assign icons via CSS based on the category ID
+     *
+     * @param categoryId The category ID which should be used
+     * @return The created button ID
+     */
+    public static String getToggleButtonId(String categoryId) {
+        return String.format("library-%s", SidebarToggleGroupBaseSkin.getToggleButtonId(categoryId));
     }
 }
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibrarySidebarToggleGroupSkin.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibrarySidebarToggleGroupSkin.java
@@ -44,13 +44,9 @@ public class LibrarySidebarToggleGroupSkin extends
         final ToggleButton categoryButton = createSidebarToggleButton(category.getName());
         // TODO: store category ID in shortcut.info?
         categoryButton.setId(
-                String.format("applications-%s", LibrarySidebarToggleGroupSkin.getToggleButtonId(category.getId())));
+                String.format("library-%s", SidebarToggleGroupBaseSkin.getToggleButtonId(category.getId())));
         categoryButton.setOnMouseClicked(event -> getControl().setSelectedElement(category));
 
         return categoryButton;
-    }
-
-    public static String getToggleButtonId(String categoryId) {
-        return String.format("%s-library-button", categoryId.toLowerCase().replace('.', '-'));
     }
 }
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
@@ -35,6 +35,7 @@ import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Optional;
 
 import static org.phoenicis.configuration.localisation.Localisation.tr;
 
@@ -90,15 +91,14 @@ public class AppsController {
                 for (CategoryDTO category : categories) {
                     cssBuilder.append(
                             "#" + ApplicationSidebarToggleGroupSkin.getToggleButtonId(category.getId()) + "{\n");
-                    URI categoryIcon = category.getIcon();
-                    if (categoryIcon == null) {
-                        cssBuilder
-                                .append("-fx-background-image: url('/org/phoenicis/javafx/views/common/phoenicis.png');\n");
-                    } else {
-                        cssBuilder.append("-fx-background-image: url('" + categoryIcon + "');\n");
-                    }
+                    final String categoryIconPath = Optional.ofNullable(category.getIcon())
+                            .map(categoryIcon -> categoryIcon.toString())
+                            .orElse("/org/phoenicis/javafx/views/common/phoenicis.png");
+
+                    cssBuilder.append(String.format("-fx-background-image: url('%s');\n", categoryIconPath));
                     cssBuilder.append("}\n");
                 }
+
                 String css = cssBuilder.toString();
                 Path temp = Files.createTempFile("defaultCategoryIcons", ".css").toAbsolutePath();
                 File tempFile = temp.toFile();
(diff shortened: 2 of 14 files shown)


plata:
Currently, I cannot see any icon. I have the same issue on master (also for the apps).

phoenicis-library/src/main/java/org/phoenicis/library/dto/ShortcutDTO.java
```diff
@@ -34,6 +34,18 @@
...
+    // overwritten if multiple shortcuts from different sources use the same category
+    // this could lead to a strange behavior in the following case:
+    // an app A with category C and category icon I(A) from repository R1 is installed
+    // if the theme does not specify an icon, I(A) is shown in the library
+    // an app B with category C and category icon I(B) from repository R2 is installed such that I(A) != I(B) (even
+    // though the category C is the same)
+    // if the theme does not specify an icon, I(B) is shown in the library
+    // app B is uninstalled
+    // if the theme does not specify an icon, I(B) is still shown in the library even though the situation is the same
+    // as after installing app A
```
madoar: Can we format improve this comment somehow? It is really hard to read.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
```diff
@@ -64,4 +77,38 @@ private void showError(Exception e) {
...
+                for (CategoryDTO category : categories) {
+                    cssBuilder.append(
+                            "#" + ApplicationSidebarToggleGroupSkin.getToggleButtonId(category.getId()) + "{\n");
+                    URI categoryIcon = category.getIcon();
+                    if (categoryIcon == null) {
+                        cssBuilder
+                                .append("-fx-background-image: url('/org/phoenicis/javafx/views/common/phoenicis.png');\n");
+                    } else {
+                        cssBuilder.append("-fx-background-image: url('" + categoryIcon + "');\n");
+                    }
```
madoar: The formatting looks really strange here. How about we change it to something like:

```java
final String categoryIconPath = Optional.ofNullable(category.getIcon())
    .map(categoryIcon -> categoryIcon.toString())
    .orElse("/org/phoenicis/javafx/views/common/phoenicis.png");

cssBuilder.append(String.format("-fx-background-image: url('%s');\n", categoryIconPath));
```

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/library/LibraryController.java
```diff
@@ -68,6 +78,48 @@ private void showError(Exception e) {
...
+     * after the app has been installed (e.g. the repository might have been deleted).
+     * - It should not be required to fetch the repository to use the library.
+     *
+     * Therefore, use the following approach:
+     * - When installing an app, store the icon for its category together with the shortcut (i.e. in the library).
+     * - Use this icon as fallback if the theme does not specify an icon for this category.
+     *
+     * @param categories list of shortcut categories
+     */
+    private void setDefaultLibraryCategoryIcons(List<ShortcutCategoryDTO> categories) {
```
madoar: This method seems to be quite similar to `setDefaultCategoryIcons` above. Can we maybe merge both methods into one single method?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibrarySidebarToggleGroupSkin.java
```diff
@@ -43,10 +43,19 @@ public LibrarySidebarToggleGroupSkin(LibrarySidebarToggleGroup control) {
     protected ToggleButton convertToToggleButton(ShortcutCategoryDTO category) {
         final ToggleButton categoryButton = createSidebarToggleButton(category.getName());
         // TODO: store category ID in shortcut.info?
```
madoar: @plata does it make sense to include the content of this TODO in this PR as well?

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
```diff
@@ -87,6 +88,16 @@ public void createShortcut(ShortcutDTO shortcutDTO) {
             } catch (IOException | IllegalArgumentException e) {
                 LOGGER.warn("Error while creating shortcut icon", e);
             }
+            try {
+                if (shortcutDTO.getCategoryIcon() != null) {
+                    File file = createFileWithFallback(shortcutDTO.getCategoryIcon(), "phoenicis.png");
+                    if (file.exists()) {
```
madoar: Is it possible at all that this file does not exist? If this is the case should we maybe throw/show an error?

madoar:
> Currently, I cannot see any icon. I have the same issue on master (also for the apps).

Can you narrow down the reason for this? Is it because of something wrong with the CSS? (we had this in the past)

phoenicis-library/src/main/java/org/phoenicis/library/dto/ShortcutDTO.java
plata: Any suggestion?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
plata: I just copied the method from `MainController`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/library/LibraryController.java
plata: Technically yes, the question is where it should be located.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibrarySidebarToggleGroupSkin.java
plata: I think it will get too huge.

phoenicis-library/src/main/java/org/phoenicis/library/dto/ShortcutDTO.java
madoar: Use a multiline comment, i.e.:

```java
/**
 *
 */
```

and add bullet points to clarify that these are different conditions

phoenicis-library/src/main/java/org/phoenicis/library/ShortcutManager.java
plata: Not quite sure. Normally not. I applied the same logic as for the shortcut icon though.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/apps/AppsController.java
madoar: I know I just think it looks better this way and you are already touching the code as it is...

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/library/LibraryController.java
madoar: I think we should either add an abstract parent class or an interface containing the method.

plata:
The CSS looked ok to to me. It's more like it doesn't get applied for whatever reason. I'm currently unsure how to debug this. Where does the `ThemeManager` actually say that the CSS files should be used?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/library/LibraryController.java
madoar: An interface may be better because a parent class should be extended by all feature panel controllers

madoar:
The stylesheets are applied here:

https://github.com/PhoenicisOrg/phoenicis/blob/e81b8b72ef2a8a81ba8b14ab6dd98a6e6bba4a5a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/themes/ThemeManager.java#L74-L84

The selection if a theme has been selected is done in line 74 - 77. If no theme has been selected the default theme is used in line 81.

plata:
Isn't this only for the `description.css`? I still don't see where `this.stylesheets` is used.

madoar:
In:

https://github.com/PhoenicisOrg/phoenicis/blob/e8dc57dd87ce0a218437f77f2f68492fbcbe49d9/phoenicis-javafx/src/main/java/org/phoenicis/javafx/views/common/PhoenicisScene.java#L31

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/library/LibraryController.java
plata: I don't like this. In my opinion, this is not some functionality that is implemented by the controllers so it should not be an interface nor an abstract class. Logically it should be in the `ThemeManager` but I don't want to have the DTOs there.
I would suggest to leave it as it is right now.

plata:
Everything works as expected now.

phoenicis-library/src/main/java/org/phoenicis/library/LibraryManager.java
```diff
@@ -80,8 +80,12 @@ public LibraryManager(String shortcutDirectory, ObjectMapper objectMapper) {
         List<ShortcutCategoryDTO> shortcuts = new ArrayList<>();
         for (Map.Entry<String, List<ShortcutDTO>> entry : categoryMap.entrySet()) {
             entry.getValue().sort(ShortcutDTO.nameComparator());
-            ShortcutCategoryDTO category = new ShortcutCategoryDTO.Builder().withId(entry.getKey())
-                    .withName(entry.getKey()).withShortcuts(entry.getValue()).build();
+            ShortcutCategoryDTO category = new ShortcutCategoryDTO.Builder()
+                    .withId(entry.getKey())
+                    .withName(entry.getKey())
+                    .withShortcuts(entry.getValue())
+                    .withIcon(entry.getValue().get(0).getCategoryIcon()) // choose one category icon
```
madoar: Is it possible that the shortcuts list is empty?

plata: No. The entry in the `categoryMap` is created only if there is at least one `ShortcutDTO`.

PR #2150, opened 2019-12-29
Load repository only if required

### LINKED ISSUE #2037: Engine settings are loaded at startup
This should not be required. The engines and engine settings should be loaded only if required (container tab).

plata:
fixes #2037

load repository only if it is really required, i.e.
- if one of the following tabs is opened
and
- if the repository has not been loaded already (avoid reload by opening the tab several times)

tabs:
- apps (apps are stored in the repository)
- containers (engine settings etc.)

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
@@ -19,6 +19,7 @@
 package org.phoenicis.javafx.controller;
 
 import javafx.application.Platform;
+import javafx.beans.value.ChangeListener;
 import org.phoenicis.javafx.components.common.skin.SidebarToggleGroupBaseSkin;
 import org.phoenicis.javafx.components.installation.control.InstallationsFeaturePanel;
 import org.phoenicis.javafx.controller.apps.AppsController;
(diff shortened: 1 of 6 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
```diff
@@ -80,12 +80,21 @@ public MainController(String applicationName,
...
+        this.mainWindow.getApplicationsTab().selectedProperty().addListener((observable, oldValue, newValue) -> {
+            if (newValue && !repositoryManager.isRepositoryLoaded()) {
+                repositoryManager.triggerRepositoryChange();
+            }
+        });
+        this.mainWindow.getContainersTab().selectedProperty().addListener((observable, oldValue, newValue) -> {
+            if (newValue && !repositoryManager.isRepositoryLoaded()) {
+                repositoryManager.triggerRepositoryChange();
+            }
+        });
```
madoar: I think you should add a comment here describing what these listeners do/should do

plata:
It works as intended. I have the same issue here like mentioned before (and on master): the category buttons do not work. Either they have no icon or the are not shown at all. It seems to be a timing issue. I think it's unrelated to this PR but feel free to try if it works for you.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/controller/MainController.java
```diff
@@ -80,12 +80,29 @@ public MainController(String applicationName,
...
```
madoar: You could think about extracting the listener to a local variable:

```java
final ChangeListener<Boolean> tabListener = 
        (observable, oldValue, newValue) -> {
            if (newValue && !repositoryManager.isRepositoryLoaded()) {
                repositoryManager.triggerRepositoryChange();
            }
        };

this.mainWindow.getApplicationsTab().selectedProperty().addListener(tabListener);
this.mainWindow.getContainersTab().selectedProperty().addListener(tabListener);
```

PR #2156, opened 2020-01-01
Travis: fix GitHub pages build

plata:


plata:
Issue must be fixed in github-pages gem.

PR #2159, opened 2020-01-03
FileUtilities#remove(): do nothing if file does not exist

plata:
fixes https://github.com/PhoenicisOrg/scripts/issues/1162

It seems that this behavior was intended anyways because we use
```Java
Files.deleteIfExists(fileToDelete.toPath());
```

### DIFF
--- a/phoenicis-tools/src/main/java/org/phoenicis/tools/files/FileUtilities.java
+++ b/phoenicis-tools/src/main/java/org/phoenicis/tools/files/FileUtilities.java
@@ -165,12 +165,12 @@ public class FileUtilities extends FilesManipulator {
     public void remove(String path) throws IOException {
         final File fileToDelete = new File(path);
 
+        assertInDirectory(fileToDelete);
+
         if (!fileToDelete.exists()) {
             LOGGER.debug(String.format("Cannot remove file or directory: path \"%s\" does not exist", path));
             return;
         }
-
-        assertInDirectory(fileToDelete);
 
         if (fileToDelete.isDirectory()) {
             FileUtils.deleteDirectory(fileToDelete);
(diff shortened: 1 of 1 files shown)


madoar:
What would happen if the to be deleted file is a directory?

Zemogiter:
@madoar Probably the same as removing the file because https://github.com/PhoenicisOrg/scripts/issues/1124 is still open

plata:
If the directory doesn't exist, nothing. Otherwise:
```Java
FileUtils.deleteDirectory(fileToDelete);
```

plata:
Which order is wrong?

phoenicis-tools/src/main/java/org/phoenicis/tools/files/FileUtilities.java
```diff
@@ -162,7 +166,8 @@ public void remove(String path) throws IOException {
         final File fileToDelete = new File(path);
 
         if (!fileToDelete.exists()) {
-            throw new IllegalArgumentException(String.format("Path \"%s\" does not exist", path));
+            LOGGER.debug(String.format("Cannot remove file or directory: path \"%s\" does not exist", path));
+            return;
         }
 
         assertInDirectory(fileToDelete);
```
madoar: I think the assertion should then be moved before `if (!fileToDelete.exists()) { .. }`

plata: Why should it matter?

madoar: The order seems wrong like this. Let us say the script wants to execute `remove("/home/<your/home>/important-file");` then I wouldn't expect a debug message that says the file does not exist but an error message that tells me that the script tries to access files it has no access to

madoar: If you want we can handle this in another PR. I just want to talk about this now that it is still fresh on my mind

plata: Ok.

PR #2161, opened 2020-01-04
Improve the logging information for "GitRepository"

madoar:
This PR improves the logging information for `GitRepository`

### DIFF
--- a/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
+++ b/phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
@@ -69,21 +69,22 @@ public class GitRepository implements Repository {
     private void cloneOrUpdateWithLock() throws RepositoryException {
         synchronized (mutex) {
             try {
-                LOGGER.info("Begin fetching process of '{}' to '{}'", repositoryUri, localFolder.getAbsolutePath());
+                LOGGER.info("Begin fetching process of '{}' to '{}'",
+                        this.repositoryUri, this.localFolder.getAbsolutePath());
 
                 boolean lockFileExists = this.lockFile.exists();
 
                 // check that the repository lock file exists
                 if (!lockFileExists) {
-                    LOGGER.info("Creating lock file '{}' for git-repository '{}'", lockFile.getAbsolutePath(),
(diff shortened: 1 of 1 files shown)


phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -100,90 +100,73 @@ private void cloneOrUpdate() throws RepositoryException {
...
-        return result;
     }
 
     @Override
     public void onDelete() {
         try {
-            FileUtils.deleteDirectory(this.localFolder);
+            FileUtils.deleteDirectory(localFolder);
 
-            LOGGER.info("Deleted " + this);
```
plata: Wasn't the purpose of logging `this` to avoid that the repository URL etc. must be printed every time? I thought this was the way how it was implemented before?

madoar: Yes it was, but in my opinion `this` (or more specific `this.toString()`) returns too much information. Therefore I thought about changing the output to something more specific and self-descriptive

plata: Ok.

phoenicis-repository/src/main/java/org/phoenicis/repository/types/GitRepository.java
```diff
@@ -69,19 +69,19 @@ private File createRepositoryLocation(String cacheDirectoryPath) {
...
             try {
-                LOGGER.info("Begin fetching process of " + this);
+                LOGGER.info("Begin fetching process of '{}' to '{}'", repositoryUri, localFolder);
 
                 boolean lockFileExists = this.lockFile.exists();
 
                 // check that the repository lock file exists
                 if (!lockFileExists) {
-                    LOGGER.info("Creating lock file for " + this);
+                    LOGGER.info("Creating lock file '{}' for git-repository '{}'", lockFile, repositoryUri);
```
plata:
```suggestion
                    LOGGER.info("Creating lock file '{}' for git-repository '{}'", lockFile.getAbsolutePath(), repositoryUri);
```

Issue #2115, opened 2019-10-29
Package creation failure


Zemogiter:
### Expected behavior
Correct package creation
### Actual behavior
I get the .deb file that is 73.6 kilobytes which can't be right. Here is the output of the bash file:
```
Warning: Version not specified. Reading from pom.xml
Using 5.0-SNAPSHOT
--2019-10-29 11:48:53--  http://download2.gluonhq.com/jpackager/11/jdk.packager-linux.zip
Translacja download2.gluonhq.com (download2.gluonhq.com)... 178.128.135.159
Łączenie się z download2.gluonhq.com (download2.gluonhq.com)|178.128.135.159|:80... połączono.
Żądanie HTTP wysłano, oczekiwanie na odpowiedź... 200 OK
Długość: 674385 (659K) [application/zip]
Zapis do: `jdk.packager-linux.zip'

jdk.packager-linux.zip                                                                                   100%[===============================================================================================================================================================================================================================================================================>] 658,58K   939KB/s    w 0,7s     

2019-10-29 11:48:54 (939 KB/s) - zapisano `jdk.packager-linux.zip' [674385/674385]

Archive:  jdk.packager-linux.zip
  inflating: jdk.packager.jar        
  inflating: jpackager               
WARNING: argument [linux-bundle-name] is not supported for current configuration.
Creating app bundle: /home/jonasz/Pobrane/phoenicis-master/phoenicis-dist/src/scripts/../../target/packages/PhoenicisPlayOnLinux
Module org.graalvm.truffle does not exist.
Module javafx.web does not exist.
Module javafx.base does not exist.
Module javafx.graphics does not exist.
Module javafx.controls does not exist.
Module javafx.media does not exist.
"Adding modules: [jdk.crypto.ec, java.base, javafx.base, javafx.web, javafx.media, javafx.graphics, javafx.controls, java.naming, java.sql, java.scripting, jdk.scripting.nashorn, jdk.internal.vm.ci, org.graalvm.truffle, java.management] to runtime image."
Exception: jdk.tools.jlink.plugin.PluginException: java.lang.module.FindException: Hash of java.xml (46fd019860d97f5811ffc5b51eb9de95a713c64000856d637727c495cdc7678c) differs to expected hash (12a209d077d43c90a5ce65d64027cc8468fe5de54e9b1b56138ce9f58aebef28) recorded in java.base
Error: Bundler "Linux Application Image" (linux.app) failed to produce a bundle.
[...]

PR #2173, opened 2020-02-09
Update build page, using JDK 13

Zemogiter:
Solves the https://github.com/PhoenicisOrg/phoenicis/issues/2115 so .deb package is working.
As for the optional step, I was unable to find a working link so I've removed it for now.

### DIFF
--- a/docs/_docs/Developers/build.md
+++ b/docs/_docs/Developers/build.md
@@ -10,16 +10,8 @@ toc: true
 * Install the dependencies:
 ```
 sudo apt update
-sudo apt install git maven openjdk-11-jdk fakeroot
+sudo apt install git maven openjdk-13-jdk fakeroot
 ```
-
-* Optional: if your distribution does not support Java 11+ **or** if you need to create packages, grab the latest JDK 
-```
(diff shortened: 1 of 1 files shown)


madoar:
@Zemogiter essentially you are saying that you need Java 13 to successfully run the build tool?

madoar:
I have two issues against this an update to Java 13:
- Java 13 is not easily available on all distributions. Especially older ones may require custom package repositories
- Java 13 is not a LTS version. This means that we would need to update the guide to Java 14 in around 2 months.

On another note it may be worth thinking about using Java 14 after it is released because it is planned to come with the new packaging tool, see also [here](https://openjdk.java.net/jeps/343).

Zemogiter:
@madoar yes I need java 13 to run the build tool and according to this:
https://en.wikipedia.org/wiki/Java_version_history
Java 14 won't be LTS either to it's kind of a moot point and 2 months isn't a short time.

madoar:
> ... according to this:
https://en.wikipedia.org/wiki/Java_version_history
Java 14 won't be LTS either to it's kind of a moot point and 2 months isn't a short time.

Yes, you are right about that. I mainly wanted to say that Java 14 comes brings additional benefits to Phoenicis (i.e. the final release of the packaging tool), which could be worth it to think about using Java 14 at least for building the installer packages.

plata:
@qparis should take a look at the packaging with Java 14.

plata:
We should update this to Java 17.

PR #2175, opened 2020-02-16
Downloader: close URL connection after download

### LINKED ISSUE #1807: Local installation: download wine?os=linux, takes time to start
### Expected behavior
Running local installation loads all necessary files fast.

### Actual behavior
Running Local install takes ages to "download wine?os=linux", why is that? (it starts very slowly)

### Steps to reproduce
Run Local installer, select arch and installer file.

### System information
- Version 5.0-alpha-2, GIT revision ends with badce66c (2018.12.31)
- Linux Mint KDE, Linux lati 4.15.5-041505-generic #201802221031 SMP Thu Feb 22 15:32:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux, 18.sth.
- san@lati ~ $ java -version
openjdk version "1.8.0_191"
OpenJDK Runtime Environment (build 1.8.0_191-8u191-b12-0ubuntu0.16.04.1-b12)
OpenJDK 64-Bit Server VM (build 25.191-b12, mixed mode)

plata:
fixes #1807

The problem was that two connections for the Wine versions json are opened. The first to check if an update is available (`Downloader#isUpdateAvailable()`), the second to actually download the file (`Downloader#get()`). If e.g. `Downloader#isUpdateAvailable()` does not open a connection but returns `true` immediately, the download in `Downloader#get()` starts directly.

I'm not completely sure what happens in the background but my assumption is that the underlying socket is closed by the `disconnect` such that the next connection can setup a new socket immediately.

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
@@ -231,18 +231,15 @@ public class LibraryFeaturePanel extends FeaturePanel<LibraryFeaturePanel, Libra
                 .withMessage(tr("Are you sure you want to uninstall {0}?", shortcutName))
                 .withOwner(getScene().getWindow())
                 .withResizable(true)
-                .withYesCallback(() -> {
-                    getShortcutManager().uninstallFromShortcut(shortcut, e -> {
-                        final ErrorDialog errorDialog = ErrorDialog.builder()
-                                .withMessage(tr("Error while uninstalling {0}", shortcutName))
-                                .withException(e)
-                                .withOwner(getScene().getWindow())
-                                .build();
+                .withYesCallback(() -> getShortcutManager().uninstallFromShortcut(shortcut, e -> {
+                    final ErrorDialog errorDialog = ErrorDialog.builder()
+                            .withMessage(tr("Error while uninstalling {0}", shortcutName))
+                            .withException(e)
+                            .withOwner(getScene().getWindow())
+                            .build();
 
-                        errorDialog.showAndWait();
-                    });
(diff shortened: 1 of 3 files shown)


madoar:
Alternatively we could also make `PhoenicisUrlConnection` implement [`AutoCloseable`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/AutoCloseable.html). Then we could automatically close the connection by using:

```java
try (PhoenicisUrlConnection connection = ...) {
   ...
}
```

plata:
It's probably a good idea to have `PhoenicisUrlConnection` behave similarly to `HttpURLConnection`. So I would keep the `disconnect` explicit.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
```diff
@@ -231,15 +231,18 @@ public void uninstallShortcut(ShortcutDTO shortcut) {
...
+                    getShortcutManager().uninstallFromShortcut(shortcut, e -> {
+                        final ErrorDialog errorDialog = ErrorDialog.builder()
+                                .withMessage(tr("Error while uninstalling {0}", shortcutName))
+                                .withException(e)
+                                .withOwner(getScene().getWindow())
+                                .build();
+
+                        errorDialog.showAndWait();
+                    });
+                    // TODO: how can the feature panel close itself?
```
madoar: Do you mean the feature panel or the corresponding details panel?

plata: The corresponding details panel. It's a bit confusing because the buttons are on the details panel but the control is in the feature panel class.

PR #2176, opened 2020-02-16
Close shortcut details panel after uninstall

### LINKED ISSUE #1891: Close shortcut details panel after uninstall

plata:
fixes #1891

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
@@ -231,15 +231,18 @@ public class LibraryFeaturePanel extends FeaturePanel<LibraryFeaturePanel, Libra
                 .withMessage(tr("Are you sure you want to uninstall {0}?", shortcutName))
                 .withOwner(getScene().getWindow())
                 .withResizable(true)
-                .withYesCallback(() -> getShortcutManager().uninstallFromShortcut(shortcut, e -> {
-                    final ErrorDialog errorDialog = ErrorDialog.builder()
-                            .withMessage(tr("Error while uninstalling {0}", shortcutName))
-                            .withException(e)
-                            .withOwner(getScene().getWindow())
-                            .build();
+                .withYesCallback(() -> {
+                    getShortcutManager().uninstallFromShortcut(shortcut, e -> {
+                        final ErrorDialog errorDialog = ErrorDialog.builder()
+                                .withMessage(tr("Error while uninstalling {0}", shortcutName))
+                                .withException(e)
+                                .withOwner(getScene().getWindow())
+                                .build();
 
-                    errorDialog.showAndWait();
-                }))
+                        errorDialog.showAndWait();
+                    });
+                    // TODO: how can the feature panel close itself?
+                })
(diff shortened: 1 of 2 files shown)


plata:
@madoar I need your help regarding this. How can a `FeaturePanel` close itself? Is it possible to unify the shortcut uninstall in both classes?

madoar:
`FeaturePanel` or the details panels?

plata:
The corresponding details panel. It's a bit confusing because the buttons are on the details panel but the control is in the feature panel class.

madoar:
The closing of the details panel is done using the `selectXYZ` fields, for example: 

https://github.com/PhoenicisOrg/phoenicis/blob/1b47d90cfcc8c1a8662af353ac95e20c210563e8/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/application/skin/ApplicationsFeaturePanelSkin.java#L52

The decision whether the details panel is shown is then done using a `When` object:

https://github.com/PhoenicisOrg/phoenicis/blob/1b47d90cfcc8c1a8662af353ac95e20c210563e8/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/application/skin/ApplicationsFeaturePanelSkin.java#L204-L205

plata:
How can I do this? Just setting it to `null` gives "A bound value cannot be set".

madoar:
You are right it is a binding and not a property, which is set in:

https://github.com/PhoenicisOrg/phoenicis/blob/1b47d90cfcc8c1a8662af353ac95e20c210563e8/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/application/skin/ApplicationsFeaturePanelSkin.java#L63

To clear the value you need to clear the selection of the list widget.

I think it would be worth to change the approach for all feature panels to the one used for the library panel. In the `LibraryFeaturePanelSkin` a combination of two properties is used to be more granular:

https://github.com/PhoenicisOrg/phoenicis/blob/1b47d90cfcc8c1a8662af353ac95e20c210563e8/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibraryFeaturePanelSkin.java#L50 

`selectedDetailsPanel` specifies whether a details panel should be shown and if yes which one

https://github.com/PhoenicisOrg/phoenicis/blob/1b47d90cfcc8c1a8662af353ac95e20c210563e8/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java#L101
 
`selectedShortcut` specifies the shortcut that should be shown

Both values are then combined to create a binding for the shown details panel:

https://github.com/PhoenicisOrg/phoenicis/blob/1b47d90cfcc8c1a8662af353ac95e20c210563e8/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibraryFeaturePanelSkin.java#L199-L210

plata:
Could you try this? Your are far more familiar with the GUI classes then me. Feel free to edit my branch (you should be allowed to do so).

madoar:
I am currently low on time and therefore don't know when I will be able to do so. If this is fine for you I can take a look it.

plata:
Ok.

madoar:
It should now be possible to close every detail panels by calling `closeDetailsView();` in the corresponding `FeaturePanel` class

PR #2189, opened 2020-03-12
Enhance the Management of the State of the currently open Details Panel

### LINKED ISSUE #1891: Close shortcut details panel after uninstall

madoar:
This PR:
- adds a `OpenDetailsPanel` interface
- adds a general `None` class which can be used for all feature panels to indicate that the details panel is closed
- adds specific implementations of the `OpenDetailsPanel` interface for all feature panels
- changes the SwitchBinding implementation to be compatible with the new DetailsPanelAction architecture
- deletes the unused LibraryView class

This PR fixes #1891 and supersedes #2176


madoar:
@plata what do you think about this approach?

The idea here is to use so called `DetailsPanelAction`s to perform changes to the details panels of all feature panels. For now I've changed the library feature panel to use these new `DetailsPanelAction`s instead of the previously used enums.

If you are fine with this approach I would continue changing the other feature panels next in this PR.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/common/actions/DetailsPanelAction.java
```diff
@@ -0,0 +1,8 @@
+package org.phoenicis.javafx.components.common.actions;
+
+/**
+ * An action that influences the currently shown details panel and its content
+ */
+public interface DetailsPanelAction {
+    // nothing
+}
```
plata: missing line break at EOF

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
```diff
@@ -100,6 +102,11 @@
      */
     private final ObjectProperty<ShortcutDTO> selectedShortcut;
 
+    /**
+     * The currently selected details panel action
+     */
+    private final ObjectProperty<DetailsPanelAction> selectedDetailsPanelAction;
```
plata: What is the meaning of "currently selected action"? If e.g. the `CloseAction` happens, it is over as soon as the event has been executed. So how could it possibly be "currently selected"?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
```diff
@@ -394,4 +402,26 @@ public ShortcutDTO getSelectedShortcut() {
...
+    public void setSelectedDetailsPanelAction(DetailsPanelAction detailsPanelAction) {
+        this.selectedDetailsPanelAction.set(detailsPanelAction);
+    }
+
+    /**
+     * Closes the currently opened details panel
+     */
+    public void closeDetailsPanel() {
+        // deselect the currently selected shortcut
+        setSelectedShortcut(null);
```
plata: Shouldn't we have `deselectShortcut` to be consistent?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibraryFeaturePanelSkin.java
```diff
@@ -110,9 +73,10 @@ public LibraryFeaturePanelSkin(LibraryFeaturePanel control) {
 
         sidebar.setOnCreateShortcut(() -> {
             // deselect the currently selected shortcut
-            this.selectedListWidgetElement.setValue(null);
+            getControl().setSelectedShortcut(null);
```
plata: see above

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/skin/LibraryFeaturePanelSkin.java
```diff
@@ -189,9 +153,34 @@ public LibraryFeaturePanelSkin(LibraryFeaturePanel control) {
...
 
-        this.selectedListWidgetElement.bindBidirectional(combinedListWidget.selectedElementProperty());
+                getControl().setSelectedShortcut(selectedItem);
+                getControl().setSelectedDetailsPanelAction(new ShortcutDetailsAction(selectedItem));
+
+                if (event.getClickCount() == 2) {
+                    getControl().runShortcut(selectedItem);
+                }
+            } else {
+                getControl().setSelectedShortcut(null);
```
plata: see above

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
madoar: Maybe the word `action` is not exactly fitting...
The idea is that the `action` defines the currently opened details panel.
Do you have a better fitting name?

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
madoar: We can add that but internally it would still call `this.setSelectedShortcut(null)`. I'm not sure how useful such one-liner methods are...

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
plata: `OpenDetailsPanel`? Then it would be `None` instead of `CloseAction`.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
plata: Ok.

plata:
Will this fix #1891 and supersede #2176?

madoar:
Yes to both!

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
madoar: Sounds good to me. I will rename the classes accordingly

madoar: Done

madoar: Do you also have a suggestion for the package name?

plata:
Can you add this to the PR description then? And also update it in general to match the current naming.

phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
plata: Maybe something like `panelState`?

madoar: camel case for the package name?

plata: I think so, yes. Do we have something else in another place?

madoar: I don't know. I just don't think it is standard to use camel case in package names. It looks weird to me

plata: I don't mind really. If it's more standard to use kebab case, feel free. Do you have a source (just interested)?

madoar: Kebab case is not allowed, because `-` is an illegal symbol for Java package names.

plata: Ok then how should it be?

madoar: I don't know actually. we could go with one of the following:
- `state`
- `panel.state`
- `detail.panel.state`
- `panelstate`
- `detailstate`

I have no source for arguing against camel case in package names I have just never seen it being used.

plata: I wouldn't use the dot because it normally indicates the directory structure. Choose one of the last two (whatever you prefer).

madoar: Done. I went for `panelstates`

madoar:
Somehow I didn't receive an email with the approve...

madoar:
@plata sorry if this blocked you in #2176

plata:
No problem. If I understand correctly, that PR is now obsolete anyways.

PR #2204, opened 2020-04-17
CLI: fix app installation

plata:


plata:
Two things we should do after merging this:
- Unify the code with `ApplicationInformationPanelSkin`.
- Change `getType` etc. to use only the ID. As this has an impact on the scripts, we should maybe postpone it until after the next release.

madoar:
This PR has a merge conflict.

How do you want to track the TODOs?

plata:
I will create new issues for the TODOs.

plata:
@madoar could you review this?

PR #2205, opened 2020-04-18
Fix flatpak

plata:
Install only .jar's in `phoenicis-dist/target/lib` to avoid issues with split packages etc.

### DIFF
--- a/docs/_docs/Developers/packages.md
+++ b/docs/_docs/Developers/packages.md
@@ -15,7 +15,14 @@ The built packages are located in the target/packages folder:
 Required extensions:
 - org.freedesktop.Platform.Compat.i386
 - org.freedesktop.Platform.GL32.nvidia-415-25
+- org.freedesktop.Sdk.Extension.openjdk11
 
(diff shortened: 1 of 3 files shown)


plata:
Phoenicis starts but I get
```
org.graalvm.polyglot.PolyglotException: java.lang.AbstractMethodError: Receiver class org.graalvm.compiler.hotspot.management.HotSpotGraalManagement does not define or inherit an implementation of the resolved method 'abstract void initialize(org.graalvm.compiler.hotspot.HotSpotGraalRuntime, org.graalvm.compiler.hotspot.GraalHotSpotVMConfig)' of interface org.graalvm.compiler.hotspot.HotSpotGraalManagementRegistration
```

Apparently, the behavior is also different depending if I use an absolute or relative module path. Any ideas?

plata:
The issue seems to be related to the OpenJDK extension. I've updated the documentation accordingly. It should not stop us from merging this (it works with the correct OpenJDK version).

plata:
I don't think so. I red somewhere that only LTS versions are supported officially, though.

PR #2222, opened 2020-06-12
JavaFX: parse command line arguments

plata:
The `CLIController` is more or less copy+paste from `phoenicis-cli`.
Running an app works, installation not.
We need a way to integrate this in the design of the JavaFX application. One possibility could be that `MainController` offers a method `installApp`. @madoar any suggestions (feel free to edit in the PR directly)?

### DIFF
--- a/phoenicis-javafx/pom.xml
+++ b/phoenicis-javafx/pom.xml
@@ -31,6 +31,11 @@
     </properties>
     <dependencies>
         <dependency>
+            <groupId>com.github.jankroken</groupId>
+            <artifactId>commandline</artifactId>
+            <version>1.7.0</version>
+        </dependency>
+        <dependency>
             <groupId>org.phoenicis</groupId>
             <artifactId>phoenicis-configuration</artifactId>
             <version>${phoenicis.version}</version>
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/CLIController.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/CLIController.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright (C) 2015-2017 PÂRIS Quentin
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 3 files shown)


madoar:
I think it would be cool if it becomes possible to use the same Phoenicis application both from the terminal and as an UI application. I think essentially we will need to add a wrapper around our current JavaFX UI application that decides whether the UI should be started or the application should be run in terminal mode.

plata:
I have no idea how this could be implemented.

plata:
When trying to install an app, you just get a white window. Not really sure why this behaves differently than `ApplicationInformationPanelSkin`.

madoar:
Essentially I am suggesting to change `main` method in `JavaFXApplication` to something similar to:

```java
    public static void main(String[] args) {
        if (args.length == 0 || args[0] == "gui") {
            try {
                Application.launch(args);
            } catch (Exception e) {
                e.printStackTrace();
            }
        } else if (args[1] == "cli") {
            // execute phoenicis in CLI mode
        }
    }
}
```

plata:
Wouldn't that imply that `phoenicis-cli` is a JavaFX application?

madoar:
> Wouldn't that imply that phoenicis-cli is a JavaFX application?

What do you mean by a `JavaFX application`?
If you mean that it would include the JavaFX dependencies, then yes.
If you mean that it would automatically open a JavaFX window when starting the application, then no.

plata:
> If you mean that it would include the JavaFX dependencies, then yes.

This is what I meant. Do we want this?

madoar:
I do not know how others think about this, but in my opinion it is not a huge problem adding the JavaFX dependencies to the applications. I mean Java itself is normally bundled with a lot of "unnecessary" libraries (including Swing). I don't think that it is a big issue to add another library if it benefits us. In the scope of this PR I would say that adding JavaFX as a dependency to phoenicis-cli makes it much easier to maintain both applications. In addition it allows us distributing both the cli and the JavaFX client as a single application, which is also a huge benefit.

plata:
This is definitely something where I would like @qparis opinion.

PR #2226, opened 2020-06-14
JavaFX: simple/advanced mode

plata:
Library
Settings

### DIFF
--- a/docs/_docs/Developers/test-plan.md
+++ b/docs/_docs/Developers/test-plan.md
@@ -12,6 +12,8 @@ Before you start, ensure that you are in a clean environment:
 * checkout and pull the desired branch/revision from upstream `phoenicis`
 * do a clean build (`mvn clean package`)
 
+The test plan should be executed in the advanced mode.
+
 ## Applications
 * open "Applications" tab
 * check that several categories are shown
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibrarySidebar.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibrarySidebar.java
@@ -8,6 +8,7 @@ import javafx.collections.ObservableList;
 import org.phoenicis.javafx.components.common.control.ExtendedSidebarBase;
 import org.phoenicis.javafx.components.common.widgets.utils.ListWidgetType;
 import org.phoenicis.javafx.components.library.skin.LibrarySidebarSkin;
+import org.phoenicis.javafx.settings.JavaFxSettingsManager;
 import org.phoenicis.javafx.views.mainwindow.library.LibraryFilter;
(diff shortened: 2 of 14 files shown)


plata:
@Gcenx do you think this would be good for macOS users?

Gcenx:
@plata seems fine

As I'm not familiar with JavaFX this is more a question, usually you want your application to match the system theming by default but can be overridden by the user.

plata:
We have different seems (e.g. for KDE Breeze). It's not selected by default but you can choose it in the settings. Currently, there's nothing for Mac OS but it should not be too difficult to create a matching theme for it.

plata:
@madoar I would like to hear your opinion regarding this as well.

madoar:
Can you explain a bit more which features you want to place in the `advanced mode` and which in the `simple mode`?

In my opinion the details panel looks a bit strange without the detail information shown. Especially the `edit` button feels out of place to me.

plata:
- Simple mode: only the very basic functionality to install/uninstall and run an app + settings.
- Advanced mode: everything else

I wasn't quite sure about the edit button. I mainly kept it because you can also edit the command line arguments e.g. in Steam. However, I'm fine removing it if you prefer that.

madoar:
In my opinion it depends on whether the user is able to create new shortcuts in the `simple mode`. If he is able to create new shortcuts he should also be able to modify and delete them.

plata:
He's not (in my opinion).

madoar:
Then I think we should remove both the `delete` (of shortcuts) and the `edit` functionality.

plata:
We need the delete because otherwise you cannot uninstall any app.

madoar:
Yes, but we should ensure that the user can not accidentally delete the shortcut without removing the container because (s)he will not be able to remove the container afterwards without a working shortcut and entering advanced mode.

madoar:
Does the change between `simple mode` and `advanced mode` require a restart of the application?

plata:
That's not quite correct. You could have a situation where multiple apps share a container. So yes, there might be situations where you get a dangling container. However, I would suggest to wait if this is really something that happens in reality before deciding what to do about it.

plata:
Restart is required (e.g. the engines tab cannot be added/removed on the fly). At least currently.

plata:
@madoar can I merge this?

madoar:
I guess we can go with this change, but I'm kind of missing "repair" functionality in case an installation script fails. I think this will be the case more often than we expect, at least while the application is still in alpha/beta. Additionally it is still not really clear to me which functionality should be part of the `simple mode` and which of the `advanced mode`. Ideally we should specify this before we perform the split otherwise I'm almost certain that we won't move wrongly placed functionality later on.

plata:
Ok. I will merge this so that we have a basic working implementation. Can you open an issue/issues for the points you would like to discuss?

PR #2228, opened 2020-06-15
Flatpak: add OpenLDAP

plata:


plata:
from: https://github.com/flathub/net.lutris.Lutris/blob/beta/net.lutris.Lutris.yml

Not sure if we should also incorporate the other stuff like faudio.

plata:
Can we know somehow if they are required? It feels a bit strange for me to add something "because Lutris has it" (although I trust that they are there for a reason). For example vkd3d or FAudio: don't we handle those via Verbs?

madoar:
I don't really know what these "extensions" contain. The reason why I mentioned adding them as well is because of my previous problems on Linux Mint to get the sound working with wine. I remember that I got this working by installing additional libraries. It would be nice if these libraries are added to the flatpak as well :)

plata:
Yes, it's about those missing libraries. My problem is that I do not know how we could figure out which ones are needed. So either, we just install some and "hope" that they are required and sufficient. Or we just wait until something doesn't work and add the required library then.

madoar:
If I remember correctly I found the missing libraries by looking at https://www.codeweavers.com/support/wiki/diag

Maybe we should add all mentioned libraries?

plata:
I will create a new issue to discuss this topic.

PR #2229, opened 2020-06-16
Checkstyle action (reviewdog)

plata:
Codacy does not work reliably lately and the configuration is more difficult compared to GitHub actions.

### DIFF
--- a/.github/workflows/checkstyle.yml
+++ b/.github/workflows/checkstyle.yml
@@ -7,8 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Checkstyle job
     steps:
-    - name: Checkout
-      uses: actions/checkout@v2
+    - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


plata:
Codacy didn't even complain when it was wrong.

plata:
Yes.

This PR needs an update after the next version of the checkstyle action has been released (such that the action actually fails if it detects an error).

madoar:
@plata any news regarding this PR?

plata:
Should be good now, I think.

madoar:
Now the question is: does it work as intended?

plata:
We will only find out by keeping the GitHub action and Codacy for some time in parallel.

madoar:
Ok feel free to merge when you think the PR is ready :)

plata:
Can you approve?

madoar (approved):

PR #2277, opened 2020-09-26
Travis: create GitHub releases

plata:
- nightly release from master
- "real" release for tags

I'm not sure how to test this before merging.


madoar (changes requested):
(Credential-)Tokens should not be added in a hardcoded way in configuration files in the repository. Please add them as secrets under the repository settings instead.

plata:
I followed https://docs.travis-ci.com/user/deployment-v2/providers/releases/.

madoar (approved):
Ah, ok so it is not an universally usable token. It is bound to a user/service?

plata:
My understanding is that it's bound to the Phoenicis Bot.

madoar:
Do we still require this PR after merging the Github Action PRs?

plata:
No.

PR #2333, opened 2021-01-20
Remove confirmation dialog to close Phoenicis

plata:
fixes #2330


madoar:
Do we really want to remove the dialog by default? I think it would be better to provide an option in the settings where the user can specify how the application should behave? I don't have a problem changing the default behaviour but I'm not sure if nobody is interesting in a confirm dialog

plata:
Question is: What do we want to achieve with the confirmation dialog? How do other applications handle it?

I see different approaches to why applications have a close dialog:
- Doesn't want to be closed but run all the time (e.g. Steam).
- Avoid to lose data unintentionally (e.g. Firefox with multiple tabs).

Both don't apply to us so I see no reason to have it.

madoar (approved):
I agree

PR #2337, opened 2021-01-24
Add GitHub action for general CI

plata:
- [x] Ubuntu (Java 11)
- [x] Ubuntu (Java 15)
- [x] Mac OS (Java 11)
- [x] Flatpak

### DIFF
--- a/.github/workflows/ci-ubuntu-java11.yml
+++ b/.github/workflows/ci-ubuntu-java11.yml
@@ -1,0 +1,16 @@
+name: Java CI
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Build with Maven
+        run: mvn clean package
--- a/.github/workflows/ci-ubuntu-java11.yml
+++ b/.github/workflows/ci-ubuntu-java11.yml
@@ -14,3 +14,7 @@ jobs:
           java-version: 11
       - name: Build with Maven
         run: mvn clean package
+      - name: Create package
+        run: |
+          cd phoenicis-dist/src/scripts
+          bash phoenicis-create-package.sh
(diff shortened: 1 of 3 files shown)


madoar:
Is it possible to add "depends on" between different jobs of different workflow files?
It this is not possible I think it may be better to use a single workflow file for the builds.
This will later on allow us to link the different builds together for a release script.

plata:
I was thinking it might be easier to understand the files if they are smaller. Not sure about the release. Maybe it makes sense to have a separate file for the release anyways (e.g. to build only on a tag or once every night).

madoar:
I mean can we use the result of job A (for example a `.deb` file) in job B (for example a release-publish job) when A and B are defined in different files?

What I envision for Phoenicis is something similar to what I did in https://github.com/Rails-18xx/Rails/blob/master/.github/workflows/release.yml

plata:
I don't know if that's possible. In any case, you yml runs is only for releases as well. For me it would be fine to have the "general" CI (which would build for different JDKs, update the translation files, check GitHub pages etc) and then the release CI which only builds the packages for releases.

madoar:
I'm fine with that I'm just a little bit confused by the amount of workflow files we are creating. I would expect a smaller number, maybe one file for each of the following tasks:

* executing the phoenicis tests on all platforms (ubuntu and macos)
* update the translation files (I guess this should only happen when the tests succeed?)
* check GitHub pages (does this depend on the tests as well?)
* build the installers and release/publish them (I guess this should depend on the tests as well?)

The open question in my opinion is whether we can implement these workflows without duplicating the testing jobs in each of the workflows

plata:
I get your point. Will it run in parallel if we have the builds in one yml?

madoar:
I think all jobs that don't depend on each other are run in parallel. If you require a job to be run after another job you need to use `needs: [windows, ubuntu, macos]`, where `windows, ubuntu, macos` are three other jobs.

plata:
Now I added Java 15 here. Checks list looks exactly the same. Do you think it's better in one file?

madoar:
It depends on what you want to achieve. I think it depends on what we want to achieve. If the goal is only to run the tests I think it should be located in the same file as the other test jobs. If we want to use the job to build an executable for a release then it should be a in a corresponding workflow file.

plata:
Should be ready to merge now.

madoar (approved):

PR #2342, opened 2021-01-24
Add GitHub action to create "latest" release

plata:


madoar (approved):
Can we test this before merging?

plata:
I'm not sure. Honestly, I'm somewhat afraid to merge this blindly as well.

madoar:
As long as we create prereleases it should be fine. As far as I know they are private, i.e. only visible by members of the organisation. If it does not work as intended we can delete the prerelease

madoar:
@plata how do we want to continue with this PR?

plata:
I need some confidence to actually try it :P.

plata:
I'm not sure if this is still needed. I'm now uploading the artifact in the CI action.

PR #2343, opened 2021-01-24
Add GitHub action to lint Java files

plata:


madoar (approved):
Can we test this before merging?

plata:
There was a bug indeed but now it works.

Issue #2353, opened 2021-01-25
Not able to create a shortcut to a file outside of the wineprefix


razzeee:
### Expected behavior
Being able to create a shortcut to a file in my download folder that's outside of a wineprefix.

### Actual behavior
Shows an error
![image](https://user-images.githubusercontent.com/5943908/105739421-614a5200-5f38-11eb-999b-d5eb348f3576.png)

### Steps to reproduce
- Advanced mode
- Create shortcut
- Insert this
![image](https://user-images.githubusercontent.com/5943908/105739092-087ab980-5f38-11eb-85e1-eb7eae7b94ac.png)
- Try to run this file https://supportportal.gemalto.com/csm?id=kb_article_view&sys_kb_id=f5e5df901b802c50f12064606e4bcb18&sysparm_article=KB0023073

### System information
* software version (git revision) of phoenicis and scripts repositories -> current flatpak
* used operating system -> Fedora 33

### Log

```
flatpak run org.phoenicis.playonlinux 
Gtk-Message: 18:02:33.847: Failed to load module "canberra-gtk-module"
Gtk-Message: 18:02:33.848: Failed to load module "pk-gtk-module"
Gtk-Message: 18:02:33.848: Failed to load module "canberra-gtk-module"
Gtk-Message: 18:02:33.848: Failed to load module "pk-gtk-module"
ERROR StatusLogger No Log4j 2 configuration file found. Using default configuration (logging only errors to the console), or user programmatically provided configurations. Set system property 'log4j2.debug' to show Log4j 2 internal initialization logging. See https://logging.apache.org/log4j/2.x/manual/configuration.html for instructions on how to configure Log4j 2
[INFO ] org.phoenicis.repository.DefaultRepositoryManager (l.133) - Adding repositories: [org.phoenicis.repository.location.GitRepositoryLocation@57f4aaca[https://github.com/PhoenicisOrg/scripts,release/5.0-alpha-3], org.phoenicis.repository.location.ClasspathRepositoryLocation@6fbe4652[/org/phoenicis/repository]] at index 0
Jan 25, 2021 6:02:38 PM com.sun.javafx.css.StyleManager$ImageCache getCachedImage
WARNING: Error loading image: file:/home/klampe/.Phoenicis/shortcuts/MobilePass.miniature
Jan 25, 2021 6:02:39 PM com.sun.javafx.css.StyleManager$ImageCache getCachedImage
[...]

PR #2356, opened 2021-01-27
Improve error message when trying to create shortcut to executable outside container

plata:
see #2353

### DIFF
--- a/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
+++ b/phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
@@ -146,6 +146,17 @@ public class LibraryFeaturePanel extends FeaturePanel<LibraryFeaturePanel, Libra
         // get container
         // TODO: smarter way using container manager
         final String executablePath = shortcutCreationDTO.getExecutable().getAbsolutePath();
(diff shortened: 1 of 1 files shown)


phoenicis-javafx/src/main/java/org/phoenicis/javafx/components/library/control/LibraryFeaturePanel.java
```diff
@@ -146,6 +146,17 @@ public void createShortcut(ShortcutCreationDTO shortcutCreationDTO) {
...
+        if (!executablePath.startsWith(getContainersPath())) {
+            Platform.runLater(() -> {
+                final ErrorDialog errorDialog = ErrorDialog.builder()
+                        .withMessage(tr("Creating shortcut to executable outside of a container is not supported"))
+                        .withOwner(getScene().getWindow())
+                        .build();
+
+                errorDialog.showAndWait();
+            });
+            return;
```
madoar, line 158: Can we move the remaining code of this method to a new method?
I don't like the `return;` here, we could try to replace it with an `if ... else ...` syntax.
What do you think?

madoar (commented):
I think we will need to reconfigure the Checkstyle Job. The findings here are kind of annoying.

madoar:
@plata how is the current state of this PR?

plata:
Changed to `if ... else`. I don't see too much use in moving part of the code to a separate method.

PR #2358, opened 2021-01-31
Add GitHub action to update translations

plata:


plata:
I'm not really sure about the part which avoids commits if only the date changed.

madoar:
Can we check whether it works as intended by changing the job to trigger on every push?

plata:
Doesn't look as expected but no commit either.

madoar:
It seems like `Run Maven` stumbles across an error:

```
[INFO] Processing pt.po
Error:  Could not execute msgcat.
org.codehaus.plexus.util.cli.CommandLineException: Error while executing process.
    at org.codehaus.plexus.util.cli.Commandline.execute (Commandline.java:697)
    ....
```

plata:
Ah thanks. Didn't see that one. Why doesn't it fail in this case?

plata:
I still don't quite understand why we got a different order in the `keys.pot` compared to Travis. But that should be only once. So I think we can merge this.

madoar (approved):

PR #2361, opened 2021-01-31
Update translations on all branches

plata:
The action is not allowed to push to protected branches (master).


madoar:
If it is not allows to push to the master branch how do we use it?

plata:
As in this PR: update directly the branch before merging to master.

madoar (approved):

PR #2396, opened 2021-05-16
Fix Nvidia detection

plata:
GL32 deps need to be updated to properly detect Nvidia graphics.

see https://github.com/flathub/org.phoenicis.playonlinux/pull/14


Pobega:
Very strange. Have you tried opening a PR with an empty commit to see if it fails there too?

plata:
Working again without any changes. No idea what was wrong there.

PR #2409, opened 2021-06-19
CI: upload artifacts

plata:


plata:
@qparis shouldn't the packaging create a .dmg for Mac OS?

PR #2423, opened 2021-07-28
Fix "engine" written twice

omeritzics:
In string [#2334](https://crowdin.com/translate/phoenicis/3/en#2334)
––––––––––
Off-topic: Hebrew is 100% now. May you review the translations and add them to the official release, please?


plata:
This should be fixed in the source file (the `.pot` is generated).

omeritzics:
> This should be fixed in the source file (the `.pot` is generated).

Thank you, now I fixed it in the source file.

plata (approved):

plata:
Thanks for the fix.

I will look at the translations. However, I cannot promise when I will manage to do so.

omeritzics:
Thank you

PR #2602, opened 2022-11-20
CI: update to Java 17 (and 20)

plata:
required for Spring 6


Zemogiter:
We need to update the packaging scripts as well, or use github actions

plata:
Does anybody want to package locally? If not, GitHub actions should be enough.

ahmedmoselhi:
I updated it to JDK 25 using chatgpt help https://github.com/ahmedmoselhi/phoenicis

PR #2694, opened 2023-07-18
fix(sec): upgrade org.springframework:spring-core to 6.0.7

realize096:
### What happened？
There are 1 security vulnerabilities found in org.springframework:spring-core 5.3.23
- [CVE-2023-20861](https://www.oscs1024.com/hd/CVE-2023-20861)

### What did I do？
Upgrade org.springframework:spring-core from 5.3.23 to 6.0.7 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
         <logback.version>1.4.8</logback.version>
         <junit.version>4.13.2</junit.version>
         <mockito.version>5.4.0</mockito.version>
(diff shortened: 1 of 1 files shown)


plata:
Yes, we should do this. However, I have some doubts that it will build (see #2693).

